PR #1, opened 2023-11-08
Small mixer algorithm changes

biglittlebigben:
This makes sone changes to the mixer processor to align it with some of the design conventions typical in egress and ingress, and adapt some of the algorithms:
- Adopt core.Fuse for readiness and stop signaling
- Use require macros in the tests to get better details about failures
- Do not divide the mixed audio by the participant count to avoid level changes when a participant joins, or low volume
- Use saturating arithmetic 
- Support other sample rates than 8kHz
- Do not crash if less than mixSize bytes are provided for a packet, zero pad instead.
- Use a linked list for the sample packet list instead of a slice. This is a little bit more verbose to use, with type assertions, but seems like a decent fit for the use case as samples are only added or removed from the head and tail
- Use a map (used as a set) for the input list.

### DIFF
--- a/pkg/sip/media.go
+++ b/pkg/sip/media.go
@@ -117,7 +117,7 @@ func createMediaSession(conf *config.Config, participantIdentity string) (*net.U
 			SSRC:    5000,
 		},
 	}
-	audioMixer := mixer.NewMixer(func(audioSample []byte) {
+	audioMixer := mixer.NewMixer(func(audioSample []byte, sampleRate) {
 		dstAddr, ok := rtpDestination.Load().(*net.UDPAddr)
 		if !ok || dstAddr == nil {
 			return


pkg/mixer/mixer.go
```diff
@@ -15,113 +15,142 @@
...
+	out := make([]byte, 2*m.mixSize)
+	for i, sample := range mixed {
+		if sample > 0x7FFF {
+			sample = 0x7FFF
+		}
+		if sample < -0x7FFF {
+			sample = -0x7FFF
+		}
+
+		// Encoder expects little endian data (???)
```
Sean-Der, line 117: libopus accepts/emits S16LE. Should something change here?

biglittlebigben: No, it's just not the interface I was expecting for the g711 library, but it does make sense I suppose. S16LE is kind of standard for audio, but audio folks like being different 😊

PR #5, opened 2023-11-29
Wrap client and server into a Service

dennwc:


cmd/server/main.go
```diff
@@ -83,10 +85,22 @@ func runService(c *cli.Context) error {
 	killChan := make(chan os.Signal, 1)
 	signal.Notify(killChan, syscall.SIGINT)
 
-	svc := service.NewService(conf, psrpcClient, bus)
+	ua, err := sipgo.NewUA(
```
Sean-Der: @dennwc Why not put this inside the `SIPService` ?

dennwc: Good point! I'll make it init both SIP client and the server in the service constructor.

cmd/server/main.go
```diff
@@ -83,26 +81,14 @@ func runService(c *cli.Context) error {
 	killChan := make(chan os.Signal, 1)
 	signal.Notify(killChan, syscall.SIGINT)
 
-	svc := service.NewService(conf, psrpcClient, bus)
-
-	sipSrv := sip.NewServer(conf, svc.HandleTrunkAuthentication, svc.HandleDispatchRules)
```
Sean-Der, line 94: @dennwc the split of ‘service’ and ‘sip service’ comes from ingress/egress. I think it’s best to keep it that way so things feel similar

dennwc: Updated. Is this what you had in mind?

PR #10, opened 2023-11-30
Update copyright in all files.

dennwc:


Sean-Der:
@dennwc should we do https://github.com/viperproject/check-license-header ?

dennwc:
Good idea. Opened an issue to track it.

PR #35, opened 2023-12-22
add video support

Hollymingyi:
hi author,
    this is my code added for video . but when it work , call from sipjs softphone into livekit ,the video of softphone is very fuzzy, i dont know how to handle it . please give me some advice , thank you.

### DIFF
--- a/pkg/media/h264/h264.go
+++ b/pkg/media/h264/h264.go
@@ -1,0 +1,48 @@
+// Copyright 2023 LiveKit, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


iAbdulaziz93:
Hi @Hollymingyi. I'm working on supporting video calls over SIP on an application I'm working on, I'm using LiveKit. Did you get your issue solved? Is this PR is working fine on your side and the video issues like the video is fuzzing is solved on your end?

iAbdulaziz93:
Hi @dennwc. Did you have a chance to review this PR? It will help us a lot if this PR gets reviewed and approved and added support of videos to LiveKit SIP.

dennwc:
Yes, I did, but at this point it needs to be updated to match the latest codebase.

And to get accepted, it would need to properly fallback to audio-only, which is not implemented in this PR.

Also note that it can only accept one video stream, while LiveKit room typically contain multiple participants. SIP currently mixes audio from multiple participants, so for proper support it must do speaker detection or something similar to switch video steam to the talking participant. Or at least fallback to audio-only if multiple participants are present.

This is the reason why we don't have plans to support it - it's adding complexity that could be avoided if client will run LiveKit in the first place. If the client is video-capable, it usually means it could probably run a browser with LiveKit and get all benefits of it without going through SIP.

PR #44, opened 2024-01-10
move room and track to interface

AnshulMalik:


dennwc:
Hey @AnshulMalik! Thank you for contributing! Could you please explain what problem you are trying to solve with this PR? Since `NewLkRoom` is still hardcoded, I'm not sure if the change will be useful as-is.

AnshulMalik:
Hey @dennwc , You're right, this change alone doesn't acheive anything, It'll be a sequence of PRs, I want to keep PRs small and reviewable. The idea is to make it customizable, where people can bring their own implementations.

davidzhao:
@AnshulMalik, you guys are welcome to use any pieces of the code for your needs at 100ms. However I would not recommend using this service *as is*. It'll be tightly coupled with the LiveKit ecosystem and you'll likely see a lot of issues trying to bridge it with your services.

PR #46, opened 2024-01-11
Delay bridging audio for a brief moment to prevent cutoff.

dennwc:
After this change is live I'll add a follow-up to measure RTP delay. Maybe we could indeed rely on the first packet received as a signal that audio will be accepted without a cutoff.

### DIFF
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -32,6 +32,12 @@ import (
 	"github.com/livekit/sip/pkg/stats"
 )
 
+const (
+	// audioBridgeMaxDelay delays sending audio for certain time, unless RTP packet is received.
+	// This is done because of audio cutoff at the beginning of calls observed in the wild.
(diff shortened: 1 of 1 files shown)


Sean-Der:
@dennwc how hard would it be to do first remote RTP packet? Have a context we cancel `haveGottenRemoteAudio`?

Everyone will need a different delay I bet.

dennwc:
I wanted to measure the delay first, but sure, let's do it now.

But It should definitely be bound by something. So I was thinking to set it whatever comes first: 1 sec delay or first RTP packet. 

Otherwise if someone messes up firewall/routing we will block for too long waiting for RTP. Reports like "I can only hear audio one way" are easier to diagnose than "I don't hear audio at all". So I'd rather proceed with bridging after some delay (and then potentially terminate if we still get no RTP for ~30 sec).

dennwc:
@Sean-Der Done!

Sean-Der (approved):

PR #52, opened 2024-01-18
Reimplement mixer to reduce latency

dennwc:
Instead of buffering a set of frames, use a single ring buffer that stores samples. Now the mixer inputs do not allocate. 

Samples are read and mixed as soon as they are available. Currently max size of ring buffer is set to 5x of the frame size. This is done only to compensate for multiple RTP packets arriving at once. If more than 5 frames arrive at the same time, oldest ones will be dropped. Mixer now avoids most allocations as well.

Also, change the way the mixer stores input. Instead of a map, use a slice. It's done for two reasons: first it is deterministic, and second is that iteration over the inputs is more frequent when adding/removing them.


paulwe:
is this better than the deque package we've been using? https://github.com/gammazero/deque

dennwc:
@paulwe I did check a few existing implementations, but didn't find specific one, so thank you for pointing it out!

I think the one implemented in this PR is simpler (never resizes, doesn't allow index access) and it implements reading/writing in bulk. We also prefer dropping older items instead of growing the queue or blocking as most other packages tend to do.

Sean-Der:
When I do protocol bridging services I just want to worry about that code. Maybe better if this goes into https://github.com/frostbyte73/core or a dedicated repo under https://github.com/livekit ?

How much of a performance improvement is it over other implementations? I don't want to throw away code, this just means more for us to maintain.

paulwe:
protocol/utils might be the place for it if you don't want to put it here

dennwc:
There's also stuff in [pkg/media](https://github.com/livekit/sip/tree/main/pkg/media) that might be later moved out, as we discussed. Maybe `github.com/livekit/media` would be the place for it all?

This ring buffer is more media-oriented too, that's why I wrote it in the first place. Haven't found anything similar. So we can move it out with other media things.

> How much of a performance improvement is it over other implementations?

I didn't bother to test it to be honest. Adding items one by one will definitely lead to more overhead than in bulk, so :man_shrugging: Other libs assume a pointer will be passed and kept by the data structure for some time, while in this case the element will be `int16`.

And I don't want to store pointers/slices to full frames there because it will cause additional allocations. Currently I just write to this internal buffer and the caller can reuse its own slice immediately.

pkg/mixer/mixer.go
```diff
@@ -15,168 +15,146 @@
...
-		input.mu.Lock()
-
-		if !input.hasBuffered || input.samples.Len() == 0 {
-			input.mu.Unlock()
+func (m *Mixer) mixInputs() {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	for _, inp := range m.inputs {
+		n, _ := inp.readSample(m.mixTmp[:len(m.mixBuf)])
+		m.mixTmp = m.mixTmp[:n]
```
biglittlebigben: Is mixTmp ever grown back to it's full length? If not, it looks like the mixer would consume media at a slower rate than it is written? 

Even if restoring mixTmp to its full size, jitter in the ticker firing time may mean that the consumption rate may not match the production rate?

biglittlebigben: Is data that was late and didn't get mixed skipped anywhere? If not, this may cause sync issues?

biglittlebigben: Is there any buffering logic to allow some packets to arrive late?

biglittlebigben (commented):
Some comments on the mixer. I also wonder whether maintaining our own circular buffer implementation is worth it.

pkg/mixer/mixer.go
juberti: yeah, I don't see how dejitter can be avoided unless you pass through packets with seqnums intact (i.e., leaving it to the downstream jitter buffer to sort things out). naturally, this only works for cases where mixing is not needed (although that seems like a very important subset).

juberti: the general way this sort of thing is handled is via an adaptive jitter buffer as in libwebrtc, where you only incur as much delay as is needed to dejitter most of the packets.

dennwc: @biglittlebigben @juberti Thank you for the valuable feedback! Love it!

Indeed the old and the new mixers do not account for any jitter. As you pointed out, there are no checks for sequence/timestamps.

> Is mixTmp ever grown back to it's full length?

Yes, it's growing to full size when passing to `readSample` and that function returns the final length.

> If not, it looks like the mixer would consume media at a slower rate than it is written?

Even though the answer is "yes", currently the mixer assumes that it should mix at constant rate. Thus, the mix buffer is exactly 1 frame wide, while inputs buffers are larger (5 frames). I think we cannot make it emit frames faster without delaying mixing of other inputs, which I wanted to avoid.

> Even if restoring mixTmp to its full size, jitter in the ticker firing time may mean that the consumption rate may not match the production rate?

That's a very good point and it can also be true temporarily, e.g. if there are no packets, and then input emits N packets in a burst. But the rate should be the same on average. I made input buffer 5x the frame size to compensate for this.

So if the ticker causes jitter, it may lead to one or more inputs to be delayed by 1 frame. Interesting. I didn't consider this can be caused by the timer itself. Is this why you made `ignoredLastPrune` variable in the mixer code, right? I'll try to add a test for this and fix it.

> Is data that was late and didn't get mixed skipped anywhere? If not, this may cause sync issues?

> Is there any buffering logic to allow some packets to arrive late?

If the packets get delayed, and then arrive late, they will be added to the buffer (max 5 frames), which will effectively delay that particular input relative to others in the mix. But the delay is limited - as soon as we receive more than 5 frames we will start dropping oldest frames/samples from the buffer.

I assume there's no way to really catch up and reduce this delay except dropping frames or trying to speed up and resample audio in oldest frames. Since we are mixing, we cannot really send multiple frames in one go (without delaying everyone else in the mix).

> yeah, I don't see how dejitter can be avoided unless you pass through packets with seqnums intact (i.e., leaving it to the downstream jitter buffer to sort things out). naturally, this only works for cases where mixing is not needed (although that seems like a very important subset).

Indeed, it looks like I should dejitter all the inputs. For now SIP server is a very naive in this regard.

Regarding "no mixing" part, I will add a pass-through mode later as well (if there's only 1 input in the mix). 

> the general way this sort of thing is handled is via an adaptive jitter buffer as in libwebrtc, where you only incur as much delay as is needed to dejitter most of the packets.

Yeah, looks like it's unavoidable. I'd like to keep the mixer simple, because it shouldn't handle the jitter on its own, right?

I'll then implement a dejitter for each RTP input stream we receive. So the mixer will receive normalized streams and will just mix those. In that case mixer won't need so much space in input buffers too.

juberti: I think a mixer has to handle jitter on its own, since it has to emit a single time-aligned stream. this is why mix-less architectures like SFU have become popular.

juberti: Personally I think your dejitter for each stream followed by mixing of normalized streams is the right general solution. However I'd also suggest supporting a special case where you pass through what you get regardless of dejitter for the 1:1 case (= 0 extra latency)

biglittlebigben: > @biglittlebigben @juberti Thank you for the valuable feedback! Love it!
> 
> Indeed the old and the new mixers do not account for any jitter. As you pointed out, there are no checks for sequence/timestamps.
> 

The new code may not be meaningfully worse than the old one at handling late packets indeed. However, I do believe that we'll need to deal with it eventually. In my pas experience, relying on timestamps has been problematic for RTP track mixing. However, replacing sequence number gaps with silence, and using small leaky queue (with a duration of roughly the duration of the max acceptable sync drift) has worked pretty well. 

Your current approach seems close to this, with a ring buffer that holds up to 100ms of data. However, it looks like you start reading from an input immediately when added without waiting for some buffering level to be reached, which may create stuttering for that track at the beginning.

> > Is mixTmp ever grown back to it's full length?
> 
> Yes, it's growing to full size when passing to `readSample` and that function returns the final length.

Indeed. I missed that.

> 
> > If not, it looks like the mixer would consume media at a slower rate than it is written?
> 
> Even though the answer is "yes", currently the mixer assumes that it should mix at constant rate. Thus, the mix buffer is exactly 1 frame wide, while inputs buffers are larger (5 frames). I think we cannot make it emit frames faster without delaying mixing of other inputs, which I wanted to avoid.
> 
> > Even if restoring mixTmp to its full size, jitter in the ticker firing time may mean that the consumption rate may not match the production rate?
> 
> That's a very good point and it can also be true temporarily, e.g. if there are no packets, and then input emits N packets in a burst. But the rate should be the same on average. I made input buffer 5x the frame size to compensate for this.
> 

I do not see the code path where N packets may be emitted in a burst (the go ticker can drop events https://pkg.go.dev/time#NewTicker). Did I miss it?

> So if the ticker causes jitter, it may lead to one or more inputs to be delayed by 1 frame. Interesting. I didn't consider this can be caused by the timer itself. Is this why you made `ignoredLastPrune` variable in the mixer code, right? I'll try to add a test for this and fix it.
> 
> > Is data that was late and didn't get mixed skipped anywhere? If not, this may cause sync issues?
> 
> > Is there any buffering logic to allow some packets to arrive late?
> 
> If the packets get delayed, and then arrive late, they will be added to the buffer (max 5 frames), which will effectively delay that particular input relative to others in the mix. But the delay is limited - as soon as we receive more than 5 frames we will start dropping oldest frames/samples from the buffer.
> 

ok

dennwc: > I do not see the code path where N packets may be emitted in a burst (the go ticker can drop events https://pkg.go.dev/time#NewTicker). Did I miss it?

Sorry for confusion, I was referring to the RTP pushing frames to one of the inputs in a burst.

In general it seems like the the change is good enough for now, except two TODO items:

1. Make a proper bypass mode (for 1:1 calls)
2. Start reading only after we get some buffering

For the item (1) I'll need to check what fields in RTP we should keep and which we'll rewrite. We cannot pass packets directly, because we still need OPUS<->PCMU conversion. Also we currently switch mixer tracks on/off when we play pin prompts. That makes it harder to do the bypass. So I'd rather do it in a separate PR.

For the item (2), as @biglittlebigben suggested, we could skip mixing inputs that have less than 2-3 frames (~40-60ms). That will be the constant latency when having multiple participants. Would that be reasonable?

biglittlebigben: I think we should wait until we have 2-3 frames before starting to mix a track, but I think we should consume all available data once running, if needed. 

I also think that we need to address possible dropped tick events, possibly by tracking the delay since the last mix, and running the mixer enough time to emit packets covering the interval.

dennwc: @biglittlebigben Done! Mixer will now wait for 3 frames until mixing the input. It will consume all data if packets stop arriving, at which point it will start buffering again. It will also catch up in case it's scheduled too late.

biglittlebigben (approved):

PR #56, opened 2024-01-29
Make sure all metrics are registered

dennwc:
Up to this point we ere missing the call errors metric. Call termination metric was correct though.


davidzhao (approved):
lg

PR #60, opened 2024-02-12
Support dropping connections to prevent port scans

dennwc:


pkg/sip/inbound.go
```diff
@@ -214,12 +238,32 @@ func (c *inboundCall) handleInvite(ctx context.Context, req *sip.Request, tx sip
...
+		FromUser:   c.from.Address.User,
+		ToUser:     c.to.Address.User,
+		ToHost:     c.to.Address.Host,
+		SrcAddress: c.src,
+		Pin:        "",
+		NoPin:      false,
+	})
+	switch disp.Result {
+	default:
+		logger.Errorw("Rejecting inbound call", fmt.Errorf("unexpected dispatch result: %v", disp.Result), "from", c.from.Address.User, "to", c.to.Address.User, "to-host", c.to.Address.Host, "src", c.src)
```
biglittlebigben, line 251: Could this error ever be triggered by bad user data? If so, we usually don't log these at the Error level as we like alarming on Errors messages.

pkg/sip/inbound.go
```diff
@@ -276,10 +320,13 @@ func (c *inboundCall) handleInvite(ctx context.Context, req *sip.Request, tx sip
 		delay.Stop()
 	case <-delay.C:
 	}
-	if requirePin {
+	switch disp.Result {
+	default:
+		panic("unreachable")
```
biglittlebigben: What is the consequence of ever reaching this case? It seems like it may become a forward compatibility issue. Generally, panicking in a service is something that I think only makes sense to avoid undetermined behavior, or security issues that are not recoverable any other way.

pkg/sip/inbound.go
dennwc: No, I don't think so. It will only happen if we add some other status to the protocol.

pkg/sip/inbound.go
dennwc: Agree, I'd really like to have some static "unreachable" check done by the compiler.

Will copy the code from the `default` case above then.

dennwc: Done

biglittlebigben (approved):

PR #61, opened 2024-02-22
Test SIP signaling and audio with multiple participants

dennwc:


.github/workflows/test.yaml
```diff
@@ -17,6 +17,7 @@ on:
   push:
     branches:
       - main
+      - ci_test
   pull_request:
```
biglittlebigben, line 21: We should probably add "main" to the pull_request section here as well?

pkg/sip/server.go
```diff
@@ -157,9 +158,12 @@ func (s *Server) Start(agent *sipgo.UserAgent) error {
 	// Ignore ACKs
 	s.sipSrv.OnAck(func(req *sip.Request, tx sip.ServerTransaction) {})
 
-	// TODO: pass proper context here
+	ctx, stop := context.WithCancel(context.Background())
+	s.sipSrvStop = stop
 	go func() {
-		panic(s.sipSrv.ListenAndServe(context.TODO(), "udp", fmt.Sprintf("0.0.0.0:%d", s.conf.SIPPort)))
+		if err := s.sipSrv.ListenAndServe(ctx, "udp", fmt.Sprintf("0.0.0.0:%d", s.conf.SIPPort)); err != nil {
+			panic(err)
```
biglittlebigben: Should we add a message beyond the error to this panic, stating that the issue is binding the server UDP port?

pkg/siptest/client.go
```diff
@@ -0,0 +1,561 @@
...
+// limitations under the License.
+
+package siptest
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"log/slog"
```
biglittlebigben, line 22: Any reason not to use the "livekit/protocol/logger" package that we usually use?

biglittlebigben (commented):
3 nits, but LGTM. Thanks for writing a lot of scaffolding code here. The ingress and egress tests may benefit from launching their own livekit server as you do here.

biglittlebigben (approved):

pkg/siptest/client.go
dennwc: There was a discussion about migrating to `slog` in the mid term. So I wanted to start doing so at least in tests and eventually update `protocol/logger` to interoperate with `slog` too. I think usage of `slog` here is reasonable because tests don't need any integrations with external services that `protocol/logger` provides.

PR #66, opened 2024-02-27
Implement outbound call hangups

Sean-Der:
* Outbound was sharing a port with server. BYE requests would be routed into the server handler. We no longer share the UserAgent so Outbound+Inbound is happening on distinct ports

* Add `Bye` handler to Outbound handler

* Properly build the outbound INVITE request. From header didn't include the CallerID

* Properly build the outbound Ack. Was missing Record-Route


dennwc (approved):
Thank you a lot!

PR #68, opened 2024-02-28
Reuse media conn code in the test client

dennwc:


pkg/siptest/client.go
```diff
@@ -489,16 +463,16 @@ func (c *Client) SendSignal(ctx context.Context, n int, val int) error {
...
 func (c *Client) WaitSignals(ctx context.Context, vals []int, w io.WriteCloser) error {
-	buf := make([]byte, 1500)
 	var ws media.PCM16WriteCloser
 	if w != nil {
 		ws = webmm.NewPCM16Writer(w, sampleRate, sampleDur)
 		defer ws.Close()
 	}
 	lastLog := time.Now()
 	for {
-		n, _, err := c.media.ReadFromUDP(buf)
```
Sean-Der, line 500: `ReadRTP` allocates every time/is more expensive. When we do performance tuning it will come up, worth doing this change?

pkg/siptest/client.go
```diff
@@ -157,21 +154,13 @@ func (c *Client) Close() {
 }
 
 func (c *Client) record(w io.WriteCloser) error {
-	buf := make([]byte, 1500)
```
Sean-Der, line 160: Same as note below. More expensive, so maybe we should keep as is?

Sean-Der (approved):
My mistake! You have your own `ReadRTP` LGTM!

pkg/siptest/client.go
dennwc: I'm not sure I understand. The buffer is not created in https://github.com/livekit/sip/pull/68/files#diff-4b88e8e2585024f0c6957070759cb05f9255f0597b5a9aba75dc3fb6ee9e2b87R30 and is reused as before. Or do you mean the struct for RTP packet itself?

PR #75, opened 2024-03-20
Add tests for outbound calls

dennwc:
Run two LK and SIP servers during tests and make a SIP call from one to the other. Verify that LK participants on both sides can communicate using SIP as a bridge.


pkg/media/pipe.go
```diff
@@ -0,0 +1,184 @@
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// This is a copy of io/pipe.go adapted to media.
+
+package media
```
biglittlebigben, line 9: Could you comment on the differences with the Go native library implementation?

dennwc: Sure, will add a comment. It's just a generic version of the same code, plus it's adapted to media interfaces in SIP (`WriteSample` instead of `Write`).

dennwc: @biglittlebigben Done!

biglittlebigben: There is always a maintenance question about forking library code. Is this used in the tests only? If so, should we move it to a test related package to discourage use in service code?

dennwc: Only used in testing code for now, but it could be useful for service code as well later.

The whole `media` package is kind of experimental. The goal is to see what interfaces make sense and if it could be used in other services or Go SDK.

The Reader/Writer interfaces are well-known, so I wanted to try using similar approach for reading/writing RTP and/or audio data. For now it's working quite well. So I'm expanding it a bit over time.

biglittlebigben (approved):

PR #93, opened 2024-05-21
Cleanup rooms after the test, more logs

dennwc:


pkg/sip/inbound.go
```diff
@@ -122,6 +127,19 @@ func (s *Server) handleInviteAuth(log logger.Logger, req *sip.Request, tx sip.Se
 func (s *Server) onInvite(req *sip.Request, tx sip.ServerTransaction) {
 	ctx := context.Background()
 	s.mon.InviteReqRaw(stats.Inbound)
+	callID := lksip.NewCallID()
+	src := req.Source()
+	log := s.log.WithValues(
+		"call-id", callID,
```
frostbyte73: we use camel case everywhere else for logging

frostbyte73 (approved):

pkg/sip/inbound.go
dennwc: Done, renamed all other fields too.

PR #108, opened 2024-06-11
Increase mixing sample rate to 48k. Fix and enable G722.

dennwc:
- Update audio pipeline to 48k sample rate. Upsample/downsample when needed.
- Differentiate between RTP clock rate and audio sample rate (these are different for G722).
- Fix and re-enable G722 codec. Tested with Telnyx.
- Re-encode the audio prompts in 48k. Repackage them in `.ogg`.
- Add `pprof` support (disabled by default).


test/lktest/livekit.go
```diff
@@ -185,7 +186,8 @@ const (
 )
 
 func (p *Participant) SendSignal(ctx context.Context, n int, val int) error {
-	signal := make(media.PCM16Sample, rtp.DefPacketDur)
+	const framesPerSec = int(time.Second / rtp.DefFrameDur)
```
frostbyte73, line 192: what is `const` for?

test/lktest/livekit.go
```diff
@@ -218,11 +220,12 @@ func (p *Participant) SendSignal(ctx context.Context, n int, val int) error {
 func (p *Participant) WaitSignals(ctx context.Context, vals []int, w io.WriteCloser) error {
 	var ws media.PCM16WriteCloser
 	if w != nil {
-		ws = webmm.NewPCM16Writer(w, rtp.DefSampleRate, rtp.DefFrameDur)
+		ws = webmm.NewPCM16Writer(w, RoomSampleRate, rtp.DefFrameDur)
 		defer ws.Close()
 	}
 	lastLog := time.Now()
-	buf := make(media.PCM16Sample, rtp.DefPacketDur)
+	const framesPerSec = int(time.Second / rtp.DefFrameDur)
```
frostbyte73, line 234: same here

test/lktest/livekit.go
frostbyte73: also will using an int here cause drift if it isn't a round number?

dennwc: The RTP packet duration is always 20 ms (at least for SIP), so it will always be a round number.

I can put a static compiler check here:
```go
var _ = [1]struct{}{}[time.Second % rtp.DefFrameDur] // a%b == 0
```

Changing to float will complicate things here a bit.

biglittlebigben (commented):
LGTM after addressing dc's comment (why const and static check)

frostbyte73 (approved):

PR #116, opened 2024-06-27
Set SIP participant attributes

dennwc:
Attributes support was recently merged to the server and SDK, and this change uses these attributes for SIP.

For inbound, Cloud IO generates a token after evaluating dispatch rules. This token will already contain all the basic metadata (phone numbers, trunk ID, etc), but does not contain provider-specific ones. Thus, after SIP invite response, the SIP participant will update it's own metadata with these new attrs.

For inbound in OSS, the path is a bit different. Since SIP can sign its own token, it can include provider-specific metadata into it without triggering the updated. I choose to disable this behavior and intentionally omit provider attributes to minimize difference between OSS and Cloud. Later LK OSS server will generate SIP token the same way as Cloud does. There will be no need to set it in the config too.

For outbound the result is similar - first, the participant has to be created in the room (to play ringtone, or just indicate its presence) and only after some time we get invite response which allows us to set provider attributes.


biglittlebigben (approved):

dennwc:
Waiting for https://github.com/livekit/server-sdk-go/pull/496.

Issue #112, opened 2024-06-19
Closing room, BYE To Participant fails - 481 Call/Transaction Does Not Exist


ftsef:
Hello, 
I’ve discovered an issue that a `BYE` send from the LIVEKIT-SIP Server is not recognized by Asterisk.
The SIP log showed a `481 Call/Transaction Does Not Exist`.

It's reproducable on multiple Systems by on of the following steps:
- removing a SIP-participant from a room
- closing/deleteing a Room with a SIP-participant in it

The SIP participant will not be hung up and stays _connected_ (to the void).
I've used the latest docker image as well as build from master branch.

I'm going to check the RFC specifications but I thing it has something to do with the 
missing FROM `tag` in the BYE package.

Do you have a hint what it could be?

Thanks in advance.

```
<--- Received SIP request (923 bytes) from UDP:192.168.178.21:57881 --->
INVITE sip:100@192.168.178.42 SIP/2.0
Via: SIP/2.0/UDP 192.168.178.21:57881;branch=z9hG4bK.u4EqNs7xF;rport
From: "1108" <sip:1108@192.168.178.42>;tag=4c95ddtPY
To: sip:100@192.168.178.42
CSeq: 20 INVITE
Call-ID: sWT0ND51UG
Max-Forwards: 70
Supported: replaces, outbound, gruu, path, record-aware
Allow: INVITE, ACK, CANCEL, OPTIONS, BYE, REFER, NOTIFY, MESSAGE, SUBSCRIBE, INFO, PRACK, UPDATE
Content-Type: application/sdp
Content-Length: 300
Contact: <sip:1108@192.168.178.21:57881;transport=udp>;expires=3599
User-Agent: Linphone-Desktop/5.2.4 (MBP.fritz.box) osx/14.5 Qt/5.15.2 LinphoneSDK/5.3.41

v=0
o=1108 3224 3649 IN IP4 192.168.178.21
s=Talk
c=IN IP4 192.168.178.21
t=0 0
a=rtcp-xr:rcvr-rtt=all:10000 stat-summary=loss,dup,jitt,TTL voip-metrics
a=record:off
m=audio 56905 RTP/AVP 0 101
a=rtpmap:101 telephone-event/8000
a=rtcp:51058
a=rtcp-fb:* trr-int 1000
a=rtcp-fb:* ccm tmmbr

<--- Transmitting SIP response (482 bytes) to UDP:192.168.178.21:57881 --->
SIP/2.0 401 Unauthorized
[...]

PR #118, opened 2024-07-09
Add TO-Tag to BYE

ftsef:
Hi,

as described in https://github.com/livekit/sip/issues/112 this PR addresses the issue that BYE is not received due to missing TO-Tag.

My fix/workaround so far is to read the FROM-tag from the next ACK and store it in the current calls TO-tag.
This works in my use case. I don't know if that a good solution in general.

### DIFF
--- a/pkg/sip/server.go
+++ b/pkg/sip/server.go
@@ -219,18 +219,22 @@ func (s *Server) Start(agent *sipgo.UserAgent, unhandled sipgo.RequestHandler) e
 
 	// Ignore ACKs
 	s.sipSrv.OnAck(func(req *sip.Request, tx sip.ServerTransaction) {
-	        fromTag, err := getTagValue(req)
-	        if err != nil {
-	            sipErrorResponse(tx, req)
-	            return
-	        }
-	
-	        if c, found := s.activeCalls[fromTag]; found {
-	            if header, ok := req.To(); ok {
-	                c.to.Params["tag"] = header.Params["tag"]
-	            }
-	        }
-    	})
+		fromTag, err := getTagValue(req)
+		if err != nil {
+			sipErrorResponse(tx, req)
+			return
+		}
+
+		s.cmu.RLock()
+		c, found := s.activeCalls[fromTag]
+		s.cmu.RUnlock()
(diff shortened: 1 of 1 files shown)


pkg/sip/server.go
```diff
@@ -218,7 +218,19 @@ func (s *Server) Start(agent *sipgo.UserAgent, unhandled sipgo.RequestHandler) e
 	s.sipUnhandled = unhandled
 
 	// Ignore ACKs
-	s.sipSrv.OnAck(func(req *sip.Request, tx sip.ServerTransaction) {})
+	s.sipSrv.OnAck(func(req *sip.Request, tx sip.ServerTransaction) {
+	        fromTag, err := getTagValue(req)
+	        if err != nil {
+	            sipErrorResponse(tx, req)
```
dennwc: We may ignore an error here, I think.

pkg/sip/server.go
```diff
@@ -218,7 +218,19 @@ func (s *Server) Start(agent *sipgo.UserAgent, unhandled sipgo.RequestHandler) e
...
+	            return
+	        }
+	
+	        if c, found := s.activeCalls[fromTag]; found {
```
```suggestion
	        s.cmu.RLock()
	        c, found := s.activeCalls[fromTag]
	        s.cmu.RUnlock()
	        if found {
```
Must hold mutex when accessing `activeCalls`.

pkg/sip/server.go
```diff
@@ -218,7 +218,19 @@ func (s *Server) Start(agent *sipgo.UserAgent, unhandled sipgo.RequestHandler) e
...
+	            if header, ok := req.To(); ok {
+	                c.to.Params["tag"] = header.Params["tag"]
```
dennwc: We might eventually want to send other SIP commands, so lets store this tag as a new field in the `inboundCall` struct and set it specifically when sending `BYE`.

I will need to check RFC to see if we must use this tag for all future commands or not.

dennwc:
I found a more direct way to set the tag, so I'll close this one in favor of a new change. Thank you again for pointing our the root cause of the issue!

PR #120, opened 2024-07-10
Forward DTMF from LiveKit room to SIP

dennwc:


pkg/sip/room.go
```diff
@@ -110,6 +112,14 @@ func (r *Room) Connect(conf *config.Config, roomName, identity, name, meta strin
 				h := rtp.NewMediaStreamIn[opus.Sample](odec)
 				_ = rtp.HandleLoop(track, h)
 			},
+			OnDataPacket: func(data lksdk.DataPacket, params lksdk.DataReceiveParams) {
+				switch data := data.(type) {
+				case *livekit.SipDTMF:
```
davidzhao, line 117: did we say that this should be generated only if DTMF was targeted for the current participant? i.e. if another SIP user dials into the room and emits DTMF, we'll would:
* send DTMF message to the room
* have the audio tones mixed in output for the current SIP participant
* it won't be necessary to send another DTMF to this current participant

pkg/sip/room.go
```diff
@@ -181,8 +191,41 @@ func (r *Room) SetOutput(out media.Writer[media.PCM16Sample]) {
...
 
+func (r *Room) SetDTMFOutput(out *rtp.Stream) {
+	if r == nil {
+		return
+	}
+	if out == nil {
+		r.outDtmf.Store(nil)
+		return
+	}
+	r.outDtmf.Store(out)
```
biglittlebigben, line 202: Do we need to single out the `out == nil` case here?

pkg/sip/room.go
dennwc: There's a TODO just below this line, but I spelled it incorrectly. It should be saying "don't generate **audio** DTMF if it's from SIP". So we always should generate digital DTMF, but skip audio tones if it's from SIP (because it's mixed from SIP audio anyway).

For now the use case of the customer is to send DTMF from LK participant to SIP, so it won't be hit by this.

Since they want this feature faster, I for now still generate audio DTMF for SIP-SIP. It would take a bit longer to skip this because SDK will need to expose a different signature for `OnDataPacket` hook. And that one might take time to review.

pkg/sip/room.go
dennwc: It's possible in this particular case, but I also use interface type in one similar place. Setting typed nil pointer to interface will lead to non-nil value that will panic when any method is called. So I'm being cautious here, just in case I will change the type to an interface later.

davidzhao (approved):

pkg/sip/room.go
davidzhao: so you are saying that in order to send DTMF to a participant. you would just send DTMF into the room? and there is no need to ensure that message is only delivered to the specific participant?

dennwc: Correct. Initially I planned to only deliver targeted DTMF. But the problem is - DTMF audio tones use the same media channel as the SIP participant itself. Thus, tones will still be delivered to all participants regardless of our logic here.

PR #121, opened 2024-07-10
Implement SIP call statuses

dennwc:


pkg/sip/client.go
```diff
@@ -158,7 +158,7 @@ func (c *Client) CreateSIPParticipant(ctx context.Context, req *rpc.InternalCrea
 		}
 		select {
 		case <-call.Disconnected():
-			call.CloseWithReason("removed")
+			call.CloseWithReason("", "removed")
```
biglittlebigben: Would it make sense to define a (typed) constant for `""` here?

biglittlebigben (approved):

PR #124, opened 2024-07-18
Add jitter buffer to all RTP streams

dennwc:
Add jitter buffer to SIP and LK RTP streams.

Also separate DTMF and audio sequence numbers and SSRC.


pkg/media/rtp/jitter.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+const (
+	jitterMaxLatency = 2 * time.Second
+)
+
+func HandleJitter(clockRate int, h Handler) Handler {
+	return &jitterHandler{
+		h:   h,
+		buf: jitter.NewBuffer(audioDepacketizer{}, uint32(clockRate), jitterMaxLatency),
```
biglittlebigben, line 31: This will add up to a 2 second delay to the SIP stream. Is this the behavior we want?

frostbyte73 (approved):

pkg/media/rtp/jitter.go
frostbyte73: how is jitter handled on the clients? we might still want it, but with a much lower max latency than 2s

davidzhao: given it's a phone convo.. I think we'll want to limit delay to <500ms

dennwc: Done, set to 500 ms.

biglittlebigben: Also wondering what the mixer is going to do. Is it going wait for the missing track, or will it mix whatever tracks it has before that?

dennwc: The mixer will still buffer ~60 ms of audio.

If because of a jitter these packets are delayed more than 60 ms, mixer will at first use the buffered frames. As soon as it reaches 0 buffer on that track, mixer will mute it track and will mix the rest. As soon as 60 ms is available on that track again, it will be unmuted and mixed as before.

biglittlebigben: Does this means that if the jitter buffer delays samples by more than 60ms, the participant will be muted until late packets arrive, and then speech will restart, but be mixed in late, up to 500ms, until the participant leaves?

dennwc: Right, if the jitter buffer delays packets up to 500 ms, the participant will be muted during this time, but then only last 100 ms of that track will resume and this delay will persist. The 60 ms I mentioned is the target latency, while 100 ms is the max buffer.

Actually, it will almost certainly drop from 100 ms to at least 80 ms because some packets might arrive in a burst and at least one will go over the max buffer, causing drop of an older frame.

Ideally mixer should buffer all 500 ms for a track, delay it after resume, but also resample so that it can eventually catch up. But this is a bit too complicated for now.

We could increase the mixer buffer up to ~250 ms, and thus let it drop half of the frames in case of extreme jitter. But still, this delay will persist until the participant leaves, so it might not be acceptable.

biglittlebigben: I just think that we should make the jitter buffer and mixer delay consistent as there is no point waiting for data we'll drop in the pipeline.

dennwc: In that case it should be 60 ms I think.

PR #129, opened 2024-07-31
nil room panic

real-danm:
room is not always set, just warn for now when it's empty and we're unable to set the participant attributes.

### DIFF
--- a/pkg/sip/outbound.go
+++ b/pkg/sip/outbound.go
@@ -386,7 +386,7 @@ func (c *outboundCall) sipSignal(conf sipOutboundConfig) error {
 		if len(extra) != 0 {
 			room := c.lkRoom.Room()
                          if room != nil {
-				c.lkRoom.Room().LocalParticipant.SetAttributes(extra)
+				room.LocalParticipant.SetAttributes(extra)
 			} else {
 				c.log.Warnw("could not set attributes on nil room", nil, "attrs", extra)
 			}
--- a/pkg/sip/outbound.go
+++ b/pkg/sip/outbound.go
@@ -384,7 +384,8 @@ func (c *outboundCall) sipSignal(conf sipOutboundConfig) error {
 			}
 		}
 		if len(extra) != 0 {
-			if c.lkRoom.Room() != nil {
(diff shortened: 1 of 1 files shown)


pkg/sip/outbound.go
```diff
@@ -383,7 +384,11 @@ func (c *outboundCall) sipSignal(conf sipOutboundConfig) error {
 			}
 		}
 		if len(extra) != 0 {
-			c.lkRoom.Room().LocalParticipant.SetAttributes(extra)
+			if c.lkRoom.Room() != nil {
```
davidzhao:
```suggestion
			room := c.lkRoom.Room()
                         if room != nil {
```

davidzhao (commented):
good catch!

davidzhao (commented):
this looks good.. though it raises a bigger question of why is Room() nil, and the flow of the app when it is unset. doing a quick scan I think we are doing Room().fnc() pattern quite often elsewhere in the code.

pkg/sip/outbound.go
```diff
@@ -383,7 +384,12 @@ func (c *outboundCall) sipSignal(conf sipOutboundConfig) error {
 			}
 		}
 		if len(extra) != 0 {
-			c.lkRoom.Room().LocalParticipant.SetAttributes(extra)
+			room := c.lkRoom.Room()
+                         if room != nil {
+				c.lkRoom.Room().LocalParticipant.SetAttributes(extra)
```
davidzhao:
```suggestion
				room.LocalParticipant.SetAttributes(extra)
```

davidzhao (approved):

real-danm:
> this looks good.. though it raises a bigger question of why is Room() nil, and the flow of the app when it is unset. doing a quick scan I think we are doing Room().fnc() pattern quite often elsewhere in the code.

it's due to the way the call is initiated.  the panic coincides with a `Closing outbound call with error` with reason `media-timeout`. So perhaps we need to ensure we're connected before accessing the Room.

connection is started here https://github.com/livekit/sip/blob/main/pkg/sip/outbound.go#L80-L82

PR #137, opened 2024-08-12
allow calls to complete during a graceful shutdown

real-danm:
Wait for existing calls to complete before shutting down when receiving SIGTERM or SIGQUIT.

Forceful/immediate shutdown will still happen with SIGINT.

### DIFF
--- a/pkg/service/service.go
+++ b/pkg/service/service.go
@@ -142,18 +142,17 @@ func (s *Service) Run() error {
 			s.DeregisterCreateSIPParticipantTopic()
 
 			if !s.killed.Load() {
-				activeCalls := s.sipServiceActiveCalls()
-				if activeCalls > 0 {
-					s.log.Infow("waiting for calls to finish", "calls", activeCalls)
-					shutdownTicker := time.NewTicker(5 * time.Second)
-					for {
-						<-shutdownTicker.C
-						activeCalls = s.sipServiceActiveCalls()
-						if activeCalls == 0 {
-							break
-						}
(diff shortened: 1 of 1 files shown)


pkg/service/service.go
```diff
@@ -145,9 +144,16 @@ func (s *Service) Run() error {
...
+					for {
+						<-shutdownTicker.C
+						activeCalls = s.sipServiceActiveCalls()
+						if activeCalls == 0 {
+							break
+						}
+						s.log.Infow("waiting for calls to finish", "calls", activeCalls)
+					}
 				}
 			}
```
dennwc: Good catch! I think the outer loop was supposed to do this, but it was lost in the initial refactor, probably :thinking: 

Now in the new version the inner loop will need to check for `s.killed` as well (if we decide to kill it after graceful shutdown was started). Also a good idea to  `defer shutdownTicker.Stop()`.

Maybe something like this?

```go
	if !s.killed.Load() {
	      shutdownTicker := time.NewTicker(5 * time.Second)
	      defer shutdownTicker.Stop()
	      for !s.killed.Load() {
		      activeCalls = s.sipServiceActiveCalls()
		      if activeCalls == 0 {
			      break
		      }
		      s.log.Infow("waiting for calls to finish", "calls", activeCalls)
		      <-shutdownTicker.C
	      }
	}
```

dennwc (approved):

pkg/service/service.go
real-danm: yeah good catch/refactor will update!

PR #144, opened 2024-08-16
Improve metrics, make it safer

dennwc:
A few small-ish fixes for the metrics:
- Simplify outbound code a bit.
- Make sure outbound call metric is decreased.
- Adjust duration buckets for short and long operations.
- Use from/to host instead of a full URI or the number for the metric.


real-danm (approved):
LGTM!

PR #156, opened 2024-08-24
Do not fail in case of single opus decode failure. Do not block track subscribed callback

biglittlebigben:


biglittlebigben:
The only case where I get the corruption so far is when using `lk load-test` as it loops an opus file. It's debatable whether we need that part of the fix?

pkg/sip/room.go
```diff
@@ -120,19 +122,25 @@ func (r *Room) Connect(conf *config.Config, rconf RoomConfig) error {
...
-				mTrack := r.NewTrack()
-				defer mTrack.Close()
 
-				odec, err := opus.Decode(mTrack, channels)
-				if err != nil {
-					r.log.Errorw("cannot create opus decoder", err, "trackID", pub.SID())
-					return
-				}
-				defer odec.Close()
+				go func() {
```
dennwc, line 126: Isn't it running in a separate goroutine already?

pkg/media/opus/opus.go
```diff
@@ -68,8 +70,15 @@ func (d *decoder) SampleRate() int {
 func (d *decoder) WriteSample(in Sample) error {
 	n, err := d.dec.Decode(in, d.buf)
 	if err != nil {
-		return err
+		// Some workflows (concatenating opus files) can cause a suprious decoding error, so ignore single corruption errors
+		if !strings.Contains(err.Error(), "corrupted stream") || d.successiveErrorCount >= 1 {
```
dennwc: Any chance we can assert this error with `errors.Is/As`?

pkg/media/opus/opus.go
```diff
@@ -2,7 +2,7 @@
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
+// You may obtain a copy of the License açt
```
dennwc: nit

dennwc (approved):
Sure, lets do it. Since you specifically handle only one error in a row, I think it's a good idea in general. If Opus can recover after a corrupted frame, I see no harm in that.

pkg/sip/room.go
biglittlebigben: Not that I can tell from a cursory look at the GO SDK, but maybe a missed something?

dennwc: If I traced it correctly, it goes from the Pion's WebRTC (`PeerConnection.onTrack`). But I agree, lets make it explicit here.

biglittlebigben: Right, I did not go that far back on the call stack because there is some code in the sdk part of the callback that doesn't get to run if we block here.

pkg/media/opus/opus.go
```diff
@@ -65,11 +66,11 @@ func (d *decoder) SampleRate() int {
...
 	n, err := d.dec.Decode(in, d.buf)
 	if err != nil {
-		// Some workflows (concatenating opus files) can cause a suprious decoding error, so ignore single corruption errors
-		if !strings.Contains(err.Error(), "corrupted stream") || d.successiveErrorCount >= 1 {
+		// Some workflows (concatenating opus files) can cause a suprious decoding error, so ignore small amount of corruption errors
+		if !errors.Is(err, opus.ErrInvalidPacket) || d.successiveErrorCount >= 5 {
 			return err
 		}
-		log.Debugw("opus decoder failed decoding a sample")
+		logger.Debugw("opus decoder failed decoding a sample")
```
dennwc: Hmm, we should probably start passing call's logger to `Decode` and `Encode` constructors then. Otherwise it will be hard to track down the call where this happened. I can do it in a separate PR, if you want.

dennwc (approved):

PR #157, opened 2024-08-24
Fix SDP format order. Update SDP dep.

dennwc:


pkg/sip/signaling.go
```diff
@@ -90,16 +90,6 @@ func sdpMediaOffer(rtpListenerPort int) []*sdp.MediaDescription {
 			Value: styp + " " + codec.Codec.Info().SDPName,
 		})
 	}
-	slices.SortFunc(formats, func(a, b string) int {
```
davidzhao, line 93: not sorting makes sense, to respect the preference of the offer. do you remember why we sorted to begin with?

davidzhao (approved):

pkg/sip/signaling.go
dennwc: I think it was a mistake to begin with. The `rtpmap` were always in priority order, I probably forgot to remove sorting here.

PR #158, opened 2024-08-24
Update sipgo. Fix address parsing.

dennwc:
Update sipgo which includes a few back-ported commits from upstream. This fixes address parsing, allowing use of Linphone to accept outbound calls.


davidzhao (approved):
nice!

PR #159, opened 2024-08-26
Do not reset resampler. Prefer latency over quality.

dennwc:


pkg/media/media.go
```diff
@@ -130,3 +165,61 @@ func (s MultiWriter[T]) WriteSample(sample T) error {
...
+
+func (w *fileWriter[T]) String() string {
+	return fmt.Sprintf("RawFile(%d)", w.sampleRate)
+}
+
+func (w *fileWriter[T]) SampleRate() int {
+	return w.sampleRate
+}
+
+func (w *fileWriter[T]) WriteSample(sample T) error {
```
biglittlebigben, line 202: Could this be replaced by a call to io.Copy (with Sample implementing WriteTo), or io.CopyBuffer?

biglittlebigben (approved):

pkg/media/media.go
dennwc: Yes, if we stick to `[]byte` as the only sample implementation.

But there reason it's done as it is right now, is to have full type safety. Even though G722 and Opus samples are `[]byte` they are type-incompatible in this API. So it's impossible to make a mistake by connecting pipes with wrong formats together.

As second benefit of this approach is that we can pass around non-byte slices like `[]int16` for PCM. And it will work well if we'd want to introduce new types like `[]float32`. This is great when the samples need to hit the mixer or any other native computation. Currently the resampler is not optimized, for example, and needs type conversion. But in general it can be optimized for zero allocation too.

PR #163, opened 2024-08-30
Unit and integration test for DTMF reception on SIP client and timing

biglittlebigben:


biglittlebigben:
I haven't been able to make sure integration tests still pass as they don't seem to run on OSX (media time out)

davidzhao (approved):
lgtm!

pkg/siptest/client.go
```diff
@@ -503,12 +524,32 @@ func (c *Client) WaitSignals(ctx context.Context, vals []int, w io.WriteCloser)
...
+
+		select {
+		case <-ctx.Done():
+			close(pkts)
+			close(done)
+			return ctx.Err()
+		case pkts <- pkt:
 		}
+
+		pkts <- pkt
```
dennwc: Looks like it's sending `pkt` twice: once here and once in the `select` above.

test/integration/sip_test.go
```diff
@@ -223,7 +224,12 @@ func runClientWithCodec(t testing.TB, conf *NumberConfig, id string, number stri
 		OnMediaTimeout: func() {
 			t.Fatal("media timeout")
 		},
-	})
+	}
+	if onDTMF != nil {
+		cconf.OnDTMF = onDTMF
```
dennwc: Could be just set in the struct init above

dennwc (approved):

Issue #136, opened 2024-08-08
Tracking Exact Moment When User Answers Outbound Call


vlxdisluv:
We are facing an issue with tracking the exact moment when a user answers a call (picks up the phone) during outbound calling. We wanted to use the participant joined webhook for this, but we noticed that it triggers when the ringing starts, not when the user actually picks up.

Is there a way to more accurately track the moment when the user picks up the phone during an outbound call? Perhaps there is another webhook or method that we can use for this purpose?

Thank you for your assistance!

davidzhao:
this makes sense.. we need to align the lifecycle better to match user expectations. we'll work on improving this behavior

vlxdisluv:
Hi @davidzhao , could you kindly clarify which status I should listen for to identify that the user has picked up the phone during an outbound call? If I'm not mistaken, I should listen to the tracks?

vlxdisluv:
> Hey @vlxdisluv! First, participant created by SIP will update it's `sip.callStatus` to `active`, which can be observed with participant attribute hook in the SDK. Second, outbound SIP participant will not subscribe to anyone's tracks until the call is dialed. So this could be another indicator.

Is there a way to reliably detect whether a SIP INVITE request was unsuccessful, particularly when the SIP trunk service is unavailable and a 500 error is returned? Can this be tracked through specific events or mechanisms? Additionally, is it possible to determine whether the call was answered by the intended recipient or an answering machine the call was forwarded to?

dennwc:
Yes, SIP service will set `DisconnectReason` for a LK participant to a specific value in this case, see https://github.com/livekit/sip/issues/191. The actual change is https://github.com/livekit/protocol/pull/845.

Regarding answering machine detection - we do not support it directly yet. Should be done externally with an Agent for now. The issue is https://github.com/livekit/sip/issues/117. We now plan to emit specific events via the data channel when SIP service detects telephony tones. This may help the Agent detect it more reliably.

onur-yildirim-infinitusai:
@dennwc Tracking on this to see whether anything in progress to emit the events for SIP Service answering machine detection

amirjabarivasal-ops:
Guessing this still hasn't been solved - major issue

Issue #114, opened 2024-06-20
Outbound Call Bot Starts Talking Before Call is Picked Up


akshatvg:
**Description:**
I'm experiencing an issue with the outbound call feature where the bot starts talking before the person picks up the call. I am using Twilio SIP for making these calls. Ideally, I want the bot to start its interaction only after the call has been answered. Below are the details of my setup:

**Steps to Reproduce:**
1. Configure an outbound call using Twilio SIP.
2. Initiate the call.
3. Observe the bot starts talking before the call is picked up by the recipient.

**Expected Behavior:**
The bot should begin speaking only after the recipient has picked up the call.

**Actual Behavior:**
The bot starts talking immediately after the call is initiated, even if the recipient hasn't picked up the call yet.

**Environment:**
- LiveKit Version:

```
livekit==0.9.2
livekit-agents==0.4.0
livekit-api==0.4.3
livekit-protocol==0.3.2
```

- Twilio SIP setup

**Possible Workaround:**
Is there any configuration in LiveKit or Twilio SIP that I might have missed which could delay the bot's speech until the call is picked up?

Thank you for your assistance in resolving this issue.

dennwc:
Indeed, SIP creates room participant before the call is established and may play a ringtone to the room during this time. Once the call is picked up, it will switch audio to it.

Thus, this problem can be solved in a few ways:
1. SIP should not create participant before the call is established. It's unlikely that we will implement it, because we'd still want see some visual indication of the call being made.
2. SIP can publish call status updates. This will require Agents framework to also wait for certain status before starting the actual processing.
3. If ringtone is disabled, SIP could avoid track publishing until call is connected. This might be the cleanest way. Agents might also need to wait for the track.

cc @keepingitneil

davidzhao:
I think we need a bit of explicit signaling (perhaps via the new attributes system) to signal that a participant is ready to receive agent speech.

For example, participant.attributes['readyForAgent'] = true

For a regular participant, we could set this when the user has subscribed to the agent's audio
For SIP participant, it could be set once the call is established.

davidzhao:
@dennwc now that we have an event to notify publisher if their track has been subscribed to. I think on the SIP side we should do the following:
1. SIP joins room early, but do not subscribe to any tracks
2. when phone picks up and negotiated, then subscribe to audio tracks in the room
3. the agent will detect that their track has been subscribed to, and will start interactions then

dennwc:
[...]

PR #165, opened 2024-09-02
Change the call flow. Defer track subscription.

### LINKED ISSUE #102: Add early SIP dialog state
Hello!
At the moment, when receiving a call, the Livekit SIP gateway answers immediately and only after 4-5 seconds the media stream is switched on. 
Is it possible to enable the progress state (183/SDP) so that the customer can hear the ringing until the media stream is generated?

dennwc:
Refactor SIP signaling an change the call flow for both inbound and outbound.

For inbound **with PIN** (almost the same as before):
- Keep sending Ringing status for the whole time until the call is accepted.
- Signal to SIP peer as soon as we can and run the audio for the pin prompts.
- Given the PIN, join selected room on LK server.
- Immediate publish SIP track and subscribe to the rest.

For inbound **without PIN**:
- Keep sending Ringing status for the whole time until the call is accepted.
- First connect to the LK room according to dispatch rules.
- Subscribe and wait for at least one audio track. This is important! SIP now **keeps ringing** until there's one other participant with enabled audio track.
- Accept the call on SIP side. Wait for the first media packet.
- Publish SIP track to LiveKit.

For outbound:
- Flow as before, but delay track subscription until media pipeline is initialized.

Should resolve #102, #136, #114

### DIFF
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -16,8 +16,10 @@ package sip
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"slices"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -41,14 +43,6 @@ const (
 	// This is done because of audio cutoff at the beginning of calls observed in the wild.
 	audioBridgeMaxDelay = 1 * time.Second
 )
-
-func (s *Server) sipErrorOrDrop(tx sip.ServerTransaction, req *sip.Request) {
-	if s.conf.HideInboundPort {
-		tx.Terminate()
-	} else {
-		sipErrorResponse(tx, req)
-	}
-}
 
 func (s *Server) handleInviteAuth(log logger.Logger, req *sip.Request, tx sip.ServerTransaction, from, username, password string) (ok bool) {
 	if username == "" || password == "" {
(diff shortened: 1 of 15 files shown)


pkg/sip/inbound.go
```diff
@@ -297,102 +273,82 @@ func (c *inboundCall) handleInvite(ctx context.Context, req *sip.Request, tx sip
...
-	switch disp.Result {
-	default:
-		c.log.Errorw("Rejecting inbound call", fmt.Errorf("unreachable dispatch result path: %v", disp.Result))
-		sipErrorResponse(tx, req)
-		c.close(true, callDropped, "unreachable-path")
-		return
-	case DispatchRequestPin:
+	if pinPrompt {
+		// Accept the call first on the SIP side, so that we can send audio prompts.
+		if !acceptCall() {
```
biglittlebigben, line 324: Is there a code path where acceptCall may be nil?

pkg/sip/inbound.go
```diff
@@ -660,3 +621,220 @@ func (c *inboundCall) handleDTMF(tone dtmf.Event) {
...
+	c.respond(180, "Ringing")
+}
+
+func (c *sipInbound) attachTag() {
+	// Set the SIP tag for following requests from us to remote (e.g. BYE).
+	c.to.Params["tag"] = string(c.id)
+}
+
+func (c *sipInbound) StartRinging() {
+	c.attachTag()
```
biglittlebigben, line 749: Will StartRinging and StopRinging always be called from the same goroutine, since c.ringing is not thread safe?

pkg/sip/outbound.go
```diff
@@ -328,196 +335,215 @@ func (c *outboundCall) sipSignal(conf sipOutboundConfig) error {
...
 		authHeader = cred.String()
 		// Try again with a computed digest
 	}
-}
 
-func (c *outboundCall) sipAccept(inviteReq *sip.Request, inviteResp *sip.Response) error {
-	tag, err := getToTag(inviteResp)
-	if err != nil {
-		return err
+	c.invite, c.inviteOk = req, resp
```
biglittlebigben, line 473: Are c.invite and c.inviteOk always be accessed from the same goroutine as the one Invite is called on?

biglittlebigben (approved):
LGTM. Couple of questions inline.

davidzhao:
awesome improvement!

For the flow without PIN, I'm wondering if we can move publishing earlier:
* Publish SIP track to LiveKit.
* Subscribe and wait for at least one audio track. This is important! SIP now keeps ringing until there's one other participant with enabled audio track.
* Accept the call on SIP side. Wait for the first media packet.

If we accepted the SIP call, but the track isn't published yet, it means if the caller speaks immediately, their first words won't be heard by the other side.

pkg/sip/inbound.go
dennwc: No, it's always set. Just a shortcut.

pkg/sip/inbound.go
dennwc: Although it's not explicitly mentioned, they should run under the same call mutex. Most signaling is done by a single thread, except the calls to `Close`, but they would still acquire the same mutex.

Having said that, I'll check it again and will probably add mutex specifically for the signaling part of the connection. We will likely reuse it for our test client later.

pkg/sip/outbound.go
dennwc: Should be roughly the same as in the other comment thread. So it should hold the call mutex while doing so. But will double-check.

dennwc:
@davidzhao We can do that, but I'm a bit worried that there will be no explicit signal when finally accepting the call in that case.

In the flow that you propose, as soon as SIP subscribes to at least one track, it will confirm the call via SIP signaling. There will be some delay until we actually receive audio packets. SIP won't emit silence automatically in this case. But maybe it's acceptable?

pkg/sip/inbound.go
dennwc: Done, everything is protected now. I should actually move it to a separate package later.

MFaiqKhan:
Hi @dennwc currently participants.attribute.get(sip.callstatus) shows ringing and then after ringing even if the user hasn't picked up the call and shows active in the call status , user hasn't picked up the call and call has been turned off but the logs shows that user has picked up and active status on sip participant.

dennwc:
@MFaiqKhan I assume you mean outbound calls in this case. It could be that the server you are dialing is proxying the call to another server and indicates that the call is dialed even before the actual user picks up. From LiveKit's perspective the call is dialed and receives audio from the remote server. We'll have to add dial tone detection to know if the dial tone is still going in the audio stream.

faiqkhan-aiconnect:
@dennwc 
these are my logs currently that is from voice cloud livekit :

here the logs shows that agent left because the phone after ringing turned off but exactly the ringing stops the agent left and the particpant is autoamtically joined when they even havent picked up the call . 

and are you saying that twilio (sip provider), is proxying the call to the different server ???
Because the server that showed in the terminal logs were of singapore but the user server livekit shows of UAE .
what could be the issue here ?


phone_user_+61 | (none) | 1/22/2025, 7:57 AM | UAE |   | 0 | 23s


Events

—	ROOM_ENDED	{"event":"ROOM_ENDED","time_stamp":"2025-01-22T02:58:23.754058Z"}	
1/22/2025, 7:58 AM

phone_user_+61	PARTICIPANT_LEFT	{"participant_id":"PA_MdKpEkJDdpP9","participant_identity":"phone_user_+61","event":"PARTICIPANT_LEFT","time_stamp":"2025-01-22T02:57:43.185201Z"}	
1/22/2025, 7:57 AM

phone_user_+	TRACK_UNPUBLISHED	{"track_source":"MICROPHONE","track_id":"TR_AMadZe7hhsnZke","participant_id":"PA_MdKpEkJDdpP9","event":"TRACK_UNPUBLISHED","time_stamp":"2025-01-22T02:57:43.185138Z"}	
1/22/2025, 7:57 AM

phone_user_+61	TRACK_PUBLISHED	{"track_source":"MICROPHONE","track_id":"TR_AMadZe7hhsnZke","participant_id":"PA_MdKpEkJDdpP9","participant_identity":"phone_user_+61","event":"TRACK_PUBLISHED","time_stamp":"2025-01-22T02:57:22.321434Z"}	
1/22/2025, 7:57 AM

agent-AJ_K5vLsMqKEhmZ	PARTICIPANT_LEFT	{"participant_id":"PA_FEd9M5EJNjiY","participant_identity":"agent-AJ_K5vLsMqKEhmZ","event":"PARTICIPANT_LEFT","time_stamp":"2025-01-22T02:57:21.482699Z"}	
1/22/2025, 7:57 AM

phone_user_+61	PARTICIPANT_ACTIVE	{"participant_id":"PA_MdKpEkJDdpP9","participant_identity":"phone_user_+61","event":"PARTICIPANT_ACTIVE","time_stamp":"2025-01-22T02:57:20.707548Z"}	
1/22/2025, 7:57 AM

phone_user_+61	PARTICIPANT_JOINED	{"participant_id":"PA_MdKpEkJDdpP9","participant_identity":"phone_user_+61","event":"PARTICIPANT_JOINED","time_stamp":"2025-01-22T02:57:20.549481Z"}	
1/22/2025, 7:57 AM

agent-AJ_K5vLsMqKEhmZ	PARTICIPANT_ACTIVE	{"participant_id":"PA_FEd9M5EJNjiY","participant_identity":"agent-AJ_K5vLsMqKEhmZ","event":"PARTICIPANT_ACTIVE","time_stamp":"2025-01-22T02:57:19.252151Z"}	
1/22/2025, 7:57 AM

agent-AJ_K5vLsMqKEhmZ	PARTICIPANT_JOINED	{"participant_id":"PA_FEd9M5EJNjiY","participant_identity":"agent-AJ_K5vLsMqKEhmZ","event":"PARTICIPANT_JOINED","time_stamp":"2025-01-22T02:57:18.433368Z"}	
1/22/2025, 7:57 AM

—	ROOM_CREATED	{"event":"ROOM_CREATED","time_stamp":"2025-01-22T02:57:16Z"}	
1/22/2025, 7:57 AM

dennwc:
@faiqkhan-aiconnect could you please file a new issue instead with all these details? Please also specify either the callID or the sequence of events as you observed it.

serhiynovos:
It would be nice in dispatch rules to have it as strategy to accept immediately or wait first track publication or even have ability to answer the call  programmatically .

 In my case I have to broadcast incoming call to other devices which means I have only one way communication

PR #181, opened 2024-09-19
Fix resampler for 8K sample rate. Add RTP audio debug.

dennwc:


pkg/media/resample_soxr.go
```diff
@@ -42,7 +40,7 @@ func resampleBuffer(dst PCM16Sample, dstSampleRate int, src PCM16Sample, srcSamp
 	inbuf := make([]byte, len(src)*2)
 	outbuf := bytes.NewBuffer(nil)
 	outbuf.Grow(resampleSize(dstSampleRate, srcSampleRate, len(src)))
-	r, err := resample.New(outbuf, float64(srcSampleRate), float64(dstSampleRate), 1, resample.I16, quality)
+	r, err := resample.New(outbuf, float64(srcSampleRate), float64(dstSampleRate), 1, resample.I16, resample.Quick)
```
davidzhao, line 43: what does quick mean?

davidzhao (approved):

pkg/media/resample_soxr.go
dennwc: It's just the fastest mode that Soxr supports. It adds 0 latency and works well for 48K->16K resampling, but not for 8K. We [used](https://github.com/livekit/sip/pull/181/commits/b5ab53c293076a7eed792f9949ccd78835bb8efd#diff-d4dc0e02b583f7b607c93e6054d252ffe38f581ee44cf2d24b731e35a4177953L28) it for everything previously.

If you wonder about that code path specifically, it only triggers for our own pin prompts and nothing else.

PR #182, opened 2024-09-19
Implement TransferParticipant using SIP REFER messages

biglittlebigben:


pkg/sip/client.go
```diff
@@ -219,7 +230,39 @@ func (c *Client) onBye(req *sip.Request, tx sip.ServerTransaction) bool {
...
+func (c *Client) onNotify(req *sip.Request, tx sip.ServerTransaction) bool {
+	tag, _ := getFromTag(req)
+	c.cmu.Lock()
+	call := c.byRemote[tag]
+	c.cmu.Unlock()
+	if call == nil {
+		return false
+	}
+	call.log.Infow("NOTIFY")
+	go func(call *outboundCall) {
```
dennwc:
```suggestion
	go func() {
```
I think it's only necessary for loops to pass it that way. Should be fine without it.

pkg/sip/client.go
```diff
@@ -219,7 +230,39 @@ func (c *Client) onBye(req *sip.Request, tx sip.ServerTransaction) bool {
...
+		err := call.handleNotify(req, tx)
+
+		var code sip.StatusCode = 200
+		var psrpcErr psrpc.Error
+		if errors.As(err, &psrpcErr) {
+			code = sip.StatusCode(psrpcErr.ToHttp())
```
dennwc: The mapping between HTTP and SIP statuses is not exactly 1:1. Maybe add a function that converts the most common ones?

pkg/sip/inbound.go
```diff
@@ -223,6 +224,46 @@ func (s *Server) onBye(req *sip.Request, tx sip.ServerTransaction) {
...
+			code = sip.StatusCode(psrpcErr.ToHttp())
+		} else if err != nil {
+			code = 500
+		}
+
+		msg := "success"
+		if err != nil {
+			msg = err.Error()
+		}
+		tx.Respond(sip.NewResponseFromRequest(req, code, msg, nil))
```
dennwc: Looks like this part can be moved to a helper function, since it's the same for both server and client.

pkg/sip/inbound.go
```diff
@@ -631,6 +679,15 @@ func (c *inboundCall) createLiveKitParticipant(ctx context.Context, rconf RoomCo
 		return ctx.Err()
 	default:
 	}
+
+	sipCallID := partConf.Attributes[livekit.AttrSIPCallID]
```
dennwc: Lets pass it explicitly in function arguments then. Just in case we change order in which attributes are set.

pkg/sip/inbound.go
```diff
@@ -696,6 +753,71 @@ func (c *inboundCall) handleDTMF(tone dtmf.Event) {
...
+		return psrpc.NewErrorf(psrpc.DeadlineExceeded, "refer timed out")
+	case <-ctx.Done():
+		return psrpc.NewErrorf(psrpc.Canceled, "refer canceled")
+	case err := <-c.referDone:
+		if err != nil {
+			return err
+		}
+	}
+
+	c.cc.sendBye()
```
dennwc: `c.Close()` below will call it as well, right?

pkg/sip/inbound.go
```diff
@@ -696,6 +753,71 @@ func (c *inboundCall) handleDTMF(tone dtmf.Event) {
...
+
+	switch method {
+	case sip.REFER:
+		c.cc.mu.RLock()
+		defer c.cc.mu.RUnlock()
+
+		if cseq != 0 && cseq != uint32(c.cc.referCseq) {
+			// NOTIFY for a different REFER, skip
+			return nil
+		}
```
dennwc: Can we move this into `sipInbound` type (aka the signaling part)? I was planning to move it into a separate package to share with the test client.

pkg/sip/protocol.go
```diff
@@ -59,3 +63,132 @@ func sendAndACK(c Signaling, req *sip.Request) {
...
+	}
+	defer tx.Terminate()
+
+	resp, err := sipResponse(tx, stop)
+	if err != nil {
+		return nil, err
+	}
+
+	switch resp.StatusCode {
+	case 200, 202:
```
dennwc: Maybe use typed constants from sip library? I don't remember what 202 means in SIP, for example.

pkg/sip/client.go
```diff
@@ -219,7 +230,39 @@ func (c *Client) onBye(req *sip.Request, tx sip.ServerTransaction) bool {
...
+		} else if err != nil {
+			code = 500
+		}
+
+		msg := "success"
+		if err != nil {
+			msg = err.Error()
+		}
+		tx.Respond(sip.NewResponseFromRequest(req, code, msg, nil))
```
dennwc, line 245: I wonder if there's any case where we don't want to return 200 for `NOTIFY` :thinking:

pkg/sip/inbound.go
```diff
@@ -696,6 +753,71 @@ func (c *inboundCall) handleDTMF(tone dtmf.Event) {
...
 }
 
+func (c *inboundCall) transferCall(ctx context.Context, transferTo string) error {
+	err := c.cc.transferCall(ctx, transferTo)
+	if err != nil {
+		return err
+	}
+
+	select {
+	case <-time.After(30 * time.Second):
```
dennwc: This could be a regular timeout set on the context.

pkg/sip/protocol.go
```diff
@@ -59,3 +63,132 @@ func sendAndACK(c Signaling, req *sip.Request) {
...
+
+	return c, nil
+}
+
+func handleNotify(req *sip.Request) (method sip.RequestMethod, cseq uint32, status int, err error) {
+	event := req.GetHeader("Event")
+	var cseq64 uint64
+
+	switch {
+	case strings.HasPrefix(strings.ToLower(event.Value()), "refer"):
```
dennwc: Parsing part will need a test, ideally.

pkg/sip/inbound.go
```diff
@@ -696,6 +753,71 @@ func (c *inboundCall) handleDTMF(tone dtmf.Event) {
 	}
 }
 
+func (c *inboundCall) transferCall(ctx context.Context, transferTo string) error {
+	err := c.cc.transferCall(ctx, transferTo)
+	if err != nil {
+		return err
+	}
+
+	select {
```
dennwc: Could we move this deeper into the call stack? Current function calls `c.cc.transferCall` which calls `sendRefer`->`sipResponse`, which in turn will `select` on a SIP transaction. So it could also wait on the context and maybe on `referDone` too? The reason I'm proposing this is that tx may also accept `CANCEL` requests, which is also exposed as a channel. You will likely need to select on that eventually too, together with `referDone`.

pkg/sip/protocol.go
```diff
@@ -59,3 +63,132 @@ func sendAndACK(c Signaling, req *sip.Request) {
...
+	}
+
+	if h, _ := inviteRequest.CallID(); h != nil {
+		sip.CopyHeaders("Call-ID", inviteRequest, req)
+	}
+
+	if h, _ := inviteRequest.CSeq(); h != nil {
+		sip.CopyHeaders("CSeq", inviteRequest, req)
+		// Increase the invite request cseq so that we use the proper cseq in the bye request eventually
+		h.SeqNo = h.SeqNo + 1
```
dennwc: This part is interesting. I assume BYE will also +1 it and will get the same CSeq as the REFER, right? Maybe we could set a correct CSeq on the BYE, without mutating the original invite?

pkg/sip/inbound.go
biglittlebigben: Sure. For reference, what is the slip between sipInbound and inboundCall? I put this into inboundCall because this is part of the call workflow, which seemed usually handled by inboundCall (vs signaling and call parameters in sipInbound)

pkg/sip/client.go
biglittlebigben: I thinks so. Currently, we'll answer 400 in the case where we can't parse the request, and I think that's appropriate.

pkg/sip/inbound.go
dennwc: My thinking was that all SIP protocol details should be hidden in `sipInbound`. That way we could test it separately and eventually update sipgo package to the latest, all without changing the `inboundCall` logic.

pkg/sip/inbound.go
biglittlebigben: I moved this one step down into the stack into sipInbound. I believe this belong to a transaction level above `sipResponse`, as we are waiting for the a totally separate request (vs STATUS responses). We should handle CANCEL by sending an appropriate error the referDone channel, I believe.

pkg/sip/protocol.go
biglittlebigben: I changed it overwrite the CSeq of all requests we send out using a counter in the object. It's a bit more clear, but more code. Ideally, we'd fix sipgo to stop making assumptions.

dennwc (approved):

PR #185, opened 2024-09-26
Add inbound media processors

frostbyte73:


pkg/media/processor.go
```diff
@@ -0,0 +1,22 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package media
+
+type Processor[T any] interface {
+	SetWriter(writer WriteCloser[T])
+	WriteCloser[T]
+}
```
dennwc: Maybe this would be a bit cleaner?
```suggestion
type Processor[T any] func(w WriteCloser[T]) WriteCloser[T]
```

pkg/sip/server.go
```diff
@@ -88,11 +91,13 @@ type CallDispatch struct {
 	DispatchRuleID      string
 	Headers             map[string]string
 	HeadersToAttributes map[string]string
+	EnabledFeatures     []rpc.SIPFeature
 }
 
 type Handler interface {
 	GetAuthCredentials(ctx context.Context, callID, fromUser, toUser, toHost, srcAddress string) (AuthInfo, error)
 	DispatchCall(ctx context.Context, info *CallInfo) CallDispatch
+	GetInboundProcessor(cd CallDispatch) media.PCM16Processor
```
```suggestion
	GetMediaProcessor(features []rpc.SIPFeature) media.PCM16Processor
```
I believe media processor won't be inbound-specific, so no need to bind it to inbound-only `CallDispatch`.

pkg/sip/inbound.go
```diff
@@ -260,6 +261,7 @@ type inboundCall struct {
 	cc          *sipInbound
 	mon         *stats.CallMonitor
 	extraAttrs  map[string]string
+	handler     Handler
```
dennwc: It already has access to it via `c.s.handler`.

pkg/sip/inbound.go
```diff
@@ -419,7 +430,7 @@ func (c *inboundCall) handleInvite(ctx context.Context, req *sip.Request, trunkI
 	}
 }
 
-func (c *inboundCall) runMediaConn(offerData []byte, conf *config.Config) (answerData []byte, _ error) {
+func (c *inboundCall) runMediaConn(offerData []byte, conf *config.Config, disp CallDispatch) (answerData []byte, _ error) {
```
dennwc: Could just pass the `Processor`, no need for a full dispatch info here.

dennwc (approved):
Looks good! You could add it now for outbound pretty easily, I think. Or could be another PR.

PR #186, opened 2024-09-26
Transfer participant integration test

biglittlebigben:
This also makes sure that we do not cancel the transfer request if the RPC handler gets cancelled, and avoids situation where we may return an old transfer result to a 2nd transfer request if one failed before.


pkg/siptest/client.go
```diff
@@ -381,6 +400,96 @@ func (c *Client) SendDTMF(digits string) error {
...
 
+func (c *Client) SendNotify(eventReq *sip.Request, notifyStatus string) error {
+	var recipient *sip.Uri
+
+	if contact, ok := eventReq.Contact(); ok {
+		recipient = &contact.Address
+	} else if from, ok := eventReq.From(); ok {
+		recipient = &from.Address
+	} else {
+		errors.New("missing destination address")
```
dennwc: Missing return?

test/integration/sip_test.go
```diff
@@ -376,10 +390,53 @@ func TestSIPJoinOpenRoom(t *testing.T) {
 		return dtmfIn == "4567"
 	}, 5*time.Second, time.Second/2)
 
+	go func() {
+		// TransferSIPParticipant is synchronous
+		_, err = lk.SIP.TransferSIPParticipant(context.Background(), &livekit.TransferSIPParticipantRequest{
+			RoomName:            "test-open",
```
dennwc:
```suggestion
			RoomName:            roomName,
```

test/integration/sip_test.go
```diff
@@ -376,10 +390,53 @@ func TestSIPJoinOpenRoom(t *testing.T) {
...
+	transferTo := referRequest.GetHeader("Refer-To")
+	require.Equal(t, "tel:"+transferNumber, transferTo.Value())
+
+	time.Sleep(notifyIntervalDelay)
+	err = cli.SendNotify(referRequest, "SIP/2.0 100 Trying")
+	require.NoError(t, err)
+
+	time.Sleep(notifyIntervalDelay)
+	err = cli.SendNotify(referRequest, "SIP/2.0 200 OK")
+	require.NoError(t, err)
```
dennwc, line 422: This part could be in `onRefer` callback, then there will be no need for a mutex and shared variable (`referRequest`). Same for the second test.

test/integration/sip_test.go
```diff
@@ -376,10 +390,53 @@ func TestSIPJoinOpenRoom(t *testing.T) {
 		return dtmfIn == "4567"
 	}, 5*time.Second, time.Second/2)
 
+	go func() {
+		// TransferSIPParticipant is synchronous
+		_, err = lk.SIP.TransferSIPParticipant(context.Background(), &livekit.TransferSIPParticipantRequest{
+			RoomName:            "test-open",
+			ParticipantIdentity: "sip_" + clientNumber,
+			TransferTo:          "tel:" + transferNumber,
```
dennwc, line 398: Would be nice to also see if we can properly refer for outbound. But that can be done separately.

pkg/sip/inbound.go
```diff
@@ -1094,14 +1096,14 @@ func (c *sipInbound) handleNotify(req *sip.Request, tx sip.ServerTransaction) er
 			// Success
 			select {
 			case c.referDone <- nil:
-			default:
+			case <-time.After(5 * time.Second):
```
dennwc: Worth adding a global constant? Same for outbound.

dennwc (approved):

PR #195, opened 2024-10-09
Ignore repeated TransferSIPParticipant RPC request with same parameters

biglittlebigben:


pkg/sip/service.go
```diff
@@ -127,6 +137,25 @@ func (s *Service) TransferSIPParticipant(ctx context.Context, req *rpc.InternalT
...
 
+	s.mu.Lock()
+	k := transferKey{
+		SipCallId:  req.SipCallId,
+		TransferTo: req.TransferTo,
+	}
+	if _, ok := s.pendingTransfers[k]; ok {
+		s.mu.Unlock()
+		s.log.Debugw("repeated request for call transfer", "callID", req.SipCallId, "transferTo", req.TransferTo)
+		return &emptypb.Empty{}, nil
```
dennwc: I'm not sure what are the semantics here for PSRPC. Will it unblock earlier if the second request immediately succeeds by going this code path? Maybe use `map[transferKey]<-chan struct{}` and wait here for the channel to close?

Also, you intentionally remove cancellation from the original context, which is desired for the main call. But maybe it's worth preserving the original context in a separate var, so that you can then select on this new done channel + the context. I assume PSRPC will cancel the request after a timeout. So at least the follow-up requests will be unblocked.

dennwc (approved):

PR #196, opened 2024-10-10
Always set contact header for outbound to IP.

dennwc:
Always set contact header for outbound to IP address instead of the hostname. Also, refactor the code to use `netip.Addr` and `netip.AddrPort`.


davidzhao (approved):
lg

PR #198, opened 2024-10-11
Use ringing timeout and max call duration.

dennwc:


pkg/sip/participant.go
```diff
@@ -14,7 +14,22 @@
...
-import "github.com/livekit/protocol/livekit"
+import (
+	"time"
+
+	"github.com/livekit/protocol/livekit"
+)
+
+const (
+	// defaultMaxCallDuration sets a default max call duration. Value is according to Twilio's default.
+	defaultMaxCallDuration = 4 * time.Hour
```
davidzhao: while I think this is fine, I feel a bit weird about enforcing a default timeout on our side, when the trunking provider is already doing their own.

it seems that it could potentially confuse users where the 4h timeout is coming from. wdyt about either:
* increasing our default to something much greater: i.e. 24h (and ensure it's documented in docs)
* do not default to timing out

pkg/sip/outbound.go
```diff
@@ -330,14 +340,20 @@ func (c *outboundCall) setStatus(v CallStatus) {
...
-		AttrSIPCallStatus: attr,
+		livekit.AttrSIPCallStatus: attr,
 	})
 }
 
 func (c *outboundCall) sipSignal(ctx context.Context) error {
 	ctx, span := tracer.Start(ctx, "outboundCall.sipSignal")
 	defer span.End()
 
+	if c.sipConf.ringingTimeout > 0 {
```
davidzhao, line 351: good for safety, but it seems that this is impossible given the defaults that we are setting.

pkg/sip/participant.go
dennwc: > while I think this is fine, I feel a bit weird about enforcing a default timeout on our side, when the trunking provider is already doing their own.

This is mostly to protect us from wasting resources on misconfigured Cloud projects. While Twilio has the limit, other providers or self-hosted SIP instances may not have this limit.

I propose to raise it to 6h or 8h so that it's still not that long but is still distinguishable from Twilio's timeout. Note that it's not a hard limit - user can set it to a higher value, if needed.

pkg/sip/outbound.go
dennwc: I'd prefer to keep it for safety, if you don't mind. In case the timeout is removed later, this won't break at least.

davidzhao: yup for sure!

pkg/sip/participant.go
davidzhao: if it's to protect ourselves, it seems that we should not allow the user to override it. (i.e. system-set upper limit).

perhaps these are two configurations?
* system-wide upper limit (defaults to 8h, but changeable in config)
* user-specified timeout (defaults to not limited)

dennwc: OK, I'll then cap the timeout at 24h max :+1:

davidzhao (approved):

PR #207, opened 2024-10-21
Allow requesting a ringtone during SIP transfers

biglittlebigben:


pkg/sip/inbound.go
```diff
@@ -766,8 +773,29 @@ func (c *inboundCall) handleDTMF(tone dtmf.Event) {
...
+	if ringtone && c.started.IsBroken() && !c.done.Load() {
+		const ringVolume = math.MaxInt16 / 2
+		rctx, rcancel := context.WithCancel(ctx)
+		defer rcancel()
+
+		// mute the room audio to the SIP participant
+		w := c.lkRoom.SwapOutput(nil)
+
+		defer func() {
+			if err != nil && !c.done.Load() {
```
dennwc: Ideally this should check an actual return var of the function. I usually declare it with a different name to make sure it's the right one:

```go
func (c *inboundCall) transferCall(...) (gerr error) {
	// ...
	defer func() {
		if gerr != nil {
			// ...
		}
	}()
```

pkg/sip/inbound.go
```diff
@@ -766,8 +773,29 @@ func (c *inboundCall) handleDTMF(tone dtmf.Event) {
...
+func (c *inboundCall) transferCall(ctx context.Context, transferTo string, ringtone bool) error {
+	var err error
+
+	if ringtone && c.started.IsBroken() && !c.done.Load() {
+		const ringVolume = math.MaxInt16 / 2
+		rctx, rcancel := context.WithCancel(ctx)
+		defer rcancel()
+
+		// mute the room audio to the SIP participant
+		w := c.lkRoom.SwapOutput(nil)
```
dennwc, line 785: Since `w` may contain C resampler, we need to make sure it's either set back via `SwapOutput`, or closed if it's no longer needed.

dennwc: Sorry for confusing, but this still is not closed. The `c.media.GetAudioWriter()` is fine, it will be closed when `c.media` is closed. The problem is with `w` -  you detach it from the room, so it won't be closed if the transfer succeeds. It should probably be in the `defer` below - it either swaps `w` back to the room, or should close it.

dennwc (approved):

PR #210, opened 2024-10-24
TLS support

### LINKED ISSUE #71: TCP/TLS support for SIP signaling
It _appears_ that the SIP adapter/service is currently hard coded to use UDP for SIP signaling. However, it appears the upstream library has support for TCP and TLS.

https://github.com/livekit/sip/blob/1ce2ec84ebf08f23d97febddb507f63d4cc21c6d/pkg/sip/server.go#L172

Would there be any aversion to adding _optional_ configuration options to support other SIP signaling transports such as TCP and/or TLS?

dennwc:
Add TLS support for signaling. Resolves #71.

### DIFF
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -131,7 +131,8 @@ func (s *Server) onInvite(req *sip.Request, tx sip.ServerTransaction) {
 		"toIP", req.Destination(),
 	)
 
-	cc := s.newInbound(LocalTag(callID), req, tx)
+	tr := transportFromReq(req)
+	cc := s.newInbound(LocalTag(callID), s.ContactURI(tr), req, tx)
 	log = LoggerWithParams(log, cc)
 	log = LoggerWithHeaders(log, cc)
 	log.Infow("processing invite")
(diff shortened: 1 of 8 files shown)


pkg/sip/inbound.go
```diff
@@ -1092,9 +1100,7 @@ func (c *sipInbound) newReferReq(transferTo string) (*sip.Request, error) {
 	}
 
 	// This will effectively redirect future SIP requests to this server instance (if host address is not LB).
-	contactHeader := &sip.ContactHeader{Address: sip.Uri{Host: c.s.signalingIp.String(), Port: c.s.conf.SIPPort}}
-
-	req := NewReferRequest(c.invite, c.inviteOk, contactHeader, transferTo)
+	req := NewReferRequest(c.invite, c.inviteOk, c.contact, transferTo)
```
biglittlebigben, line 1103: Thanks! :)

pkg/sip/participant.go
```diff
@@ -75,6 +75,7 @@ func (v CallStatus) DisconnectReason() livekit.DisconnectReason {
 
 const (
 	callDropped = CallStatus(iota)
+	callFlood
```
biglittlebigben, line 78: This constant should not be exported?

biglittlebigben (approved):

pkg/sip/participant.go
dennwc: We don't distinguish it in participant attributes, so it's unexported.

PR #215, opened 2024-10-28
Revert "Set the expected room name in ListRooms requests in the integration test (#214)"

biglittlebigben:
This reverts commit 1266354ea066e80c32dd214f5734570e075cae39.


biglittlebigben:
This is not needed after all

davidzhao (approved):

PR #216, opened 2024-10-28
Send analytics events

biglittlebigben:


pkg/sip/types.go
```diff
@@ -60,15 +74,38 @@ const (
 type URI struct {
 	User      string
 	Host      string
+	Port      uint16
```
dennwc: Why adding a separate port, if it's already in `Addr`?

pkg/sip/types.go
```diff
@@ -140,6 +180,20 @@ func (u URI) GetContactURI() *sip.Uri {
 	return su
 }
 
+func (u URI) ToSIPUri() *livekit.SIPUri {
+	url := &livekit.SIPUri{
+		User:      u.User,
+		Host:      u.GetHost(),
+		Port:      fmt.Sprintf("%d", u.GetPort()),
```
dennwc: This will change after updating to https://github.com/livekit/protocol/pull/869

pkg/sip/types.go
biglittlebigben: If the Host field ends up containing a hostname and port, not an ip address, the Addr argument will be invalid (the preexisting line `netip.AddrPortFrom(u.Addr.Addr(), uint16(port))` will return "Invalid AddPort", with no port field if `u.Addr.Addr()` is invalid).

pkg/sip/types.go
biglittlebigben: Sure. I'll update

pkg/sip/types.go
dennwc: I also thought that it's the case, but it will still work if we test port separately from the address. `addrport.IsValid()` always returns `addrport.Addr().IsValid()`.

But if you think it's inconvenient to use that separately, we could change the field type to `netip.Addr` and add a `AddrPort/SetAddrPort` helper methods.

dennwc (approved):

PR #218, opened 2024-10-31
Implement Via headers for requests for outbound calls

biglittlebigben:
For outbound calls, when making requests (BYE/REFER), the Via header should have the IP address/port of the instance sending the request.

We use the private IP because the Transaction logic in sipgo uses this information to bind the socket when making establishing the connection. This should work for UDP, but this will still fail for TCP.

### DIFF
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -1064,10 +1064,27 @@ func (c *sipInbound) swapSrcDst(req *sip.Request) {
 	req.AppendHeader((*sip.FromHeader)(c.to))
 	req.RemoveHeader("To")
 	req.AppendHeader((*sip.ToHeader)(c.from))
+	req.RemoveHeader("Via")
+	req.PrependHeader(c.generateViaHeader(req))
 	if route, ok := req.RecordRoute(); ok {
 		req.RemoveHeader("Record-Route")
 		req.AppendHeader(&sip.RouteHeader{Address: route.Address})
 	}
+}
+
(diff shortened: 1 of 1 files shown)


pkg/sip/inbound.go
```diff
@@ -1145,12 +1145,31 @@ func (c *sipInbound) swapSrcDst(req *sip.Request) {
 	req.AppendHeader((*sip.FromHeader)(c.to))
 	req.RemoveHeader("To")
 	req.AppendHeader((*sip.ToHeader)(c.from))
+	// Remove all Via headers
+	for req.RemoveHeader("Via") {
+	}
```
dennwc, line 1150: If I understand correctly, this might be incorrect. Each router is supposed to append it's own Via and then remove that line when requests go back. So I believe we are supposed to add own own line and that's it.

biglittlebigben: The behavior you state above is correct for responses within a given transaction: the responses must have the same VIA headers as the request, and proxies on the way back are supposed to "peel" their own Via entries from the response,. The REFER request is however a new request, in a new transaction, sent directly to the server from the INVITE Contact header. As such, only the IP address of the request initiator should be in the Via header. I believe the examples in the REFER RFC confirm the above: https://www.rfc-editor.org/rfc/rfc3515#section-4

dennwc: Ah, you are right, I forgot that's a new request. Makes sense :+1:

dennwc (approved):

PR #229, opened 2024-11-19
Disable call ID reuse across auth attempts

biglittlebigben:


biglittlebigben:
This fixes trunk auth with Telnyx

PR #232, opened 2024-11-25
Support SRTP

dennwc:
Supporting protocol PR:
- https://github.com/livekit/protocol/pull/892

### TODOs:
- [x] Support media timeout.

### DIFF
--- a/pkg/media/rtp/conn.go
+++ b/pkg/media/rtp/conn.go
@@ -16,17 +16,21 @@ package rtp
 
 import (
 	"net"
+	"net/netip"
 	"sync"
 	"sync/atomic"
 	"time"
 
 	"github.com/frostbyte73/core"
 	"github.com/pion/rtp"
+
+	"github.com/livekit/protocol/logger"
 )
 
 var _ Writer = (*Conn)(nil)
 
 type ConnConfig struct {
+	Log                 logger.Logger
 	MediaTimeoutInitial time.Duration
 	MediaTimeout        time.Duration
 	TimeoutCallback     func()
@@ -40,6 +44,9 @@ func NewConnWith(conn UDPConn, conf *ConnConfig) *Conn {
 	if conf == nil {
 		conf = &ConnConfig{}
 	}
+	if conf.Log == nil {
+		conf.Log = logger.GetLogger()
+	}
 	if conf.MediaTimeoutInitial <= 0 {
 		conf.MediaTimeoutInitial = 30 * time.Second
 	}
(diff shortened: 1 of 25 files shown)


pkg/media/rtp/session.go
```diff
@@ -0,0 +1,207 @@
...
+
+func (s *session) OpenWriteStream() (WriteStream, error) {
+	return s.w, nil
+}
+
+func (s *session) AcceptStream() (ReadStream, uint32, error) {
+	s.rmu.Lock()
+	defer s.rmu.Unlock()
+	for {
+		n, err := s.conn.Read(s.rbuf[:])
```
biglittlebigben, line 80: Is there a way to detect that the packet was > 1500 bytes and we only got a partial read? I get it'd be very uncommon, but returning a proper error at least if it happens may save us a lot of debugging time.

pkg/sip/media_port.go
```diff
@@ -184,30 +235,104 @@ func (p *MediaPort) SetOffer(offerData []byte) (*sdp.Answer, *MediaConf, error)
...
+			p.log.Infow("accepting media", "ssrc", ssrc)
+			go p.rtpReadLoop(r)
+		} else {
+			p.log.Warnw("ignoring media", nil, "ssrc", ssrc)
+		}
+	}
+}
+
+func (p *MediaPort) rtpReadLoop(r rtp.ReadStream) {
+	buf := make([]byte, 1500)
```
biglittlebigben: Can we also throw an error here if but overflow?

biglittlebigben (commented):
LGTM. I just wonder about hardcoding the max packet size to 1500. It should be ok in most cases, but could load to hard to debug issues.

pkg/media/rtp/session.go
dennwc: Done, will log an error (once), but will continue reading other packets, in case some of them are still smaller.

pkg/sip/media_port.go
dennwc: Done, same as above.

pkg/media/rtp/conn.go
```diff
@@ -153,13 +161,21 @@ func (c *Conn) ListenAndServe(portMin, portMax int, listenAddr string) error {
...
 	for {
 		n, srcAddr, err := conn.ReadFromUDP(buf)
 		if err != nil {
 			return
 		}
 		c.dest.Store(srcAddr)
+		if n > MTUSize {
+			if !overflow {
+				overflow = true
+				c.log.Errorw("RTP packet is larger than MTU limit", nil)
```
biglittlebigben, line 176: Do we want to alert on this? (sure sending invalid packet, potentially as an attempt to break our system)? If not, should this be a Warn or Info?

dennwc: I'd prefer to not pass broken packet. Go crashes are fine and easy to recover from/fix, but crashes from C code are not. We could specifically fuzz the system to (hopefully) detect those.

As for the error/warn, even though this case won't break the system, it will still cause audio glitches in a specific call. I'd like to know when it happens, especially when a fix is as easy as increasing the buffer size. So I'd like to keep it as an error here.

PR #244, opened 2024-12-11
Log failed call transfers as warnings.

dennwc:


biglittlebigben (commented):
This will log a waring every case the transferred to number doesn't pick up, for instance. What is the motivation?

dennwc:
I need at least some indication that transfer failed in an unexpected way. Will add a SIP status code check before logging to warning, then.

PR #257, opened 2025-01-11
Log more information in room callbacks

biglittlebigben:
Getting the roomID in the callback logs required the used of an extra utilities because the room SID is only available after join, while some callbacks may fire before Join returns. room.SID() also takes the room lock which is already held in the participant callbacks. This means that we can't call room.SID() in the participant callbacks.

### DIFF
--- a/pkg/sip/room.go
+++ b/pkg/sip/room.go
@@ -44,19 +44,18 @@ type ParticipantInfo struct {
 }
 
 type Room struct {
-	log            logger.Logger
-	roomLog        logger.Logger // deferred logger
-	roomLogResolve func()
-	room           *lksdk.Room
-	mix            *mixer.Mixer
-	out            *media.SwitchWriter
-	outDtmf        atomic.Pointer[dtmf.Writer]
-	p              ParticipantInfo
-	ready          core.Fuse
-	subscribe      atomic.Bool
-	subscribed     core.Fuse
-	stopped        core.Fuse
-	closed         core.Fuse
+	log        logger.Logger
+	roomLog    logger.Logger // deferred logger
+	room       *lksdk.Room
+	mix        *mixer.Mixer
+	out        *media.SwitchWriter
+	outDtmf    atomic.Pointer[dtmf.Writer]
+	p          ParticipantInfo
+	ready      core.Fuse
+	subscribe  atomic.Bool
+	subscribed core.Fuse
+	stopped    core.Fuse
+	closed     core.Fuse
 }
 
 type ParticipantConfig struct {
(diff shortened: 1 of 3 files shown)


biglittlebigben:
Looking at the test failure, JoinWIthToken causes the TrackPublished callback to be called. We do not have the roomID at this point. We can either make the roomID logging a "best effort" thing, when it's available, or bounce the callback to a different goroutine. I think the latter option would be safe here.

davidzhao (approved):

pkg/sip/room.go
```diff
@@ -316,7 +344,7 @@ func (r *Room) CloseWithReason(reason livekit.DisconnectReason) error {
 	if r == nil {
 		return nil
 	}
-	r.ready.Store(false)
+	r.ready = core.Fuse{}
```
dennwc: Linters (including `go vet`) will complain about copying mutexes here.

pkg/sip/room.go
```diff
@@ -126,6 +132,21 @@ func (r *Room) subscribeTo(pub *lksdk.RemoteTrackPublication, rp *lksdk.RemotePa
 	r.subscribed.Break()
 }
 
+func (r *Room) getRoomLogger() logger.Logger {
+	select {
+	case <-r.ready.Watch():
+	case <-r.stopped.Watch():
+	}
```
dennwc: Is this really necessary? If it gets stuck here, it would be pretty hard to debug...

Could we instead either:
- Log `SID()` at a later time, so that logs for that call will eventually have it.
- ...or make SDK not block on `SID()` call (I'd prefer this one).

biglittlebigben: I'd expect that removing the blocking on room.SID would break something as this doesn't seem like something that would be added randomly. It's also not trivial to audit project wide. The code would also still deadlock unless we still dispatch all log calls outside the participant callbacks as room.SID() attempts to take a lock that is already taken when the participant callback is called (these participant callbacks are overall not very safe right now)

It's about as easy to replace r.log with a version that includes the roomID once `ready` is true, when we know that room.SID() won't block. We can make logging the roomID best effort, but I'm afraid that it'll make the effort moot as we're debugging an issue that happens very shortly after join. I can still start there.

pkg/sip/room.go
```diff
@@ -75,6 +77,21 @@ type RoomConfig struct {
 func NewRoom(log logger.Logger) *Room {
 	r := &Room{log: log, out: media.NewSwitchWriter(RoomSampleRate)}
 	r.mix = mixer.NewMixer(r.out, rtp.DefFrameDur)
+
+	roomLog, resolve := log.WithDeferredValues()
+	r.roomLog = roomLog
```
dennwc, line 82: Should we just set `r.log` here then?

biglittlebigben: Anything logged to roomLog will not be displayed until the room join succeeds, or the session is closed. It felt like it may be helpful to get some of the session lifecycle logs earlier.

PR #279, opened 2025-02-11
Tiny typo in readme

umarniz:
Should mention the media ports 10000-20000 instead of 1000-2000 similar to rest of the docs: https://docs.livekit.io/home/self-hosting/sip-server/


biglittlebigben (approved):
Indeed. Thanks for catching this.

PR #282, opened 2025-02-14
fix: prevent log resolver panic

davidzhao:
If the room is closed quickly after connect, we could end up with a nil `r.room`.


real-danm (approved):

davidzhao:
@livekit/media-devs I'm merging this one. it seems like the right fix.

PR #305, opened 2025-02-26
Adopt logging utilities from protocol

biglittlebigben:


pkg/config/config.go
```diff
@@ -161,7 +162,7 @@ func (c *Config) InitLogger(values ...interface{}) error {
 	values = append(c.GetLoggerValues(), values...)
 	l := zl.WithValues(values...)
 	logger.SetLogger(l, c.ServiceName)
-	lksdk.SetLogger(l)
+	lksdk.SetLogger(medialogutils.NewOverrideLogger())
```
dennwc: This might not work as you would expect. SIP [overrides](https://github.com/livekit/sip/blob/main/pkg/sip/room.go#L262) the SDK logger for each call, so that WebRTC logs include call ID, etc.

dennwc (approved):

PR #323, opened 2025-03-28
make media and signaling possible use different ips

zusrut:
also reduce annoying ringings every second


pkg/sip/inbound.go
```diff
@@ -1088,8 +1088,8 @@ func (c *sipInbound) StartRinging() {
 	tx := c.inviteTx
 	cancels := tx.Cancels()
 	go func() {
-		// TODO: check spec for the exact interval
-		ticker := time.NewTicker(time.Second)
+		// minimum one per 60 seconds - for nat hairpin better more often like 15 sec
+		ticker := time.NewTicker(15 * time.Second)
```
dennwc: Lets leave it at 3 sec at least. Or add another config option.

zusrut: ok i will create separate PR for this changes

dennwc (approved):

PR #328, opened 2025-04-02
configurable ringing interval

zusrut:


pkg/config/config.go
```diff
@@ -61,17 +61,18 @@ type Config struct {
...
-	Logging           logger.Config       `yaml:"logging"`
-	ClusterID         string              `yaml:"cluster_id"` // cluster this instance belongs to
-	MaxCpuUtilization float64             `yaml:"max_cpu_utilization"`
+	HealthPort         int                 `yaml:"health_port"`
+	PrometheusPort     int                 `yaml:"prometheus_port"`
+	PProfPort          int                 `yaml:"pprof_port"`
+	SIPPort            int                 `yaml:"sip_port"`        // announced SIP signaling port
+	SIPPortListen      int                 `yaml:"sip_port_listen"` // SIP signaling port to listen on
+	SIPHostname        string              `yaml:"sip_hostname"`
+	SIPRingingInterval int                 `yaml:"sip_ringing_interval"` // from 1 sec up to 60 (default 1)
```
dennwc: Maybe change to `time.Duration`? I see that there's a clear limit, but it will be self-documented that way at least.

zusrut: changed to duration

dennwc (approved):

PR #344, opened 2025-04-30
Only send RTP after the call is fully established.

dennwc:


biglittlebigben (commented):
LGTM. What specific issue is this solving?

dennwc:
@biglittlebigben one of the customers mentioned that we start sending RTP while the call state is dialing which makes it confusing for a variety of reasons. For example, if the system is recording audio for a particular conversation, it would think that it started with a few second of silence. Call duration stats would also be confusing, etc.

boks1971 (approved):
Looks good, but don't have full context, approving to unblock, would be good to get another pair of 👀 later.

PR #345, opened 2025-04-30
Support media-level c= fallback to session-level connection in SDP parsing

hamidzr:
👋🏼  This patch addresses an issue in the media SDP parsing logic where session-level connection information (c=) was always used, even if a media-level c= was present. As I understand it media sections may override the session-level connection address; and if absent, they should inherit it.

Changes
- pkg/media/sdp/helpers.go
  - Updated GetAudioDest to first check audio.ConnectionInformation (media-level), then fall back to session.ConnectionInformation.
  - Preserves existing validity checks on ci.NetworkType and handles nil cases gracefully.
- pkg/media/sdp/offer.go
  - Existing logic now errors on invalid addresses instead of passing it along w/o errors.

#### Impact
- Downstream components relying on correct RTP addresses should no longer see spurious “invalid AddrPort” errors when media-level c= is present.
- In case of inavlid AddrPort there will be an explicit error in parsing media SDPs.
**In other words this means all SDPs parsed by this util will get an error if no media AddrPort is found: Is this a good assumption for media SDP parsing? We can omit this if we're not sure.**

#### Validation
- Added unit tests covering:
- SDP with only session-level c=
- SDP with both session- and media-level c= (media-level should win)
- SDP with no c= fields (should yield an invalid AddrPort)
- Invalid errors are bubbled up to the caller.

#### Links:
- https://datatracker.ietf.org/doc/html/rfc4566#section-5.7
- [RFC 4566 – Session Description Protocol](https://tools.ietf.org/html/rfc4566)
- [RFC 3605 – SDP Bandwidth Modifications for RTCP](https://tools.ietf.org/html/rfc3605)

### DIFF
--- a/pkg/media/sdp/helpers.go
+++ b/pkg/media/sdp/helpers.go
@@ -29,14 +29,22 @@ func GetAudio(s *sdp.SessionDescription) *sdp.MediaDescription {
 	return nil
 }
 
+// GetAudioDest returns the RTP dst address:port for an audio m=
+// it first uses media-level c=, then session-level c=.
 func GetAudioDest(s *sdp.SessionDescription, audio *sdp.MediaDescription) netip.AddrPort {
 	if audio == nil || s == nil {
 		return netip.AddrPort{}
 	}
-	ci := s.ConnectionInformation
+
+	// pick media-level c=; if absent, fall back to session-level c=
+	ci := audio.ConnectionInformation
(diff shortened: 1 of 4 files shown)


dennwc (approved):
Thank you for the fix! It definitely makes sense. I also made this function return an error directly, plus added a fallback to origin address (just in case).

PR #354, opened 2025-05-12
Stop RTP loop if the errors persist.

dennwc:


biglittlebigben (commented):
Does the RTP loop stopping trigger the call status to change to failed?

dennwc:
No, currently it only kills one incoming RTP stream (there could be multiple ones with different SSRC). Then if it turns out we don't have any RTP streams active, a regular media timeout will trigger.

PR #357, opened 2025-05-14
Log opus encoding/decoding errors.

dennwc:


pkg/media/opus/opus.go
```diff
@@ -92,9 +109,10 @@ func (d *decoder) WriteSample(in Sample) error {
 	if err != nil {
 		// Some workflows (concatenating opus files) can cause a suprious decoding error, so ignore small amount of corruption errors
 		if !errors.Is(err, opus.ErrInvalidPacket) || d.successiveErrorCount >= 5 {
+			d.log.Errorw("error decoding opus sample", err)
```
biglittlebigben: Do we want to trigger an alert on this? If now, we may want to use Warning or Info.

biglittlebigben (approved):

pkg/media/opus/opus.go
dennwc: You are right, let me switch decoding to warn. Encoding should be an alert still, I think.

PR #364, opened 2025-05-19
Update to support move participant. Switch to media SDK.

dennwc:


anunaym14 (approved):
The package name in media-sdk is still "media", so you can replace imports without changing anything else as well!

res/embed.go
anunaym14, line 1: Oh, `res` is also part of the media-sdk btw, we can get rid of it from SIP

super-ad:
@dennwc - since this merged (and assumed deployed) we are experiencing a new issue related to duplicate sessions being created for a single SIP caller where two agents are dispatched and then talk to each other on a single call line. It happens 1 out of every 6 calls on average.

it’s hard for me to grok what’s going on here without a PR description. Any chance this would affect dispatch to sessions created?

cc: @cassidypignatello @supervika

PR #367, opened 2025-05-20
Use signaling IP for Contact. Check if hostname resolves.

dennwc:


pkg/sip/protocol.go
```diff
@@ -151,8 +151,12 @@ func transportPort(c *config.Config, t Transport) int {
 }
 
 func getContactURI(c *config.Config, ip netip.Addr, t Transport) URI {
+	hostname := "" // use signaling IP by default, it's more robust
```
davidzhao, line 154: how does this get set as the host IP?

dennwc: It return the URI struct, which is then used only through `GetHost` and similar helpers. If hostname is empty, it [picks the IP](https://github.com/livekit/sip/blob/main/pkg/sip/types.go#L107). Below, you could see that the URI has IP address set as well.

davidzhao (approved):

PR #369, opened 2025-05-20
Update inbound.go

skotow:
I found issue when call won't hung-up if did not met Dispatch Rules . Caller will ring forever until decide to hung-up


skotow:
best one set to `c.cc.RespondAndDrop(sip.StatusBusyHere, "")`

dennwc:
This is intentional, so that we are not answering SIP spam. Automated scanners would (hopefully) think it's not a SIP server if we do not respond with a proper status code.

skotow:
I see . That makes sense.

Then maybe a good option would be enable response or disable via config
YAML  .
Anyway, thank you for your reply .
Best regards

dennwc:
YAML option makes sense. Actually we already have it: `HideInboundPort`. So let's add it to switch this behavior.

skotow:
Sounds good!

dennwc:
Closing this due to inactivity.

PR #372, opened 2025-05-22
fix(sip): send 487 Request Terminated after OK (CANCEL) for pending INVITE

### LINKED ISSUE #371: 487 not sent after CANCEL
I have a scenario where a SIP INVITE is cancelled before being answered.

I am bridging a Brekeke SIP server with LiveKit SIP.

Livekit SIP server respondes with 200 OK on Cancel but in the origin SIP server the call is still being shown as in calling state.

No 487 Request terminated is being sent by the LiveKit SIP. Is that the problem?

mubashiroliyantakath:
When a CANCEL is received for an INVITE that hasn't been answered, the SIP gateway now correctly sends a 487 response to the original INVITE. This ensures proper SIP transaction termination and avoids issues where upstream proxies (e.g., Brekeke) keep the call in an INVITE state due to the missing final response.

closes #371

### DIFF
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -1161,9 +1161,7 @@ func (c *sipInbound) StartRinging() {
 			case r := <-cancels:
 				close(c.cancelled)
 				_ = tx.Respond(sip.NewResponseFromRequest(r, sip.StatusOK, "OK", nil))
-				c.mu.Lock()
-				c.drop()
-				c.mu.Unlock()
+				c.RespondAndDrop(sip.StatusRequestTerminated, "Request Terminated")
 				return
 			case <-ticker.C:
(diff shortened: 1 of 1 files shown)


mubashiroliyantakath:
@dennwc thanks for the suggestion. Fixed it.

zusrut:
@mubashiroliyantakath   you also need to check if the INVITE already got final response and in this case send BYE to terminate the session instead of 487.

@dennwc livekit/sip does not support sip signaling and requires deep changes.

dennwc:
@zusrut although you are right about sipgo supporting dialogs, we do use an older version of it. We need to migrate to a newer version before we can rely on it.

dennwc (approved):

PR #373, opened 2025-05-24
Fix NPE when inbound close races with ack.

dennwc:


paulwe (approved):

super-ad:
@dennwc @paulwe @yepher  - is there anyway to get more transparency here?

This was a production facing bug that occurred from a recent deploy— causing a duplicate agent to be dispatched for our customers. 1 out of 6 customer calls was causing agents answering other agent speech when dispatched. Over the last 5 days we’ve scrambled to find a root cause in our source code even though it wasn’t linked to any of our recent deploys.

Livekit cloud using its deployment schedule that isn’t opt-in is very challenging for us as paying customers. 

PRs here generally do not have a description and they don’t always indicate a deploy from my understanding. How can we improve this process to have a little more transparency for upstream changes?

nishadmusthafa:
Hi @super-ad  this PR addressed a rare edge case and shouldn't have caused the issue you are talking about. Have you raised a support ticket on that issue ? If not, I'd recommend doing that so we can get more details and get to the bottom of it.

PR #385, opened 2025-06-07
Normalizing From and To Numbers for inbound calls

nishadmusthafa:
Certain carriers don't follow the E.164 format with numbers in the headers. This would lead to trunk matches failing and calls not working as expected. This takes care of that.


dennwc (changes requested):
This is a dangerous change. SIP username may contain arbitrary name, not just numbers. We cannot blindly normalise everything.

nishadmusthafa:
Moving this further inside for only cloud as it's failing only there.

PR #396, opened 2025-06-18
Adding a custom struct to use for call end hooks

nishadmusthafa:
The struct has project id, the scl call id generated by livekit/livekit and the protocol level sip call id


pkg/sip/outbound.go
```diff
@@ -302,7 +304,11 @@ func (c *outboundCall) close(err error, status CallStatus, description string, r
 
 		// Call the handler asynchronously to avoid blocking
 		if c.c.handler != nil {
-			go c.c.handler.OnCallEnd(context.Background(), c.state.callInfo, description)
+			go c.c.handler.OnSessionEnd(context.Background(), &CallIdentifier{
+				ProjectID: c.projectID,
+				CallID:    c.state.callInfo.CallId,
+				SipCallID: c.state.callInfo.ParticipantAttributes[AttrSIPCallIDFull],
```
dennwc: It should be available as a field in the call itself. If not, feel free to add it. Please avoid getting things from attribute map.

pkg/sip/inbound.go
```diff
@@ -827,7 +851,7 @@ func (c *inboundCall) close(error bool, status CallStatus, reason string) {
 
 	// Call the handler asynchronously to avoid blocking
 	if c.s.handler != nil {
-		go c.s.handler.OnCallEnd(context.Background(), c.state.callInfo, reason)
+		c.s.handleSessionEnd(context.Background(), c.call, c.state, c.state.callInfo.ParticipantAttributes[AttrSIPCallIDFull], reason, c.projectID)
```
dennwc: Same note for the call ID as below: it should be a field on a call, getting things from attributes should be avoided.

pkg/sip/inbound.go
```diff
@@ -247,11 +257,21 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 	case AuthDrop:
 		cmon.InviteErrorShort("flood")
 		log.Debugw("Dropping inbound flood")
+		sipCallID := ""
+		if h := req.CallID(); h != nil {
+			sipCallID = h.Value()
+		}
+		s.handleSessionEnd(ctx, callInfo, state, sipCallID, "flood", r.ProjectID)
```
dennwc: This might trigger very frequently. And we won't have a projectID here, so I don't think it's useful, since we can't surface it in the dashboard.

pkg/sip/inbound.go
```diff
@@ -128,6 +128,16 @@ func (s *Server) onInvite(log *slog.Logger, req *sip.Request, tx sip.ServerTrans
 	_ = s.processInvite(req, tx)
 }
 
+func (s *Server) handleSessionEnd(ctx context.Context, callInfo *rpc.SIPCall, state *CallState, sipCallID string, reason string, projectID string) {
+	if s.handler != nil {
+		go s.handler.OnSessionEnd(context.Background(), &CallIdentifier{
+			ProjectID: projectID,
+			CallID:    callInfo.LkCallId,
+			SipCallID: sipCallID,
+		}, state.callInfo, reason)
```
dennwc: This might race on `state.callInfo`, I guess.

pkg/service/service.go
```diff
@@ -254,6 +254,6 @@ func (s *Service) DeregisterTransferSIPParticipantTopic(sipCallId string) {
 	}
 }
 
-func (s *Service) OnCallEnd(ctx context.Context, callInfo *livekit.SIPCallInfo, reason string) {
-	s.log.Infow("SIP call ended", "callID", callInfo.CallId, "reason", reason)
+func (s *Service) OnSessionEnd(ctx context.Context, callIdentifier *sip.CallIdentifier, callInfo *livekit.SIPCallInfo, reason string) {
+	s.log.Infow("Post SIP call end processing", "callID", callInfo.CallId, "reason", reason)
```
dennwc: No actual processing here. So maybe just keep "call ended" log?

pkg/sip/inbound.go
nishadmusthafa: Only issue with not having this here is it would lead to a bunch of pcaps not getting cleaned out. There is logic to upload pcaps under a missingprojects folder if we can't identify the project. Afaict, it will not get triggered more than once per SIP session. Are you saying that it's much higher scale vs that of call connections ?

pkg/sip/inbound.go
nishadmusthafa: We pass that along in case it could be used for any post processing and expect it to be read only. So I think we are good here.

pkg/sip/inbound.go
nishadmusthafa: I'll remove this for now. But we might need to revisit in case those unconnected calls are too many and we have to deal with secondary memory leakage. We've got 10GB spare so I doubt it will be a problem.

PR #409, opened 2025-07-09
Sending Request terminated status when Call is cancelled

alastor09:
We observed that LiveKit SIP sometimes returned SIP status 486 Busy Here when a SIP caller canceled the call. This behavior was inconsistent with expected SIP cancellation handling.

Removed duplicate cancellation handling during ringing:
- Cancellation logic was previously executed in multiple places — both during ringing and WaitMedia.
- This caused incorrect status (like 486 Busy Here) to be returned when the call was simply canceled by the caller.

- Updated closeWithCancelled() to consistently mark the call as CallCancelled.
- Based on this status, the SIP response now correctly returns 487 Request Terminated, as per SIP spec for cancellations.

- Updated attributes to reflect that as well


alastor09:
@dennwc Fixes: #449

PR #412, opened 2025-07-14
Add an option to insert Record-Route header for inbound.

dennwc:


biglittlebigben (approved):
LGTM, but what is the context here?

dennwc:
This change is needed for correctly routing requests using our new SIP frontend.

PR #414, opened 2025-07-15
Adding additional logging for failed auths

nishadmusthafa:


pkg/sip/inbound.go
```diff
@@ -72,37 +83,93 @@ func (s *Server) getInvite(from string) *inProgressInvite {
 }
 
 func (s *Server) handleInviteAuth(log logger.Logger, req *sip.Request, tx sip.ServerTransaction, from, username, password string) (ok bool) {
+	log = log.WithValues(
+		"from", from,
```
dennwc: `from` should already be set on the logger (as `fromUser`)

pkg/sip/inbound.go
```diff
@@ -72,37 +83,93 @@ func (s *Server) getInvite(from string) *inProgressInvite {
...
+		"username", username,
+		"passwordHash", hashPassword(password),
+		"hasPassword", password != "",
```
dennwc: This one is covered by `hashPassword` returning `<empty>`.

pkg/sip/inbound.go
```diff
@@ -72,37 +83,93 @@ func (s *Server) getInvite(from string) *inProgressInvite {
...
+		"method", req.Method.String(),
+		"uri", req.Recipient.String(),
+	)
+
+	log.Infow("Starting SIP invite authentication")
+
 	if username == "" || password == "" {
+		log.Infow("Skipping authentication - no credentials provided")
```
dennwc: This should be a `Debug` probably, might be hit frequently.

pkg/sip/inbound.go
```diff
@@ -72,37 +83,93 @@ func (s *Server) getInvite(from string) *inProgressInvite {
...
+		log.Debugw("Created digest challenge",
+			"realm", inviteState.challenge.Realm,
+			"nonce", inviteState.challenge.Nonce,
+			"algorithm", inviteState.challenge.Algorithm,
+		)
+
 		res := sip.NewResponseFromRequest(req, 407, "Unauthorized", nil)
 		res.AppendHeader(sip.NewHeader("Proxy-Authenticate", inviteState.challenge.String()))
 		_ = tx.Respond(res)
+		log.Infow("Sent 407 Unauthorized response with Proxy-Authenticate header")
```
dennwc: We could collapse this with the line above ("Requesting inbound auth ..."), this way we will only emit one log line.

pkg/sip/inbound.go
```diff
@@ -111,15 +178,36 @@ func (s *Server) handleInviteAuth(log logger.Logger, req *sip.Request, tx sip.Se
 	})
 
 	if err != nil {
+		log.Errorw("Failed to compute digest response", err,
```
dennwc: We don't want it to page, so it should be `Warn`

pkg/sip/inbound.go
```diff
@@ -72,37 +83,93 @@ func (s *Server) getInvite(from string) *inProgressInvite {
...
+		log.Errorw("Failed to parse Proxy-Authorization credentials", err,
+			"headerValue", h.Value(),
+		)
+		_ = tx.Respond(sip.NewResponseFromRequest(req, 401, "Bad credentials", nil))
+		return false
+	}
+
+	log.Debugw("Parsed credentials successfully",
+		"credUsername", cred.Username,
+		"credURI", cred.URI,
```
dennwc: I see this one is set later as just `uri`, so maybe we should just set it once on the logger?

pkg/sip/inbound.go
```diff
@@ -72,37 +83,93 @@ func (s *Server) getInvite(from string) *inProgressInvite {
...
+		"credRealm", cred.Realm,
```
dennwc: You could also just dump the whole struct as `cred`. But I haven't looked if there are sensitive fields there or not.

dennwc (approved):

PR #418, opened 2025-07-18
Using sipCallID to map the inviteState and the challenge data between…

nishadmusthafa:
… sip auth invites

We were using the from number and that ran into issues when there were simultaneous calls from the same from number

### DIFF
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -1162,14 +1162,17 @@ type sipInbound struct {
 }
 
 func (c *sipInbound) ValidateInvite() error {
+	if c.callID == "" {
+		return errors.New("no Call-ID header in INVITE")
+	}
 	if c.from == nil {
-		return errors.New("no From header")
+		return errors.New("no From header in INVITE")
 	}
 	if c.to == nil {
-		return errors.New("no To header")
+		return errors.New("no To header in INVITE")
 	}
 	if c.tag == "" {
-		return errors.New("no tag in From")
+		return errors.New("no tag in From in INVITE")
 	}
 	return nil
 }


pkg/sip/inbound.go
```diff
@@ -103,8 +103,13 @@ func (s *Server) handleInviteAuth(log logger.Logger, req *sip.Request, tx sip.Se
 		_ = tx.Respond(sip.NewResponseFromRequest(req, 100, "Processing", nil))
 	}
 
-	inviteState := s.getInvite(from)
-	log = log.WithValues("inviteStateFrom", from)
+	// Extract SIP Call ID for tracking in-progress invites
+	sipCallID := ""
+	if h := req.CallID(); h != nil {
```
dennwc, line 108: We should require CallID parameter the same way as we do with From header.

pkg/sip/inbound.go
```diff
@@ -266,12 +271,19 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 		cc.Processing()
 	}
 
+	// Extract SIP Call ID directly from the request
+	sipCallID := ""
+	if h := req.CallID(); h != nil {
```
dennwc, line 276: Same here, we should require CallID to be set.

dennwc (approved):

pkg/sip/inbound.go
nishadmusthafa: I actually thought about that. But I was wondering if we should do that at the sipgo level(if it isn't already being done) when parsing the invite. Invite without sipcallID won't be valid at all right ?

dennwc: It may be done on the dialog level of sipgo, but for us it looks like it can still be nil. I was thinking that we could add it to the call structure or just get the header earlier and pass it along. We do a similar check for [other headers](https://github.com/livekit/sip/blob/main/pkg/sip/inbound.go#L1152).

PR #427, opened 2025-07-31
fix(outbound) read error message from twilio auth header

jtgi:
- Fixes twilio typo to get proper error message during auth loop


nishadmusthafa (approved):
Thanks for submitting this @jtgi !

PR #431, opened 2025-08-12
Do not force 200 OK response destination.

dennwc:


pkg/sip/inbound.go
```diff
@@ -1378,13 +1361,18 @@ func (c *sipInbound) AcceptBye(req *sip.Request, tx sip.ServerTransaction) {
...
+		dest = ConvertURI(&contact.Address).GetDest()
 	} else {
 		req.Recipient = c.from.Address
 	}
+	if route := c.invite.RecordRoute(); route != nil {
+		dest = ConvertURI(&route.Address).GetDest()
+	}
 	req.SetSource(c.inviteOk.Source())
-	req.SetDestination(c.inviteOk.Destination())
+	req.SetDestination(dest)
```
nishadmusthafa, line 1375: Could you share more on why we are using this order(checking contact first and then record route) when determiningh what the destination needs to be. Also is this handling the case of REFER using swapSrcDst ?

nishadmusthafa (approved):
I guess there may be some tests to fix or augment for these use cases

PR #443, opened 2025-08-27
Initial implementation of Hold/Resume call

Kaian:
We've been exploring the idea of enabling real attended call transfers from LiveKit rooms, with the first step being support for call holding. When connecting LiveKit Trunks from a PBX, this allows music-on-hold configured in the PBX to be played to the participant.

This PR introduces the initial implementation of call hold and unhold functionality in the LiveKit SIP component. It allows the LiveKit Server to put SIP calls on hold and resume them (unhold), by adding the necessary handling of re-INVITE messages for hold/unhold.

Further PRs will be required in other LiveKit projects to fully implement this functionality across the platform (including livekit, livekit-sdks, livekit-protocol and so on).

We’d love to hear if the team is interested in this functionality and any feedback on the approach to summit additional Pull Requests.

### DIFF
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/frostbyte73/core"
 	"github.com/icholy/digest"
+	psdp "github.com/pion/sdp/v3"
 	"github.com/pkg/errors"
 
 	msdk "github.com/livekit/media-sdk"
@@ -1003,6 +1004,7 @@ func (c *inboundCall) runMediaConn(tid traceid.ID, offerData []byte, enc livekit
 	if err != nil {
 		return nil, err
 	}
+	c.cc.nextSDPVersion = answer.SDP.Origin.SessionVersion + 1
 	c.mon.SDPSize(len(answerData), false)
 	c.log().Debugw("SDP answer", "sdp", string(answerData))
 
@@ -1421,6 +1423,48 @@ func (c *inboundCall) transferCall(ctx context.Context, transferTo string, heade
 
 }
 
+func (c *inboundCall) holdCall(ctx context.Context) error {
+	c.log().Infow("holding inbound call")
+
+	// Disable media timeout during hold to prevent call termination
+	if c.media != nil {
+		c.media.EnableTimeout(false)
+		c.log().Infow("media timeout disabled for hold")
+	}
+
+	err := c.cc.holdCall(ctx)
+	if err != nil {
+		c.log().Infow("inbound call failed to hold", "error", err)
+		// Re-enable timeout if hold failed
+		if c.media != nil {
+			c.media.EnableTimeout(true)
+		}
+		return err
+	}
+
+	c.log().Infow("inbound call held")
+	return nil
+}
+
+func (c *inboundCall) unholdCall(ctx context.Context) error {
+	c.log().Infow("unholding inbound call")
+
+	err := c.cc.unholdCall(ctx)
+	if err != nil {
+		c.log().Infow("inbound call failed to unhold", "error", err)
+		return err
+	}
+
+	// Re-enable media timeout after unhold
+	if c.media != nil {
+		c.media.EnableTimeout(true)
+		c.log().Infow("media timeout re-enabled after unhold")
+	}
+
+	c.log().Infow("inbound call unheld")
+	return nil
+}
+
 func (s *Server) newInbound(log logger.Logger, contact URI, invite *sip.Request, inviteTx sip.ServerTransaction, getHeaders setHeadersFunc) *sipInbound {
 	c := &sipInbound{
 		log:      log,
@@ -1467,6 +1511,7 @@ type sipInbound struct {
 	referDone  chan error
 
 	mu              sync.RWMutex
+	sdpMu           sync.RWMutex
 	lastSDP         []byte
 	inviteOk        *sip.Response
 	nextRequestCSeq uint32
@@ -1474,6 +1519,7 @@ type sipInbound struct {
 	ringing         chan struct{}
 	acked           core.Fuse
 	setHeaders      setHeadersFunc
+	nextSDPVersion  uint64
 }
 
 func (c *sipInbound) ValidateInvite() error {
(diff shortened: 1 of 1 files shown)


dennwc (commented):
Thank you a lot for contributing the hold/resume functionality! The idea and implementation makes perfect sense. New internal API is the most straightforward way to achieve this.

What I'm wondering about is, would it make sense to hold/unhold automatically when SIP participant looses all audio track subscriptions? In this case, if a person/agent that is in the room suddenly drops, we could indicate a hold for the SIP participant instead of just relaying silence.

This way, there will be no extra API required - existing participant admin API can be used to unsubscribe SIP from existing tracks, move it to another empty room or disconnecting the other peer from the room.

What do you think about this approach?

Kaian:
Hi!

Thanks for the feedback!

We're currently testing this feature in the following scenario: An incoming call is dispatched to a room where an agent can be requested to perform a transfer, first checking whether the destination is available. The agent places the first participant on hold and creates a second participant in the same room by making an outbound call. When this second participant answers, their availability is verified, and based on the response, we remove the second participant from the room and perform a cold transfer from the first participant (or not). This approach has some advantages, such as providing a full transcription or recording of the entire process while only using a single room.

I have little to no experience with LiveKit, so I cannot say what the best approach would be. However, I’m concerned that some providers may not accept in-dialog re-INVITEs the way PBXs do, so making this behavior implicit could cause future issues in other scenarios.

Of course, your opinion is what matters the most, and if you decide to implement it differently, we’ll be more than happy to adapt how our agent works!

Best regards,

Kaian:
Hi!

Sorry for the long silence, I've been quite busy.

I've been thinking about your proposal, and I understand the point of not changing the API to avoid impacting other modules (SDK, protocol, etc.).

I think at least a Trunk flag will be necessary to mark which providers support receiving an in-dialog re-INVITE, something like `holdSupport`, disabled by default for inbound/outbound trunks. Media direction would only change for those capable of playing music on hold.

I've rebased and updated the patch by removing some unnecessary code, fixing a problem with `SDPVersion` not being updated in new messages, and removing all internal API-related code. As a result, the `hold`/`unhold` functions are not actually used in this current proposed code. I'm not sure where track subscriptions should be handled to trigger the hold/unhold process (maybe in the LiveKit core project?) so I’ll need some guidance.

Also, regarding the failing tests, I don’t know what the root cause of the problem is and I’m not sure how our changes might have affected them.

Feel free to request any additional information; we’re entirely at your disposal.

Thanks for your time!

pkg/sip/outbound.go
```diff
@@ -530,6 +531,7 @@ func (c *outboundCall) sipSignal(ctx context.Context) error {
 	}
 	c.mon.SDPSize(len(sdpOfferData), true)
 	c.log.Debugw("SDP offer", "sdp", string(sdpOfferData))
+	c.cc.nextSDPVersion = sdpOffer.SDP.Origin.SessionVersion + 1
```
dennwc: Maybe set it in the `cc.Invite`?

pkg/sip/inbound.go
```diff
@@ -1830,3 +1875,169 @@ func (c *sipInbound) CloseWithStatus(code sip.StatusCode, status string) {
...
+	for _, mediaDesc := range desc.SDP.MediaDescriptions {
+		if mediaDesc == nil {
+			continue
+		}
+
+		// Find and remove existing direction attributes
+		var newAttributes []psdp.Attribute
+		for _, attr := range mediaDesc.Attributes {
+			// Keep all attributes except direction-related ones
+			if attr.Key != "sendrecv" && attr.Key != "sendonly" &&
```
dennwc: Maybe use `switch` so that it's easier to read?
```go
switch attr.Key {
default:
    newAttributes = append(newAttributes, attr)
case "sendrecv","sendonly","recvonly","inactive":
   // skip
}
```
Or better yet, use `slices.DeleteFunc`.

pkg/sip/inbound.go
```diff
@@ -1830,3 +1875,169 @@ func (c *sipInbound) CloseWithStatus(code sip.StatusCode, status string) {
...
+	req.AppendHeader(c.contact)
+	req.AppendHeader(sip.NewHeader("Content-Type", "application/sdp"))
+	req.AppendHeader(sip.NewHeader("Allow", "INVITE, ACK, CANCEL, BYE, NOTIFY, REFER, MESSAGE, OPTIONS, INFO, SUBSCRIBE"))
+
+	// Modify SDP to set direction to sendonly (hold)
+	sdpOffer := c.inviteOk.Body()
+	if len(sdpOffer) > 0 {
+		modifiedSDP, err := c.setMediaDirection(sdpOffer, "sendonly")
+		if err != nil {
+			return err
```
dennwc: Missing `c.mu.Unlock()` here. Maybe add a separate function that creates a request, so that it can defer the unlock?

pkg/sip/inbound.go
```diff
@@ -1830,3 +1875,169 @@ func (c *sipInbound) CloseWithStatus(code sip.StatusCode, status string) {
...
+	modifiedSDP, err := desc.SDP.Marshal()
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal modified SDP: %w", err)
+	}
+
+	return modifiedSDP, nil
+}
+
+func (c *sipInbound) holdCall(ctx context.Context) error {
+	c.mu.Lock()
```
dennwc, line 2052: We might want a second mutex here, like `holdMu` or `sdpMu` that will prevent multiple calls to hold/unhold.

pkg/sip/inbound.go
```diff
@@ -1830,3 +1875,169 @@ func (c *sipInbound) CloseWithStatus(code sip.StatusCode, status string) {
...
+	}
+	defer tx.Terminate()
+
+	resp, err := sipResponse(ctx, tx, c.s.closing.Watch(), nil)
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode != sip.StatusOK {
+		return &livekit.SIPStatus{Code: livekit.SIPStatusCode(resp.StatusCode)}
```
dennwc: We now also extract the reason for these status codes. Maybe worth adding a helper for it at this point.

pkg/sip/inbound.go
```diff
@@ -1830,3 +1875,169 @@ func (c *sipInbound) CloseWithStatus(code sip.StatusCode, status string) {
...
+	req.AppendHeader(c.contact)
+	req.AppendHeader(sip.NewHeader("Content-Type", "application/sdp"))
+	req.AppendHeader(sip.NewHeader("Allow", "INVITE, ACK, CANCEL, BYE, NOTIFY, REFER, MESSAGE, OPTIONS, INFO, SUBSCRIBE"))
+
+	// Modify SDP to set direction to sendrecv (unhold)
+	sdpOffer := c.inviteOk.Body()
+	if len(sdpOffer) > 0 {
+		modifiedSDP, err := c.setMediaDirection(sdpOffer, "sendrecv")
+		if err != nil {
+			return err
```
dennwc: Same - missing unlock.

pkg/sip/outbound.go
```diff
@@ -1046,3 +1092,177 @@ func (c *sipOutbound) Close() {
 		c.drop()
 	}
 }
+
+func (c *sipOutbound) setMediaDirection(sdpData []byte, direction string) ([]byte, error) {
```
dennwc: Looks like it doesn't access anything on `c` and the code is basically the same as for inbound. Maybe it should be on a `MediaPort` instead? It would also be easier to test if it works with a media timeout, etc.

dennwc (commented):
Thank you for you patience and for working on this change! We finally have a window to help move it over the finish line.

Kaian:
First of all, thank you for taking the time to review this.

As a reminder, these changes are just a code snippet. By themselves, they don't add any functionality since they aren't called from anywhere yet. The version we have been testing in our fork was creating two API endpoints, but that require updating more livekit projects.

You should know that my Go skills are quite limited, but I have some knowledge of the SIP protocol. Please feel free to reimplement anything as you see fit, I truly value your time and wouldn't want you spending it extensively reviewing a novice's PRs.

I've rewritten part of the code to separate the functionality and have only focused on the inbound dialog (I think this feature will be mostly for holding incoming calls). Based on the resulting code, it should be easier to extrapolate for outgoing dialogs.

Best Regards,

pkg/sip/inbound.go
Kaian: I've done a combination of both

```
		newAttributes := slices.DeleteFunc(mediaDesc.Attributes, func(attr psdp.Attribute) bool {
			switch attr.Key {
			case "sendrecv", "sendonly", "recvonly", "inactive":
				return true
			default:
				return false
			}
		})
```

pkg/sip/inbound.go
Kaian: I have moved this to a function that creates the reInvite request

pkg/sip/outbound.go
Kaian: Maybe the proper place for this would be media-sdk/sdp, that seems to be the place where all SDP changes are made, but that's another project

pkg/sip/outbound.go
Kaian: This is a dialog-scope value that must be incremented with every SDP update. I saw the `c.cc.nextRequestCSeq` flag and thought it had a similar purpose.

dennwc: Sure, I'm just saying maybe it's a good idea to set it in one of the methods of the object itself as opposed to setting it externally.

PR #452, opened 2025-09-11
Distinguish between inbound call and leg transports.

dennwc:


pkg/sip/protocol.go
```diff
@@ -151,6 +153,26 @@ func transportFromReq(req *sip.Request) Transport {
 	return ""
 }
 
+// legTransportFromReq returns the SIP transport used between SIP server and LK SIP edge.
+// For the transport used between LK SIP and the provider, see callTransportFromReq.
+func legTransportFromReq(req *sip.Request) Transport {
+	if via := req.Via(); via != nil {
+		return Transport(strings.ToLower(via.Transport))
```
nishadmusthafa, line 160: via.Transport is set using what exactly in the packet ?

pkg/sip/inbound.go
```diff
@@ -235,9 +235,9 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 	)
 
 	var call *inboundCall
-
-	tr := transportFromReq(req)
-	cc := s.newInbound(LocalTag(callID), s.ContactURI(tr), req, tx, func(headers map[string]string) map[string]string {
+	tr := callTransportFromReq(req)
+	legTr := legTransportFromReq(req)
+	cc := s.newInbound(LocalTag(callID), s.ContactURI(legTr), req, tx, func(headers map[string]string) map[string]string {
```
nishadmusthafa, line 240: I guess this was affecting all FE traffic. So we were using the wrong transport when sending BYE, 200OK and REFERs ?

pkg/sip/protocol.go
dennwc: `Via` is the first hop that the response packet must take. So it is the transport of the first leg/hop.

pkg/sip/inbound.go
dennwc: Roughly, yes. It was forcing TLS transport on the `Contact` header only. This is incorrect, since that leg must use TCP to/from FE.

PR #455, opened 2025-09-17
Properly end the call if accept fails.

dennwc:


pkg/sip/inbound.go
```diff
@@ -1608,6 +1620,7 @@ func (c *sipInbound) CloseWithStatus(code sip.StatusCode, status string) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	if c.inviteOk != nil {
+		// TODO: add cause for a failure, if any
```
nishadmusthafa, line 1623: So this was meant to be this generic logic for call end where if we sent a 200OK for the invite, then we need to send a BYE  ?

dennwc: Exactly. And in this case we were not setting the `c.inviteOk` when no ACK was received, and thus no BYE was sent.

PR #456, opened 2025-09-17
chore(mixer): move to media-sdk

anunaym14:


pkg/sip/room.go
```diff
@@ -96,7 +96,7 @@ func NewRoom(log logger.Logger, st *RoomStats) *Room {
 	}
 	r := &Room{log: log, stats: st, out: msdk.NewSwitchWriter(RoomSampleRate)}
 	out := newMediaWriterCount(r.out, &st.OutputFrames, &st.OutputSamples)
-	r.mix = mixer.NewMixer(out, rtp.DefFrameDur, &st.Mixer)
+	r.mix, _ = mixer.NewMixer(out, rtp.DefFrameDur, &st.Mixer, 1)
```
dennwc: We should at least panic here in case of an error. Debugging `r.mix == nil` deeper in the code is never fun 😅

test/lktest/livekit.go
```diff
@@ -179,7 +179,7 @@ func (lk *LiveKit) ConnectParticipant(t TB, room, identity string, cb *RoomParti
 		pr.Close()
 	})
 	p.AudioIn = pr
-	p.mixIn = mixer.NewMixer(pw, rtp.DefFrameDur, nil)
+	p.mixIn, _ = mixer.NewMixer(pw, rtp.DefFrameDur, nil, 1)
```
dennwc:
```suggestion
	p.mixIn, err = mixer.NewMixer(pw, rtp.DefFrameDur, nil, 1)
	if err != nil {
		t.Fatal(err)
	}
```

pkg/sip/room.go
anunaym14: yep, my bad.

PR #457, opened 2025-09-19
Add an option to proceed with the call if ACK is missing.

dennwc:


pkg/sip/inbound.go
```diff
@@ -617,14 +617,19 @@ func (c *inboundCall) handleInvite(ctx context.Context, req *sip.Request, trunkI
...
+				return false, err
 			}
+			c.log.Warnw("Call accepted, but no ACK received", err)
+			err = nil // ignore
+		}
+		if err != nil {
+			c.log.Errorw("Cannot accept the call", err)
+			c.close(true, callAcceptFailed, "accept-failed")
 			return false, err
 		}
```
alexlivekit, line 634: Minor style question - Do we prefer less blocky and thus get to `if err == something_specific { ... ;  err = nil; }` followed by `if err != nil { ... }`? I'm comparing this to something like the following which has more indentation.
```
if err != nil {
	if errors.Is(err, errNoACK) {
		if !c.s.conf.Experimental.IgnoreMissingACK {
			c.log.Errorw("Call accepted, but no ACK received", err)
			c.close(true, callNoACK, "no-ack")
			return false, err
		} 
		c.log.Warnw("Call accepted, but no ACK received", err)
	} else {
		c.log.Errorw("Cannot accept the call", err)
		c.close(true, callAcceptFailed, "accept-failed")
		return false, err
	}
}
```

dennwc: Yes, it' the "return early" style, where you keep the longest/normal code flow unnested. If you can write a short return condition and avoid block nesting - that's preferable. [Here](https://go.dev/wiki/CodeReviewComments#indent-error-flow) is the link to the Go style guide for it.

PR #459, opened 2025-09-19
Adding configurable display name

alexlivekit:
Now, I have not added any tests for this, but it looks like we don't have any tests for the sip package. Do we not want to do unit tests, and do more high-level ones instead?


alexlivekit:
Umm, ping?

pkg/sip/outbound.go
```diff
@@ -123,7 +124,7 @@ func (c *Client) newCall(ctx context.Context, conf *config.Config, log logger.Lo
 			return headers
 		}
 		return AttrsToHeaders(r.LocalParticipant.Attributes(), c.sipConf.attrsToHeaders, headers)
-	})
+	}, sipConf.displayName)
```
nishadmusthafa: NIT: maybe just add it before the setHeadersFunc argument for readability

nishadmusthafa (approved):

PR #461, opened 2025-09-23
Do not block when ACK is missing.

dennwc:


pkg/sip/inbound.go
```diff
@@ -52,11 +52,13 @@ import (
...
 
 const (
+	stateUpdateTick = 10 * time.Minute
+
 	// audioBridgeMaxDelay delays sending audio for certain time, unless RTP packet is received.
 	// This is done because of audio cutoff at the beginning of calls observed in the wild.
 	audioBridgeMaxDelay = 1 * time.Second
 
-	inviteOkAckTimeout = 5 * time.Second
+	inviteOkAckTimeout = 3 * time.Second
```
alexlivekit: Not sure what we want to do with this, but the general SIP spec has exponential backoff for resending invites or 200 responses. T1 is by default 500ms (Some companies shorten it to allow for quicker recovery) and it doubles every time one re-sends a request or INVITE-200.
The transaction timeout if no ACK arrives is 64*T1 (= 32 seconds by default, some companies limited that to ~12 seconds).

So the idea here is that we're allowing 3x transmissions 0ms,500ms, 15000ms) and then some time for the ACK to get back (remaining 1500ms). Do I understand your drift correctly?

pkg/sip/media_port.go
```diff
@@ -254,25 +282,40 @@ func (p *MediaPort) timeoutLoop(timeoutCallback func()) {
...
+				timeout time.Duration
+			)
+			// First timeout could be different. Usually it's longer to allow for a call setup.
+			// In some cases it could be shorter (e.g. when we notice an issue with signaling and suspect media will fail).
+			if isInitial {
+				since = sinceStart
+				timeout = p.opts.MediaTimeoutInitial
+				if ptr := p.timeoutInitial.Load(); ptr != nil {
+					timeout = *ptr
+				}
```
alexlivekit, line 299: QQ: Will there ever be a situation where these are empty and the timer is enabled? IIUC, we always set both `MediaTimeoutInitial` and `MediaTimeout` when enabling a timer, no?

pkg/sip/inbound.go
```diff
@@ -734,8 +742,21 @@ func (c *inboundCall) handleInvite(ctx context.Context, req *sip.Request, trunkI
...
+			}
 			c.closeWithTimeout()
 			return psrpc.NewErrorf(psrpc.DeadlineExceeded, "media timeout")
+		case <-ackReceived:
+			ackTimeout = nil // all good, disable timeout
+		case <-ackTimeout:
+			// Only warn, the other side still thinks the call is active, media may be flowing.
+			c.log.Warnw("Call accepted, but no ACK received", errNoACK)
+			// We don't need to wait for a full media timeout initially, we already know something is not quite right.
+			c.media.SetTimeout(min(inviteOkAckTimeout, c.s.conf.MediaTimeoutInitial), c.s.conf.MediaTimeout)
```
alexlivekit: Have we not already waited `inviteOkAckTimeout` in this case (we did not wait for ack, no ack was processed, `ackTimeout` happened)?
So we want to wait until the full `ack_to` + `initial_media_to`, or 2x`ack_to`?

pkg/sip/media_port.go
dennwc: Yes, that's just defensive coding. In case someone removes the initial setting for these atomics later, it won't panic here. Currently it doesn't make sense, as you noticed.

pkg/sip/inbound.go
dennwc: Right, we just need to set something shorter than the default media timeout, which is currently ~30 sec. Actually this waits for ACK timeout on top of anything that happens during our media init (e.g. waiting for an agent to join). I guess I should move the timeout earlier.

pkg/sip/inbound.go
dennwc: The assumption was that the underlying transport in the library will take care of it, so the timeout was only meant for detecting routing issues where we just cannot get their response back. Now I added extra code to force retries in our app layer. Please take a look if it makes sense now.

pkg/sip/inbound.go
```diff
@@ -734,8 +748,21 @@ func (c *inboundCall) handleInvite(ctx context.Context, req *sip.Request, trunkI
...
 			return psrpc.NewErrorf(psrpc.DeadlineExceeded, "media timeout")
+		case <-ackReceived:
+			ackTimeout = nil // all good, disable timeout
+		case <-ackTimeout:
+			// Only warn, the other side still thinks the call is active, media may be flowing.
+			c.log.Warnw("Call accepted, but no ACK received", errNoACK)
+			// We don't need to wait for a full media timeout initially, we already know something is not quite right.
+			c.media.SetTimeout(min(inviteOkAckLateTimeout, c.s.conf.MediaTimeoutInitial), c.s.conf.MediaTimeout)
+			noAck = true
 		}
```
alexlivekit, line 766: I might be reading it wrong, but would the following not be simpler? Asking a different way, what is the purpose for the extra timeout wait? I we want to wait longer, set `inviteOkAckLateTimeout` to a bigger number, no?

```suggestion
		case <-ackReceived:
			ackTimeout = nil // all good, disable timeout
		case <-ackTimeout:
			c.log.Errorw("Call accepted, but no ACK received", errNoACK)
			c.closeWithNoACK()
			return psrpc.NewError(psrpc.DeadlineExceeded, errNoACK)
		}
```

pkg/sip/inbound.go
```diff
@@ -1390,23 +1432,51 @@ func (c *sipInbound) Accept(ctx context.Context, sdpData []byte, headers map[str
...
+		// Still retry, but limit it to ~750ms.
+		maxRetries = inviteOKRetryAttemptsNoACK
+	}
+	if c.legTr != TransportUDP {
+		maxRetries = 1
+		// That actually becomes an ACK timeout here.
+		retryAfter = inviteOkRetryIntervalMax
+	}
+	var acceptErr error
+retries:
```
alexlivekit, line 1447: Do we always specify loop labels like this, even when there are no additional nested break-continue blocks?

pkg/sip/inbound.go
```diff
@@ -1390,23 +1432,51 @@ func (c *sipInbound) Accept(ctx context.Context, sdpData []byte, headers map[str
...
+		if c.legTr != TransportUDP && !c.s.conf.Experimental.InboundWaitACK {
+			// Reliable transport and we are not waiting for ACK - return immediately.
+			break retries
+		}
+		t := time.NewTimer(retryAfter)
+		select {
+		case <-c.inviteTx.Acks():
+			t.Stop()
+			break retries
+		case <-c.acked.Watch():
```
alexlivekit, line 1461: I see this already exists, but what's the difference between these? If something else is already waiting for an ACK, and then setting `c.acked`, or if that's this piece of code, don't we want to set the ACK here?

I see we're configuring `s.sipSrv.OnAck(s.onAck)`, so do we need to listed to both here?

alexlivekit (approved):
Still lgtm overall, and still more questions :)

pkg/sip/inbound.go
dennwc: This branch is hit only if config says we should ignore missing ACK. Since we sent 200 OK, other side might think the call is established and send media. This is why here we do not exit immediately and instead just lower the media timeout. If media starts flowing, we will proceed with the call. If not, we will soon fail with a media timeout. That was the thinking process here.

pkg/sip/inbound.go
dennwc: Right, it's only needed in case we want to `break`, but that command affects something else instead of a loop (`select` in this case). So the label allows us to exit loop properly.

pkg/sip/inbound.go
dennwc: I assume you mean two ack signals? In theory, the ack channel on a transaction should trigger. But it doesn't. So I added a manual check and another channel. If we fix the underlying library, we could remove the second channel as well.

pkg/sip/inbound.go
alexlivekit: Ah, thanks for reminding me of this. That makes sense.

I think a comment here may be useful, something like `// Fail only if there's no media, that serves as an implicit ACK`

pkg/sip/inbound.go
alexlivekit: Yes, my question was about having both of `c.inviteTx.Acks()` and `c.acked.Watch()` (and even `InviteACK()`, which us just the latter wrapped in a function). In other places in the code we're only using `c.acked.Watch()`.

 `c.acked.Watch()` feels more useful - You never need to ACK a transaction twice, and being able to always immediately trigger the channel to fire already comes in handy.

So in theory we should be able to drop the use of `c.inviteTx.Acks()` here, correct?

alexlivekit: Come to think about it, here's an important Q AFAICT:
If we listen to `c.inviteTx.Acks()` here, and it consumes the message queued to the channel, would that not prevent `s..onAck()` from firing and properly executing `c.acked.break()`?

dennwc: That should be a different code path, I believe (worth double checking). If config says we should properly wait for ACK, it will be handled here. If not - it will be handled in a top level select that we discussed (the one with media timeout).

But also, closed channels fire indefinitely, which is the case for `acked`. So it cannot be consumed completely.

alexlivekit: Makes sense. So the transaction closes the back channel once one is received?

Then should we be able to use that channel directly instead of acked.watch() in a potential future change?

alexlivekit: Interesting. I'm looking at [sipgo//transaction/server_tx.go](https://github.com/livekit/sipgo/blob/ca5b8ca7b53daab19eaf693ffcff36b4081e6d2c/transaction/server_tx.go#L19) and the channel backing Acks() is not explicitly closed at any point. So the `c.inviteTx.Acks()` channel is not a good fit.

However, the handlers `s.onAck()` are later called directly, independently of any channel.

Basically:
```
Layer.handleRequest()
  ServerTx.Receive(req)
    ServerTx.receiveRequest(req)
    ServerTx.spinFsm(server_input_ack)
      ServerTx.inviteStateAccepted()
        ServerTx.actPassupAck() <<< Results in c.inviteTx.Acks() channel push, once on every ack received.
  Layer.reqHandler(req, tx)
    ServerDialog.onRequestDialog()
      ServerDialog.publish() <<< We don't use these listeners in the ACK handling at the very least
      Server.handleRequest()
        handler := srv.getHandler(req.Method)
        handler(req, tx) <<< Results in `s.onAck()` being called.
```

dennwc: Yeah, we need to fix the library to actually send/close on that channel. Currently it doesn't and we cannot close their channel because it's receive only (`<-chan something`) and closing an already closed channel will panic.

In the meantime we have our own channel as a workaround. Later we could remove it and rely on the transaction channel.

PR #462, opened 2025-09-25
Validate only TLS cert validity, not the hostname.

dennwc:
In SIP, each call leg may have a different transport, including TLS. And the leg of the call is usually identified with an IP instead of the hostname, which breaks default TLS cert verification. This change removes the check for the hostname, while still verifying the cert chain against root CAs.


alexlivekit (approved):
We're not setting `CurrentTime:   c.config.time(),`, I'm assuming we made sure `Verify()` still checks for expiry time.

otherwise lgtm

PR #464, opened 2025-09-25
Log invite retry attempts.

dennwc:
This will help us detect issues with the singling earlier.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -33,6 +33,7 @@ require (
 )
 
 require (
+	github.com/hashicorp/golang-lru/v2 v2.0.7 // indirect
 	github.com/nyaruka/phonenumbers v1.6.5 // indirect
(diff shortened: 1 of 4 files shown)


dennwc:
@alexlivekit good point! This is a copy from our FE code where we have a way to distinguish invites vs reinvites, but not here. I'll see if I can add a check that makes sense. Do you know a good way to distinguish them? In the worst case, I could emit info logs instead of warnings and we can use it as a statistic during deployments.

alexlivekit:
A re-invite will have a different sequence number compared to the original invite, so there's that.

dennwc:
@alexlivekit done, its now tracking cseq for the initial invite (either with auth or without it).

alexlivekit (approved):
lgtm

PR #471, opened 2025-10-03
Parse the BYE reason.

dennwc:


pkg/sip/inbound.go
```diff
@@ -1062,17 +1080,49 @@ func (c *inboundCall) closeWithNoACK() {
...
+	}
+	c.closeWithReason(CallHangup, "hangup", reason)
+}
+
+func (c *inboundCall) closeWithReason(status CallStatus, reasonName string, reason ReasonHeader) {
 	c.state.DeferUpdate(func(info *livekit.SIPCallInfo) {
 		info.DisconnectReason = livekit.DisconnectReason_CLIENT_INITIATED
+		if info.Error == "" {
+			if !reason.IsZero() {
+				info.Error = reason.String()
```
nishadmusthafa, line 1103: Do we want to populate error for all reasons ?

pkg/sip/inbound.go
```diff
@@ -1062,17 +1080,49 @@ func (c *inboundCall) closeWithNoACK() {
...
+			switch reason.Cause {
+			case 16: // Normal call clearing
+				change = false
+			}
+		}
+		if change {
+			reasonName = fmt.Sprintf("bye-%s-%d", strings.ToLower(reason.Type), reason.Cause)
+		}
+	}
+	c.close(false, status, reasonName)
```
nishadmusthafa, line 1112: should the first argument be false in all cases ?

nishadmusthafa (approved):

pkg/sip/inbound.go
dennwc: Yes, we got a response from the remote, so it's not a critical error to log.

pkg/sip/inbound.go
dennwc: Good point. I'll check for a normal clearing here.

nishadmusthafa (approved):

PR #472, opened 2025-10-05
Generating and logging the sipCallID earlier

nishadmusthafa:


pkg/sip/outbound.go
```diff
@@ -126,6 +131,9 @@ func (c *Client) newCall(ctx context.Context, conf *config.Config, log logger.Lo
 		return AttrsToHeaders(r.LocalParticipant.Attributes(), c.sipConf.attrsToHeaders, headers)
 	})
 
+	// Set the SIP call ID in the sipOutbound struct
+	call.cc.callID = sipCallID
```
dennwc, line 135: We should pass it to `newOutbound` then.

pkg/sip/outbound.go
```diff
@@ -108,7 +108,12 @@ func (c *Client) newCall(ctx context.Context, conf *config.Config, log logger.Lo
 		jitterBuf: jitterBuf,
 		projectID: projectID,
 	}
-	call.log = call.log.WithValues("jitterBuf", call.jitterBuf)
+	// Generate SIP call ID early so it's available for all logging
+	toUri := CreateURIFromUserAndAddress(sipConf.to, sipConf.address, tr)
+	toHeader := &sip.ToHeader{Address: *toUri.GetURI()}
```
dennwc, line 113: Not against it, but it's a bit unfortunate that it now requires building these two objects outside of the constructor function.

nishadmusthafa:
Closing as Alex is doing the same with https://github.com/livekit/sip/pull/488/files

PR #475, opened 2025-10-10
Improve logging for media

dennwc:
- Log media timeout resets.
- Add media timeout tests.
- Print call stats (including media) regularly.
- Track closed states for the call components.

### DIFF
--- a/pkg/sip/media_port.go
+++ b/pkg/sip/media_port.go
@@ -58,6 +58,8 @@ type PortStats struct {
 
 	DTMFPackets atomic.Uint64
 	DTMFBytes   atomic.Uint64
+
+	Closed atomic.Bool
 }
 
 type UDPConn interface {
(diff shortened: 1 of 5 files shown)


pkg/sip/media_port.go
```diff
@@ -306,7 +309,7 @@ func (p *MediaPort) timeoutLoop(timeoutCallback func()) {
 			}
 
 			// Ticker is allowed to fire earlier than the full timeout interval. Skip if it's not a full timeout yet.
-			if since < timeout {
+			if since+timeout/10 < timeout {
```
nishadmusthafa, line 312: why are we doing this ?

dennwc: Tests were detecting a timeout that triggers too late. It was off by a few ms, hence this change. Not a big deal if we trigger it later (as it is right now), but this makes tests a bit cleaner.

alexlivekit: > Tests were detecting a timeout that triggers too late.

Is this dues to go scheduling? What's the average wake delay we're working with?

PR #476, opened 2025-10-12
Utilising the provider Info from GetSipTrunkAuthentication to set it …

nishadmusthafa:
…in SipCallInfo


nishadmusthafa:
Landing as the test failures aren't to do with this change. Investigating the test failures separately.

PR #480, opened 2025-10-14
Basic support for reinvites.

dennwc:


nishadmusthafa:
Essentially, we are just responding with an OK if it is a reinvite. We wont be handling complicated cases like hold etc. Looks good to me. Will want @alexlivekit to have a look too.

PR #483, opened 2025-10-15
preventing transfers when PreventTransferForInternalProvider is set t…

nishadmusthafa:
…o true


pkg/sip/service.go
```diff
@@ -348,3 +354,37 @@ func (s *Service) processParticipantTransfer(ctx context.Context, callID string,
 
 	return psrpc.NewErrorf(psrpc.NotFound, "unknown call")
 }
+
+func (s *Service) checkInternalProviderRequest(ctx context.Context, callID string) error {
+	// Look for call both in client (outbound) and server (inbound)
+	s.cli.cmu.Lock()
+	out := s.cli.activeCalls[LocalTag(callID)]
```
dennwc, line 361: Would be great if we can move the validation to `processParticipantTransfer`, so that we don't have to search for the call twice. But I understand you probably don't want to touch the existing logic too much.

pkg/config/config.go
```diff
@@ -100,6 +100,9 @@ type Config struct {
 	EnableJitterBuffer     bool    `yaml:"enable_jitter_buffer"`
 	EnableJitterBufferProb float64 `yaml:"enable_jitter_buffer_prob"`
 
+	// PreventTransferForInternalProvider prevents SIP transfers when the provider is internal and config is set.
+	PreventTransferForInternalProvider bool `yaml:"prevent_transfer_for_internal_provider"`
```
dennwc: Is it something specific to a particular provider? Can IO send this setting instead as a part of `ProviderInfo` (e.g. `ProviderInfo.PreventTransfers`)?

pkg/sip/service.go
nishadmusthafa: decided to not do this now

pkg/sip/service.go
```diff
@@ -381,8 +381,8 @@ func (s *Service) validateCallProvider(state *CallState) error {
 		return nil // No provider info to validate
 	}
 
-	// Check if provider is internal and the prevent transfer config is enabled
-	if state.callInfo.ProviderInfo.Type == livekit.ProviderType_PROVIDER_TYPE_INTERNAL && s.conf.PreventTransferForInternalProvider {
+	// Check if provider is internal and prevent transfer is enabled
+	if state.callInfo.ProviderInfo.Type == livekit.ProviderType_PROVIDER_TYPE_INTERNAL && state.callInfo.ProviderInfo.PreventTransfer {
```
dennwc, line 385: We may remove the internal check here now. It will just honor whatever IO decides for this provider.

PR #486, opened 2025-10-19
Fix opus encoder frame buffering to prevent invalid argument errors

riyaddataqueue:
Implements proper frame buffering in opus encoder to handle partial samples and prevent "opus: invalid argument" errors during RTP processing. The encoder now accumulates input samples until a complete frame is available before encoding, and properly flushes remaining samples on close.


dennwc:
It just copies our own change word-to-word. Not cool :confused:

PR #487, opened 2025-10-20
Disable resampler on the input audio.

dennwc:
Requires https://github.com/livekit/media-sdk/pull/28

### DIFF
--- a/pkg/sip/media_port.go
+++ b/pkg/sip/media_port.go
@@ -131,6 +131,7 @@ type MediaOptions struct {
 	MediaTimeout        time.Duration
 	Stats               *PortStats
 	EnableJitterBuffer  bool
+	NoInputResample     bool
 }
 
 func NewMediaPort(tid traceid.ID, log logger.Logger, mon *stats.CallMonitor, opts *MediaOptions, sampleRate int) (*MediaPort, error) {
@@ -158,6 +159,10 @@ func NewMediaPortWith(tid traceid.ID, log logger.Logger, mon *stats.CallMonitor,
 		conn = c
 	}
 	mediaTimeout := make(chan struct{})
+	inSampleRate := sampleRate
+	if opts.NoInputResample {
+		inSampleRate = -1 // set only after SDP is accepted
+	}
 	p := &MediaPort{
 		tid:              tid,
 		log:              log,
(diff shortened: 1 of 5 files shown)


pkg/sip/media_port.go
```diff
@@ -157,6 +158,10 @@ func NewMediaPortWith(log logger.Logger, mon *stats.CallMonitor, conn UDPConn, o
 		conn = c
 	}
 	mediaTimeout := make(chan struct{})
+	inSampleRate := sampleRate
+	if opts.NoInputResample {
+		inSampleRate = -1 // set only after SDP is accepted
```
alexlivekit, line 164: Wouldn't `msdk.NewSwitchWriter(inSampleRate)` panic if `inSampleRate == -1`?
Edit: No, not with the new change.

pkg/sip/media_port.go
```diff
@@ -167,7 +172,7 @@ func NewMediaPortWith(log logger.Logger, mon *stats.CallMonitor, conn UDPConn, o
 		jitterEnabled:    opts.EnableJitterBuffer,
 		port:             newUDPConn(log, conn),
 		audioOut:         msdk.NewSwitchWriter(sampleRate),
-		audioIn:          msdk.NewSwitchWriter(sampleRate),
+		audioIn:          msdk.NewSwitchWriter(inSampleRate),
```
alexlivekit, line 177: Doesn't this leave the possibility of `audioIn.SapmeRate != audioOut.SampleRate`?

pkg/sip/media_port.go
```diff
@@ -130,6 +130,7 @@ type MediaOptions struct {
 	MediaTimeout        time.Duration
 	Stats               *PortStats
 	EnableJitterBuffer  bool
+	NoInputResample     bool
```
alexlivekit, line 134: So this is not quite disabling resampling as much as delaying the onset of resampling until we have an input sample rate confirmed, correct?

pkg/sip/media_port.go
dennwc: SDP will still negotiate a specific rate for both in and out. This is for our internal use only. Previously it was automatically doing 8/16K->48K resampling when sending to LK, now it won't. Everything else stays the same.

pkg/sip/media_port.go
dennwc: We were not enabling/building the input pipeline until we got an SDP. So event sequence is still the same. But it relies on the existing init order. We know we won't touch it until SDP is ready and the sample rate is set.

pkg/sip/media_port.go
alexlivekit: As long as we tested this case to make sure there isn't some assumption somewhere that these are identical, then all good.

PR #488, opened 2025-10-20
Reuse the same ID for both auth-less and auth-ful INVITEs

alexlivekit:
See [this thread](https://live-kit.slack.com/archives/C067F1RS4RH/p1760806064236849) for more details. Main point:

When we authenticate for the purposes of a SIP session (when credentials are not being cached), the following happens:
- client -> livekit : I want to start a call (INVITE)
- client <- livekit : You need to authenticate (SIP-407 Unauthorized).
  - This may count as a failure.
  - The 407 response contains an auth challenge.
- client -> livekit : I want to start a call (INVITE), this time with the credentials.
  - This call has the exact same callid as the first one.
  - It also contains the auth response.
- client -> livekit : credentials look good, I'll forward your call request (SIP-100 Trying).

If we happen to count the 407 response as a failed call, and we generate unique SCL_* ids for both.

### DIFF
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -40,9 +40,9 @@ import (
 	"github.com/livekit/protocol/livekit"
 	"github.com/livekit/protocol/logger"
 	"github.com/livekit/protocol/rpc"
-	"github.com/livekit/protocol/utils"
 	lksip "github.com/livekit/protocol/sip"
 	"github.com/livekit/protocol/tracer"
+	"github.com/livekit/protocol/utils"
 	"github.com/livekit/protocol/utils/traceid"
 	"github.com/livekit/psrpc"
 	lksdk "github.com/livekit/server-sdk-go/v2"
@@ -201,12 +201,6 @@ func (s *Server) handleInviteAuth(tid traceid.ID, log logger.Logger, req *sip.Re
 		_ = tx.Respond(sip.NewResponseFromRequest(req, 100, "Processing", nil))
 	}
 
-	// Extract SIP Call ID for tracking in-progress invites
-	sipCallID := ""
-	if h := req.CallID(); h != nil {
-		sipCallID = h.Value()
-	}
-
 	h := req.GetHeader("Proxy-Authorization")
 	if h == nil {
 		inviteState.challenge = digest.Challenge{
@@ -333,12 +327,10 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 		return psrpc.NewError(psrpc.MalformedRequest, errors.Wrap(err, "cannot parse source IP"))
 	}
 	sipCallID := legCallIDFromReq(req)
-	tid := traceid.FromGUID(callID)
 	tr := callTransportFromReq(req)
 	legTr := legTransportFromReq(req)
 	log := s.log.WithValues(
 		"sipCallID", sipCallID,
-		"traceID", tid.String(),
 		"fromIP", src.Addr(),
 		"toIP", req.Destination(),
 		"transport", tr,
@@ -390,8 +382,10 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 		inviteProgress.lkCallID = callID
 	}
 	cc.id = LocalTag(callID)
+	tid := traceid.FromGUID(sipCallID)
 
 	log = log.WithValues("callID", callID)
+	log = log.WithValues("traceID", tid.String())
 	cc.log = log
 	log.Infow("processing invite")
 
@@ -399,10 +393,10 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 	defer span.End()
 
 	s.cmu.RLock()
-	existing := s.byCallID[cc.SIPCallID()]
+	existing := s.byCallID[sipCallID]
 	s.cmu.RUnlock()
 	if existing != nil && existing.cc.InviteCSeq() < cc.InviteCSeq() {
-		log.Infow("accepting reinvite", "sipCallID", existing.cc.ID(), "content-type", req.ContentType(), "content-length", req.ContentLength())
+		log.Infow("accepting reinvite", "content-type", req.ContentType(), "content-length", req.ContentLength())
 		existing.log().Infow("reinvite", "content-type", req.ContentType(), "content-length", req.ContentLength(), "cseq", cc.InviteCSeq())
 		cc.AcceptAsKeepAlive(existing.cc.OwnSDP())
 		return nil
@@ -429,7 +423,7 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 
 	callInfo := &rpc.SIPCall{
 		LkCallId:  callID,
-		SipCallId: cc.SIPCallID(),
+		SipCallId: sipCallID,
 		SourceIp:  src.Addr().String(),
 		Address:   ToSIPUri("", cc.Address()),
 		From:      ToSIPUri("", from),
@@ -500,7 +494,7 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 			// We will send password request anyway, so might as well signal that the progress is made.
 			cc.Processing()
 		}
-		s.getCallInfo(cc.SIPCallID()).countInvite(log, req)
+		s.getCallInfo(sipCallID).countInvite(log, req)
 		if !s.handleInviteAuth(tid, log, req, tx, from.User, r.Username, r.Password, inviteProgress) {
 			cmon.InviteErrorShort("unauthorized")
 			// handleInviteAuth will generate the SIP Response as needed
@@ -508,7 +502,7 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 		}
 		// ok
 	case AuthAccept:
-		s.getCallInfo(cc.SIPCallID()).countInvite(log, req)
+		s.getCallInfo(sipCallID).countInvite(log, req)
 		// ok
 	}
 


pkg/sip/inbound.go
```diff
@@ -64,6 +65,8 @@ const (
 	inviteOKRetryAttempts      = 5
 	inviteOKRetryAttemptsNoACK = 2
 	inviteOkAckLateTimeout     = inviteOkRetryIntervalMax
+
+	inviteCredentialValidity = 60 * time.Minute // Allow reuse of credentials for 1h
```
alexlivekit, line 69: Not sure this is smart, might want to extend this to max_call_duration or something.

Also, keep in mind that this is per Call-ID for now, so new calls would still need re-auth.

pkg/sip/inbound.go
```diff
@@ -137,20 +140,23 @@ func (s *Server) getCallInfo(id string) *inboundCallInfo {
 func (s *Server) getInvite(sipCallID string) *inProgressInvite {
 	s.imu.Lock()
 	defer s.imu.Unlock()
-	for i := range s.inProgressInvites {
-		if s.inProgressInvites[i].sipCallID == sipCallID {
-			return s.inProgressInvites[i]
-		}
-	}
-	if len(s.inProgressInvites) >= digestLimit {
-		s.inProgressInvites = s.inProgressInvites[1:]
```
alexlivekit, line 146: While this does keep the slice at `< digestLimit`:
- The underlying storage keeps growing.
- Doesn't quite scale, with possible high-turnover servers. Do we care?
- Not the most efficient iteration

Changed type to map.

pkg/sip/inbound.go
```diff
@@ -178,8 +184,6 @@ func (s *Server) handleInviteAuth(log logger.Logger, req *sip.Request, tx sip.Se
 	}
 	ci := s.getCallInfo(sipCallID)
 	ci.countInvite(log, req)
-	inviteState := s.getInvite(sipCallID)
-	log = log.WithValues("inviteStateSipCallID", sipCallID)
```
alexlivekit, line 180: Note: This is a subtle change - We now have "sipCallID" already populated, so there's no need of "inviteStateSipCallID". However, if we have dashboards that depend on this, we might want to update those.

pkg/sip/inbound.go
```diff
@@ -220,7 +224,6 @@ func (s *Server) handleInviteAuth(log logger.Logger, req *sip.Request, tx sip.Se
 	// Check if we have a valid challenge state
 	if inviteState.challenge.Realm == "" {
 		log.Warnw("No challenge state found for authentication attempt", errors.New("missing challenge state"),
-			"sipCallID", sipCallID,
```
alexlivekit, line 233: Already in.

pkg/sip/inbound.go
```diff
@@ -295,18 +298,18 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 		s.log.Errorw("cannot parse source IP", err, "fromIP", src)
 		return psrpc.NewError(psrpc.MalformedRequest, errors.Wrap(err, "cannot parse source IP"))
 	}
-	callID := lksip.NewCallID()
+	sipCallID := legCallIDFromReq(req)
 	tr := callTransportFromReq(req)
 	legTr := legTransportFromReq(req)
 	log := s.log.WithValues(
-		"callID", callID,
```
alexlivekit, line 313: The one log line that will not have `callID` now is `"Bad request"`, when validation fails.

pkg/sip/inbound.go
```diff
@@ -134,23 +137,27 @@ func (s *Server) getCallInfo(id string) *inboundCallInfo {
 	return c
 }
 
-func (s *Server) getInvite(sipCallID string) *inProgressInvite {
+func (s *Server) getInvite(sipCallID, toTag, fromTag string) *inProgressInvite {
+	key := fmt.Sprintf("%s:%s:%s", sipCallID, toTag, fromTag)
```
alexlivekit: Now using full-blown dialog ID as the key here.

pkg/sip/outbound.go
```diff
@@ -828,7 +828,7 @@ authLoop:
 		if err != nil {
 			return nil, fmt.Errorf("invalid challenge %q: %w", challengeStr, err)
 		}
-		toHeader := resp.To()
```
alexlivekit, line 846: Now that we're doing the right validation on the inbound side, the outbound side E2E test caught this error!
But this also means out clients might run into the same issue, in case some of them are not spec-compliant.

pkg/sip/inbound.go
dennwc: Would be nice to use a struct as a map key here. It avoids extra allocation by `Sprintf`:
```go
type inviteKey struct {
   CallID string
   ToTag string
   FromTag string
}
// ...
key := inviteKey{CallID: sipCallID, ToTag: toTag, FromTag: fromTag}
// ...
is, ok := s.inProgressInvites[key]
```

pkg/sip/inbound.go
dennwc: It won't grow, but it does churn memory by periodically re-allocating. Yeah, we should have changed this to a map a long time ago.

pkg/sip/inbound.go
```diff
@@ -134,23 +137,27 @@ func (s *Server) getCallInfo(id string) *inboundCallInfo {
...
+	if ok {
+		return is
 	}
-	is := &inProgressInvite{sipCallID: sipCallID}
-	s.inProgressInvites = append(s.inProgressInvites, is)
+	is = &inProgressInvite{sipCallID: sipCallID}
+	s.inProgressInvites[key] = is
+
+	go func() {
+		time.Sleep(inviteCredentialValidity)
```
dennwc: Better avoid spawning goroutines that wait for the whole hour without a clear cancellation signal.

Usually, you'd have one goroutine periodically cleaning the expired cache. `time.AfterFunc` could work, but again, there's no real need to create thousands of timers all waiting for an hour.

pkg/sip/inbound.go
dennwc: I don't think we used it anywhere.

pkg/sip/inbound.go
```diff
@@ -330,6 +332,29 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
...
+
+	// Establish ID
+	fromTag, _ := req.From().Params.Get("tag") // always exists, via ValidateInvite() check
+	toTag, ok := req.To().Params.Get("tag")    // To() always exists, via ValidateInvite() check
+	if !ok {
+		// No to-tag on the invite means we need to generate one per RFC 3261 section 12.
+		// Generate a new to-tag early, to make sure both INVITES have the same ID.
+		uuid, _ := uuid.NewV4() // Same as NewResponseFromRequest in sipgo
+		toTag = uuid.String()
+		req.To().Params.Add("tag", toTag)
```
dennwc: This may panic if `Params` is `nil`, I think.

pkg/sip/inbound.go
```diff
@@ -1293,11 +1312,11 @@ func (c *inboundCall) transferCall(ctx context.Context, transferTo string, heade
 
 }
 
-func (s *Server) newInbound(log logger.Logger, id LocalTag, contact URI, invite *sip.Request, inviteTx sip.ServerTransaction, getHeaders setHeadersFunc) *sipInbound {
+func (s *Server) newInbound(log logger.Logger, contact URI, invite *sip.Request, inviteTx sip.ServerTransaction, getHeaders setHeadersFunc) *sipInbound {
 	c := &sipInbound{
 		log:      log,
 		s:        s,
-		id:       id,
+		id:       "unassigned",
```
dennwc: Would be helpful if it logs/panics if something tries to read this unassigned ID.

pkg/sip/inbound.go
```diff
@@ -330,6 +332,29 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
...
+	toTag, ok := req.To().Params.Get("tag")    // To() always exists, via ValidateInvite() check
+	if !ok {
+		// No to-tag on the invite means we need to generate one per RFC 3261 section 12.
+		// Generate a new to-tag early, to make sure both INVITES have the same ID.
+		uuid, _ := uuid.NewV4() // Same as NewResponseFromRequest in sipgo
+		toTag = uuid.String()
+		req.To().Params.Add("tag", toTag)
+	}
+	inviteProgress := s.getInvite(sipCallID, toTag, fromTag)
+	callID := inviteProgress.lkCallID
```
dennwc, line 379: No locking here. I assume our re-invite handling will catch a duplicate invite if it arrives at the same time?

pkg/sip/inbound.go
```diff
@@ -319,8 +323,6 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 	})
 	log = LoggerWithParams(log, cc)
 	log = LoggerWithHeaders(log, cc)
-	cc.log = log
-	log.Infow("processing invite")
```
dennwc, line 335: We will no longer log invalid invites this way. This may hide some failed attempts in the logs.

pkg/sip/inbound.go
```diff
@@ -30,6 +30,7 @@ import (
 
 	msdk "github.com/livekit/media-sdk"
 	"github.com/livekit/protocol/rpc"
+	uuid "github.com/satori/go.uuid"
```
dennwc: We could use our ID generator from the protocol potentially.

pkg/sip/inbound.go
alexlivekit: That's right - locking is done in `getInvite()`, but we're using `inviteProgress.lkCallID` outside of it. 
If two separate INVITEs come along, and they're retransmissions (same via branch), sipgo swallows it. If they're not retransmissions, these are separate invites and should be processed independently, which due to the early to-tag generation here they would (different toTag = different map key)

pkg/sip/inbound.go
alexlivekit: I would rather not add shims or verification for this. Unless it's real clean, I'd rather leave it blank. Would that be better in your opinion?

pkg/sip/inbound.go
alexlivekit: There's something in specs about being securely random, so I originally figured "just keep what sipgo is doing" rather than proof reading uuid generators. I guess it's time.

1. Current sipgo upstream: github.com/satori/go.uuid github.com/google/uuid
2. Our sipgo version: github.com/satori/go.uuid
3. Protocol: our [in-house generator](https://github.com/livekit/protocol/blob/4c4d1f4c8da9fb9a96b49c170e6ecf0962b9b46b/utils/guid/id.go#L134).

Apparently, github.com/satori/go.uuid is a security risk. So lets go with our in-house stuff for now.

pkg/sip/server.go
```diff
@@ -132,8 +139,10 @@ type Server struct {
 	sipListeners []io.Closer
 	sipUnhandled RequestHandler
 
-	imu               sync.Mutex
-	inProgressInvites []*inProgressInvite
+	imu                  sync.Mutex
+	inProgressInvites    map[dialogKey]*inProgressInvite
+	inviteTimeoutQueue   utils.TimeoutQueue[*dialogKey]
```
```suggestion
	inviteTimeoutQueue   utils.TimeoutQueue[dialogKey]
```
This must be without a pointer. Otherwise you are using the pointer address as a queue key, so it will never match.

pkg/sip/inbound.go
```diff
@@ -134,23 +137,44 @@ func (s *Server) getCallInfo(id string) *inboundCallInfo {
...
-		s.inProgressInvites = s.inProgressInvites[1:]
+}
+
+func (s *Server) getInvite(sipCallID, toTag, fromTag string) *inProgressInvite {
+	key := dialogKey{
+		sipCallID: sipCallID,
+		toTag:     toTag,
+		fromTag:   fromTag,
+	}
+	s.imu.Lock()
```
dennwc: A common practice to avoid lock contention is to use `RWMutex` and doing a two stage lock:
```go
s.imu.RLock()
is, ok := s.inProgressInvites[key]
s.imu.RUnlock()
if ok {
   return is
}
s.imu.Lock()
defer s.imu.Unlock()
is, ok := s.inProgressInvites[key]
if ok {
   return is
}
// ... the rest ...
```
This allows multiple readers to get the invite state without blocking each other. Also notice that we redo the check after getting a write lock - other routine might create the state earlier.

pkg/sip/server.go
```diff
@@ -132,8 +139,10 @@ type Server struct {
...
+	isCleanupTaskRunning atomic.Bool
```
dennwc: Seem unused

pkg/sip/server.go
```suggestion
	inviteTimeoutQueue   utils.TimeoutQueue[dialogKey]
```
We probably don't need a pointer here.

pkg/sip/inbound.go
```diff
@@ -134,23 +137,44 @@ func (s *Server) getCallInfo(id string) *inboundCallInfo {
...
+	defer s.imu.Unlock()
+	is, ok := s.inProgressInvites[key]
+	if ok {
+		return is
 	}
-	is := &inProgressInvite{sipCallID: sipCallID}
-	s.inProgressInvites = append(s.inProgressInvites, is)
+	is = &inProgressInvite{sipCallID: sipCallID}
+	s.inProgressInvites[key] = is
+	s.inviteTimeoutQueue.Reset(&utils.TimeoutQueueItem[*dialogKey]{Value: &key})
```
dennwc: Looks like we do not reset the cache expiry time if we have a cache hit. Is it intentional?

pkg/sip/server.go
dennwc: There's a convention in Go called a [Mutex hat](https://dmitri.shuralyov.com/idiomatic-go/entries/2), which implies `inviteTimeoutQueue` is protected by `imu`.

But, the queue implementation already has a mutex internally. So might be worth moving it out of the `imu` group.

alexlivekit: Oh, this is neat, thank you for poinitng it out!

pkg/sip/inbound.go
alexlivekit: This pattern:
- Adds a few lines of code - maintenance complexity over time.
- Adds a few operations - Every first invite will have to go through two lock/unlock cycles (userspace, but still not 0us). 
- Optimization? - Does it save time in this case? Sipgo de-dupes retransmissions of all requests, and in-dialog INVITEs are not possible before we conclude the initial invite. So the only possible race condition here is two separate re-INVITE requests that are not retransmissions. In that case, the flow is `lock` -> `check exists` -> `unlock` -> `return existing` with or without this pattern. What am I missing here? 
- Consistency - If we just always do it that way, that make a solid practice and approach.

My initial approach was "keep it simple", I'll happily change that for consistency's sake, but I would love to make sure I understand the performance aspect of this.

dennwc: My main concern that under load we've seen a pattern where provider keeps retrying invites, and somehow they may still pass through sipgo. In that case each invite will temporarily block any other valid invites because of the mutex, which will make things worse. With read locks they will not block other calls, since we already have an invite state for that call ID.

Note that the first lock sequence is not `Lock/Unlock`, it's `RLock/RUnlock`.

PR #489, opened 2025-10-20
Fix outbound destination setting to honor Route headers

alexlivekit:


pkg/sip/outbound.go
```diff
@@ -889,15 +888,14 @@ func (c *sipOutbound) AckInviteOK(ctx context.Context) error {
...
-func (c *sipOutbound) attemptInvite(ctx context.Context, callID sip.CallIDHeader, dest string, to *sip.ToHeader, offer []byte, authHeaderName, authHeader string, headers Headers, setState sipRespFunc) (*sip.Request, *sip.Response, error) {
+func (c *sipOutbound) attemptInvite(ctx context.Context, callID sip.CallIDHeader, to *sip.ToHeader, offer []byte, authHeaderName, authHeader string, headers Headers, setState sipRespFunc) (*sip.Request, *sip.Response, error) {
 	ctx, span := tracer.Start(ctx, "sipOutbound.attemptInvite")
 	defer span.End()
 	req := sip.NewRequest(sip.INVITE, to.Address)
 	c.setCSeq(req)
 	req.RemoveHeader("Call-ID")
 	req.AppendHeader(&callID)
 
-	req.SetDestination(dest)
```
alexlivekit, line 900: Removed, it now defaults to `req.Destination()`, which already respects `Route` headers.
This looks like it was introduced in the first [iteration of outbound calling altogether](https://github.com/livekit/sip/commit/ec274a008f458df2c420a0dac21ff9cca6e4697c).
The one possible downside I see to this is we will potentially re-resolve if DNS cache expires exactly in-between attempt 1 and any following re-transmit/re-auth attempt, causing an extra bit of latency.
However. I think this is  still the right thing to do here.

alexlivekit: Update: Denys recalled the possible reason for this was to avoid DNS re-resolution mid call, possibly arriving to a different host without the previous INVITE credentials.

That is absolutely a valid concern,m but not for this change - `req.SetDestination()` simply stores a string, so we're subject to DNS re-resolution today, with  SetDestination or without.

dennwc (approved):

PR #496, opened 2025-10-24
Reduce 5xx spam

alexlivekit:
Today we return `psrpc.Canceled`, which then gets [translated to HTTP code 500](https://github.com/livekit/psrpc/blob/bdebea7dacf4fcf3a735e98fc4dc04852cfb43e4/errors.go#L84-L85). This can be quite problematic when we get return codes like: `480 Temporarily Unavailable`, `603 Decline`, `486 Busy Here`, `404 Not Found`. These are either user errors, outright rejections, or numbers incapable of handling two simultaneous calls.

This change will get us to emit `UpstreamServerError` and `UpstreamClientError` as responses, that would translate to HTTP codes 502 and 424 respectively, correctly identifying failures in resources upstream of the sip server.

### DIFF
--- a/pkg/sip/protocol.go
+++ b/pkg/sip/protocol.go
@@ -375,12 +375,16 @@ func handleReferNotify(cseq uint32, status int, reason string, referCseq uint32,
 			Code:   livekit.SIPStatusCode(status),
 			Status: reason,
 		}
-		psrpcErrCode := psrpc.UpstreamServerError
-		if status < 500 || status >= 600 {
-			// Common 6xx codes: 603 Declined, 608 Rejected
-			psrpcErrCode = psrpc.UpstreamClientError
+		// Converts SIP status to GRPC via SIPStatus.GRPCStatus(), then converts to psrpc via ErrorCodeFromGRPC()
+		errorCode, _ := psrpc.GetErrorCode(st)
+		if errorCode == psrpc.Internal || errorCode == psrpc.Unavailable {
+			// Temporarily overwrite the code until we support a direct SIPStatus -> psrpc.ErrorCode conversion
+			errorCode = psrpc.UpstreamServerError
+			if status < 500 || status >= 600 { // Common 6xx codes: 603 Declined, 608 Rejected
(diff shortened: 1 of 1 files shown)


pkg/sip/protocol.go
```diff
@@ -375,7 +375,12 @@ func handleReferNotify(cseq uint32, status int, reason string, referCseq uint32,
 			Code:   livekit.SIPStatusCode(status),
 			Status: reason,
 		}
-		result = psrpc.NewErrorf(psrpc.Canceled, "call transfer failed: %w", st)
+		psrpcErrCode := psrpc.UpstreamServerError
```
alexlivekit: Going over twirp, this will still produce a bad error, due to [Twirp not having the appropriate response codes](https://github.com/livekit/psrpc/blob/bdebea7dacf4fcf3a735e98fc4dc04852cfb43e4/errors.go#L234-L235) to convert this to. However, data indicates the vast majority of errors are not 5XX errors and are going to be `UpstreamClientError`. which would convert to `twirp.InvalidArgument`.

davidzhao: this isn't related to Twirp, but our error code mapping in psrpc? why don't we just map it to the right thing there?

dennwc: This completely ignores the [existing mapping](https://github.com/livekit/protocol/blob/main/livekit/sip.go#L49) to gRPC/PSRPC/Twirp error codes that we have. Maybe we should update the mapping there instead?

alexlivekit: If you refer to the [code here ](https://github.com/livekit/psrpc/blob/bdebea7dacf4fcf3a735e98fc4dc04852cfb43e4/pkg/server/rpc.go#L264-L272), I suspect we never reach the unwrap point since `if errors.As(err, &e) {` succeeds, due to us returning a psrpc.Error.
If I read this right `SIPStatus.GRPCStatus()` gets called when we don't return a psrpc error, but a SIPStatus instead.

dennwc (approved):

PR #502, opened 2025-10-31
Added test with cancel

alexlivekit:


nishadmusthafa (approved):

pkg/sip/inbound.go
```diff
@@ -633,6 +632,22 @@ func (s *Server) newInboundCall(
...
+	ptr := c.logPtr.Load()
+	if ptr == nil {
+		return nil
+	}
+	return *ptr
+}
+
+func (c *inboundCall) appendLogValues(kvs ...any) {
+	c.setLog(c.log().WithValues(kvs...))
+}
```
alexlivekit, line 657: Does this solve the race?
You can still have two threads getting the same logger via `c.log()...`, creating a derivative via `.WithValues(kvs...)`, and racing to store their version via ` c.setLog()`. What am I missing?

Having said that, it should be just fine landing this since this is an improvement and carries no degradation.

PR #504, opened 2025-11-05
Adding basic metrics for transfer

nishadmusthafa:


pkg/stats/monitor.go
```diff
@@ -60,20 +60,23 @@ const (
...
+	callsTerminated    *prometheus.CounterVec
+	packetsRTP         *prometheus.CounterVec
+	durSession         *prometheus.HistogramVec
+	durCall            *prometheus.HistogramVec
+	durJoin            *prometheus.HistogramVec
+	durCheck           *prometheus.HistogramVec
+	cpuLoad            prometheus.Gauge
+	sdpSize            *prometheus.HistogramVec
+	nodeAvailable      prometheus.GaugeFunc
+	transfersTotal     *prometheus.CounterVec
```
dennwc, line 77: We could also add active transfers metric as a gauge and inc/dec it in the existing helper functions.

pkg/sip/service.go
```diff
@@ -388,3 +394,49 @@ func (s *Service) validateCallProvider(state *CallState) error {
...
+	// Check for context errors
+	if errors.Is(err, context.Canceled) {
+		return "canceled"
+	}
+	if errors.Is(err, context.DeadlineExceeded) {
+		return "timeout"
+	}
+
+	// Extract error message and normalize
+	errMsg := err.Error()
```
dennwc: We usually just put `"other"` here and look at the logs to see if we should add a separate reason for it.

pkg/sip/service.go
```diff
@@ -388,3 +394,49 @@ func (s *Service) validateCallProvider(state *CallState) error {
...
+
+// extractTransferErrorReason extracts a user-friendly reason string from an error
+func extractTransferErrorReason(err error) string {
+	if err == nil {
+		return "unknown"
+	}
+
+	// Check for psrpc errors
+	var psrpcErr psrpc.Error
+	if errors.As(err, &psrpcErr) {
```
dennwc, line 414: We should probably check for the `livekit.SIPStatus` first before falling back to PSRPC error codes.

pkg/sip/service.go
```diff
@@ -344,15 +347,19 @@ func (s *Service) processParticipantTransfer(ctx context.Context, callID string,
...
 			return err
 		}
-
+		s.mon.TransferSucceeded(stats.Inbound)
 		return nil
 	}
 
-	return psrpc.NewErrorf(psrpc.NotFound, "unknown call")
+	err := psrpc.NewErrorf(psrpc.NotFound, "unknown call")
+	s.mon.TransferFailed(stats.Inbound, "unknown_call")
```
dennwc: This decrements transfer stat without incrementing it first.

pkg/sip/service.go
```diff
@@ -388,3 +395,69 @@ func (s *Service) validateCallProvider(state *CallState) error {
...
+// extractTransferErrorReason extracts a user-friendly reason string from an error
+func extractTransferErrorReason(err error) string {
+	if err == nil {
+		return "unknown"
+	}
+
+	// Check for livekit.SIPStatus errors first
+	var sipStatus *livekit.SIPStatus
+	if errors.As(err, &sipStatus) {
+		switch int(sipStatus.Code) {
```
dennwc: Just a nit, but you could check against `livekit.SIP_STATUS_*` instead of the sipgo constants to avoid `int` casts.

pkg/sip/service.go
```diff
@@ -388,3 +395,69 @@ func (s *Service) validateCallProvider(state *CallState) error {
...
+		case int(lksip.StatusBadRequest):
+			return "bad_request"
```
dennwc: I think there's also a `Name()` or `String()` method, so maybe we could just use that instead.

PR #512, opened 2025-11-08
Better SDP errors and logs.

dennwc:


pkg/sip/inbound.go
```diff
@@ -720,8 +720,24 @@ func (c *inboundCall) handleInvite(ctx context.Context, tid traceid.ID, req *sip
...
-		answerData, err := c.runMediaConn(tid, req.Body(), enc, conf, disp.EnabledFeatures)
+		log := c.log()
+		if h := req.ContentLength(); h != nil {
+			log = log.WithValues("contentLength", int(*h))
+		}
+		if h := req.ContentType(); h != nil {
+			log = log.WithValues("contentType", h.Value())
+			switch h.Value() {
+			default:
+				log.Infow("unsupported offer type")
```
nishadmusthafa, line 731: Is there any point continuing if it's not "application/sdp" ?

nishadmusthafa (approved):

pkg/sip/inbound.go
dennwc: Not really. But I want to see first that this doesn't happen in the wild first.

PR #516, opened 2025-11-11
Drop preset Route headers before creating the ACK 

alexlivekit:


pkg/sip/outbound.go
```diff
@@ -863,6 +868,10 @@ authLoop:
 		}
 	}
 
+	// We currently don't plumb the request back to caller to construct the ACK with.
+	// Thus, we need to modify the request to update any route sets.
+	for req.RemoveHeader("Route") {
+	}
```
alexlivekit, line 885: This is the actual fix, everything else is either new test shims or incidentilomas.

pkg/sip/outbound_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestOutboundRouteHeaderWithRecordRoute(t *testing.T) {
+	// Make sure the ACK doesn't carry over initial Route header.
+	// Steps:
+	// 1. Create a SIP participant with an initial Route header.
+	// 2. Make sure the Route header is properly populates in INVITE.
+	// 3. Fake a 200 response with Record Route headers.
+	// 4. Make sure the ACK doesn't carry over initial Route header..
```
alexlivekit, line 33: This is the test added to make sure we're doing this right.

pkg/sip/outbound_utilities_test.go
alexlivekit, line 1: All of this file is test utilities that basically mock out sipgo.client, lk.Room, IOClient, and a psrpc service.

pkg/sip/room.go
```diff
@@ -61,6 +61,37 @@ type ParticipantInfo struct {
...
+
+var newRoomFunc NewRoomFunc = func(log logger.Logger, st *RoomStats) RoomInterface {
+	return NewRoom(log, st)
+}
+
+func SetNewRoomFunc(fn NewRoomFunc) (oldFunc NewRoomFunc) {
+	old := newRoomFunc
+	newRoomFunc = fn
+	return old
+}
```
alexlivekit: Room shim definition & override function.

test/integration/livekit_test.go
```diff
@@ -64,6 +64,7 @@ func runLiveKit(t testing.TB) *LiveKit {
 			//       Maybe run on the host network instead? We might need it for RTP anyway.
 			"--redis-host", dockerBridgeIP + ":" + port,
 			"--bind", "0.0.0.0",
+			"--rm",
```
alexlivekit: Nice-to-have

test/integration/sip_test.go
```diff
@@ -912,7 +912,7 @@ func TestSIPOutbound(t *testing.T) {
 						},
 					})
 					t.Cleanup(func() {
-						srvIn.DeleteTrunk(t, trunkOut)
+						srvOut.DeleteTrunk(t, trunkOut)
```
alexlivekit, line 915: minor cleanup fix.

pkg/sip/outbound.go
```diff
@@ -372,7 +377,7 @@ func (c *outboundCall) connectToRoom(ctx context.Context, lkNew RoomConfig) erro
 
 	attrs[livekit.AttrSIPCallStatus] = CallDialing.Attribute()
 	lkNew.Participant.Attributes = attrs
-	r := NewRoom(c.log, &c.stats.Room)
+	r := newRoomFunc(c.log, &c.stats.Room)
```
alexlivekit: Overwriting this is what allows us to mock out Room.

pkg/sip/client.go
```diff
@@ -38,14 +38,55 @@ import (
...
+	}
+	return &sipgoClientWrapper{client: client}, nil
+}
+
+// SetNewClientFunc allows tests to override the client creation function.
+func SetNewClientFunc(fn NewClientFunc) (oldfunc NewClientFunc) {
+	old := newClientFunc
+	newClientFunc = fn
+	return old
+}
```
alexlivekit, line 54: sipgo.Client shim definition & overwrite method.

pkg/sip/outbound_test.go
```diff
@@ -0,0 +1,125 @@
...
+	defer cancel()
+	go func() { // Allow test to continue
+		_, err := client.CreateSIPParticipant(ctx, req)
+		if err != nil && ctx.Err() == nil {
+			// Only log error if context wasn't cancelled
+			t.Logf("CreateSIPParticipant error: %v", err)
+		}
+	}()
+
+	fmt.Println("Waiting for INVITE to be sent")
```
dennwc:
```suggestion
	t.Log("Waiting for INVITE to be sent")
```

pkg/sip/outbound_utilities_test.go
```diff
@@ -0,0 +1,558 @@
...
+	lksdk "github.com/livekit/server-sdk-go/v2"
+	"github.com/livekit/sip/pkg/config"
+	"github.com/livekit/sip/pkg/stats"
+)
+
+// MockIOInfoClient is a no-op implementation of rpc.IOInfoClient for testing
+type MockIOInfoClient struct{}
+
+// Egress methods
+func (m *MockIOInfoClient) CreateEgress(ctx context.Context, req *livekit.EgressInfo, opts ...psrpc.RequestOption) (*emptypb.Empty, error) {
```
dennwc, line 48: I think there's some default `UnimplementedIOInfoClient` that already contains all of these.

pkg/sip/client.go
```diff
@@ -89,7 +129,7 @@ func (c *Client) Start(agent *sipgo.UserAgent, sc *ServiceConfig) error {
 	}
 
 	var err error
-	c.sipCli, err = sipgo.NewClient(agent,
+	c.sipCli, err = newClientFunc(agent,
```
dennwc: Could it be a hidden option on the `ServiceConfig`? Or an optional argument to `Start`:
```go
type ClientStartOptions struct {
   NewClientFunc NewClientFunc
}

type ClientStartOption func(opts *ClientStartOptions)

func WithNewClientFunc(fnc NewClientFunc) ClientStartOption {
   return func(opts *ClientStartOptions) {
      opts.NewClientFunc = fnc
   }
}

func (c *Client) Start(agent *sipgo.UserAgent, sc *ServiceConfig, opts ...ClientStartOption) error {
   var copts ClientStartOptions
   for _, opt := range opts {
      opt(&copts)
   }
```

pkg/sip/outbound_utilities_test.go
alexlivekit: Couldn't find it. Can you please point to the code?

pkg/sip/client.go
alexlivekit: sure, it could.
I deliberately didn't want to add more fields for test-only purposes, and also didn't want it to be a per-client thing.
So I'd love to understand the rationale behind wanting it done like that.

dennwc: Avoiding global state. Go tests can run in parallel, which will break an assumptions here (that the client can be received from the channel).

If your main concern is a new public API introduced this way, you could easily unexport everything I proposed (e.g. `Start` calls `start`, which has extra options).

pkg/sip/client.go
```diff
@@ -39,36 +38,95 @@ import (
...
 
+// An interface mirroring sipgo.Client to be able to mock it in tests.
+type SIPClient interface {
+	TransactionRequest(req *sip.Request) (sip.ClientTransaction, error)
+	WriteRequest(req *sip.Request) error
+	Close() error
+}
+
+// Passthrough implementation of SIPClient with a sipgo.Client.
+type sipgoClientWrapper struct {
```
dennwc: Do we need a wrapper here? Can we pass `*sipgo.Client` directly as an implementation? Looks like the method signatures match exactly.

pkg/sip/outbound.go
```diff
@@ -874,6 +868,10 @@ authLoop:
 		}
 	}
 
+	// We currently don't plumb the request back to caller to construct the ACK with.
+	// Thus, we need to modify the request to update any route sets.
+	for req.RemoveHeader("Route") {
```
dennwc, line 884: Does it need to be conditional based on the existence of `Record-Route`? For example, if response contains none, should we keep our routing table? But I guess in that case the `Contact` will be set at least.

pkg/sip/client.go
alexlivekit: ~~Unfortunately I have not managed to find a clean way to do this, compiler doesn't like this. I assume something needs to create the interface first.~~
Nevermind, I thought you meant the function, not the interface itself. Yes. Done.

PR #518, opened 2025-11-12
pkg/sip/outbound.go: allow overwriting autogenerated SIP headers with ones suppiled by the user

colpari:
This allows users to override generated SIP headers and enables to achieve custom compatibility with a wide range of SIP-Servers/PBX-Systems.

For us this is a blocker for using LiveKit Cloud because it does not work with our existing telephony infrastructure.

Running livekip/sip locally with this change, we are able to place calls with our proprietary PBX system by constructing the From header ourselves.

### DIFF
--- a/pkg/sip/outbound.go
+++ b/pkg/sip/outbound.go
@@ -919,6 +919,9 @@ func (c *sipOutbound) attemptInvite(ctx context.Context, callID sip.CallIDHeader
 		req.AppendHeader(sip.NewHeader(authHeaderName, authHeader))
 	}
 	for _, h := range headers {
+		if req.GetHeader(h.Name()) != nil {
+			req.RemoveHeader(h.Name())
+		}
 		req.AppendHeader(h)
 	}
 


dennwc (changes requested):
I'm sorry, but we cannot allow changing a few core headers in our SIP deployments. This should either be an optional flag or a custom patch in your fork. Otherwise it may interfere with routing and we won't be able to provide any support for it.

Having said that, we are open for discussions if you want to be able to control specific parts of the `From` header. For example, recently we [exposed an option](https://github.com/livekit/protocol/blob/main/protobufs/livekit_sip.proto#L692) for setting a display name for the `From` URI. Is there something specific you want to configure there?

colpari:
@dennwc Thanks for the direct reply! A patch in our fork is what we obviously have but if i get it correctly, we cannot use LiveKit cloud with a local livekip-sip component, right?

How would you imagine this "optional flag" variant to be implemented? A flag in sipOutboundConfig that enables header overwriting or a second set of (unsupported) header modifications being passed down? We'd be happy to implement it :)

In our specific case, the PBX expects the From Header Address to be `<extension_number@pbx_fqdn>` while livekit-sip  sends `<configured_phone_number@host_ip>`.
So what would also work for us, were if livekit-sip would not add `@host_ip` if `configured_phone_number` already contains an `@`, as we can configure the phone number to be `extension_number@pbx_fqdn` (which currently ends up as `<extension_number@pbx_fqdn@host_ip>`

From our experience with hundreds of quirky implementations and thousands of quirky setups of SIP out there, we'd suggest that a general "customize everything and be unsupported, because you know what you are doing"-solution would be preferable over adding more and more options trying to reach compatibility for specific details and situations. :)

dennwc:
> ... we cannot use LiveKit cloud with a local livekip-sip component, right?

That's correct. We could enable it for inbound pretty quickly, but the outbound part will require a more significant work on our side.

> How would you imagine this "optional flag" variant to be implemented?

I was assuming you want this in OSS, in which case it would be in the SIP config file. But for Cloud, we just cannot add it, since that assumes a certain level of support.

> In our specific case, the PBX expects the From Header Address to be `<extension_number@pbx_fqdn>` while livekit-sip sends `<configured_phone_number@host_ip>`.

You may have noticed that if you put `configured_phone_number=extension_number` and `host_ip=pbx_fqdn` it may just work as-is :wink:

There's nothing forcing you to use an IP instead of FQDN and nothing preventing you from putting some extension number in the number field. If that's all you need - it's possible in Cloud right now.

> customize everything and be unsupported, because you know what you are doing

Sure, for OSS that's what we are trying to do. But for the Cloud, we might be a bit more opinionated than that :slightly_smiling_face:

colpari:
Partially understanding your opinion on the cloud, although it will force us out of using it.
Why not say "customizing SIP headers only works when using your own, external PBX + you loose support for that"? :grin: 

> You may have noticed that if you put `configured_phone_number=extension_number` and `host_ip=pbx_fqdn` it may just work as-is 😉
> There's nothing forcing you to use an IP instead of FQDN and nothing preventing you from putting some extension number in the number field. If that's all you need - it's possible in Cloud right now.

Not completely sure what you mean, maybe i'm blind at the moment...
livekip-sip always adds `@host_ip` (local or external address of the server/container running livekit-sip) which breaks our neck.
Otherwise we'd be fine setting `phone_number` to our desired `extension_number@pbx_fqdn`.

We're not aware how to influence the `host_ip` part of the From-Header. By a quick look at the code it looks like that comes from  `ServiceConfig.SignalingIP` which we could only override by `config.Config.NAT1To1IP`. But since it in our case would point to a totally different host/ip that does not seem right, is it?
If it's only used for the from header but not for actual networking it could work, though...

dennwc:
Ah, okay, I see now. I assume you do not own that FQDN, right? Cloud does support custom domain names for SIP and will populate From based on it.

colpari:
No, we do not own `pbx_fqdn` :pensive: It's maintained by our provider and only has an A record pointing to the on-premise PBX server IPv4 (or one of it's failovers) at any time.

colpari:
> Cloud does support custom domain names for SIP and will populate From based on it.

We're trying to figure out how to use a "custom domain name" in LiveKit cloud but we cannot find any options or docs for it. Can you give us a hint?

davidzhao:
> > Cloud does support custom domain names for SIP and will populate From based on it.
> 
> We're trying to figure out how to use a "custom domain name" in LiveKit cloud but we cannot find any options or docs for it. Can you give us a hint?

please contact me at dz at livekit dot io. this is an enterprise feature that we've enabled for some customers

dennwc:
This should help for OSS: https://github.com/livekit/sip/pull/702

PR #519, opened 2025-11-13
Send SDP with 200 OK for a reinvite.

dennwc:


pkg/sip/inbound.go
```diff
@@ -1583,8 +1591,14 @@ func (c *sipInbound) accepted(inviteOK *sip.Response) {
...
 
-func (c *sipInbound) AcceptAsKeepAlive() {
-	c.RespondAndDrop(sip.StatusOK, "OK")
+func (c *sipInbound) AcceptAsKeepAlive(sdp []byte) {
+	c.respondWithData(sip.StatusOK, "OK", "application/sdp", sdp)
+}
+
+func (c *sipInbound) OwnSDP() []byte {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
```
alexlivekit, line 1600: This is just here to explicitly create a memory barrier, right?

dennwc: Right, the slices is ~24 bytes wide, so it won't be an atomic read.

But in general, in Go it's better to always lock things. Compiler could be aggressive when optimizing and it's safer to assume that the state will be used by other goroutines sooner or later. It may lead to very subtle bugs if we rely on implicit atomic reads.

PR #527, opened 2025-11-21
add udp jumboframe support

zusrut:


pkg/config/config.go
```diff
@@ -80,6 +80,8 @@ type Config struct {
 	NAT1To1IP     string `yaml:"nat_1_to_1_ip"`
 	ListenIP      string `yaml:"listen_ip"`
 
+	// ENABLE JUMBO FRAME FOR UDP
+	UDPJumboFrame bool `yaml:"udp_jumbo_frame"`
```
dennwc: Maybe just expose it as `udp_mtu_size` instead?

zusrut: There’s not much point in managable limit, since it only skipping the packet and logging an error. And there’s no handler for managing oversized data anyway.
Also, keep in mind how the transport package uses this value — it currently subtracts 200 from the initial limit, so you don’t have full control if the package changes in the future.

dennwc: I renamed it to `UDPMaxPayload` to indicate that it's not an MTU specifically. We may use it later and propagate it down to the sipgo package.

zusrut: hello @dennwc  - good will it be merged?

PR #531, opened 2025-11-26
Optimize Docker image size by using smaller base image

jjatinggoyal:
The final base image in the Dockerfile can be trimmed by using a lightweight debian-slim image, since go and full debian image are not required in the final image.

Size Comparison:

|| Before | After |
| :--- | :---: | :---: |
| Compressed | 336MB | 55MB |
| Uncompressed | 970MB | 162MB |

The new image is ~83% smaller than the older image.


jjatinggoyal:
@dennwc can you please review this PR?

dennwc:
PR looks good, I just wanted to check if TLS root certs are still available in the final image.

jjatinggoyal:
@dennwc added the `ca-certificates` package. Also tested the final image in our self hosted setup for inbound SIP calls, everything works well.

PR #538, opened 2025-12-05
Implementing psrpc mapping to proper codes

alexlivekit:


pkg/errors/errors.go
```diff
@@ -26,3 +29,14 @@ var (
 func ErrCouldNotParseConfig(err error) psrpc.Error {
 	return psrpc.NewErrorf(psrpc.InvalidArgument, "could not parse config: %v", err)
 }
+
+// If the provided error wraps a livekit.SIPStatus, this will apply the error code mapping to the resulting error
+func ApplySIPStatus(err error) error {
```
dennwc, line 34: Could you please elaborate why this is needed? In theory, the default error handle of PSRPC should unwrap `SIPStatus`, since it implements `GRPCStatus()`. So I wonder, why do we have to force PSRPC conversion here?

alexlivekit: The short version is when we read a `livekit.SIPStatus` error from a transport on `cloud-controller`, `errors.As(err, &sipStatus)` doesn't work.

The longer version has to do with how we're serializing and de-serializing wrapped errors. Other options include rewriting how we send errors (hard pass on this), hacking in some `if strings.Contains(err, "sip status: ") { code, text := go_extract_sip_code(err) }` into mainline cloud-controller.

I actually think this is absolutely the right thing to do. The native error of psrpc is a psrpc error, and we should be emitting these natively. I'd agree that it feels odd to do `error > SIPStatus > grpcCode > psrpcError`, but imo this is still a good step in the right direction.

PR #542, opened 2025-12-11
Allow setting a source port range for TCP dial.

dennwc:


pkg/sip/service.go
```diff
@@ -214,6 +213,14 @@ func (s *Service) Start() error {
 		sipgo.WithUserAgent(UserAgent),
 		sipgo.WithUserAgentLogger(slog.New(logger.ToSlogHandler(s.log))),
 	}
+	if tconf := s.conf.TCP; tconf != nil {
```
alexlivekit, line 216: nit: shouldn't we check for non-zero values as well?

dennwc: No, the library takes care of the zero port values itself.

nishadmusthafa (approved):

PR #548, opened 2025-12-16
Add inbound call info callback.

dennwc:


pkg/sip/server.go
```diff
@@ -124,6 +124,7 @@ type Handler interface {
 	RegisterTransferSIPParticipantTopic(sipCallId string) error
 	DeregisterTransferSIPParticipantTopic(sipCallId string)
 
+	OnInboundInfo(log logger.Logger, callInfo *rpc.SIPCall, headers Headers)
```
alexlivekit, line 127: Why not have something more generic like `OnRemoteHeaders()`, applicable to both outbound and inbound?

The inbound call can stay where it is, and the outbound can happen on 200 (or any non-100 reponse)

dennwc: There's different logic here for inbound vs outbound. I could add both hooks now, but they won't do anything for now.

alexlivekit: okay, would be nice, but not a showstopper.
Lets just remember to rename this to something that may apply to both directions once we introduce that.

alexlivekit (approved):

PR #557, opened 2026-01-06
Handle silence supression in streams

alexlivekit:


pkg/sip/media_port.go
```diff
@@ -727,6 +728,121 @@ func (p *MediaPort) setupOutput(tid traceid.ID) error {
...
+	isSilenceSupression, missingFrameCount := h.isSilenceSuppression(header)
+
+	if isSilenceSupression {
+		h.gapCount++
+		h.gapSizeSum += uint64(missingFrameCount)
+		if time.Since(h.lastPrintTime) > 15*time.Second {
+			h.lastPrintTime = time.Now()
+			h.log.Infow("timestamp gap",
+				"rtpSeq", header.SequenceNumber,
+				"rtpTimestamp", header.SequenceNumber,
```
milos-lk: minor: `"rtpTimestamp", header.Timestamp,` ?

pkg/sip/media_port.go
```diff
@@ -727,6 +728,121 @@ func (p *MediaPort) setupOutput(tid traceid.ID) error {
...
+
+	// A key characteristic of DTX or silence supression is no sequence gaps, but >1 frame TS gaps
+
+	if seqDiff != 0 || tsDiff == 0 {
+		// This also filters out out-of-order packets, due to seqDiff != 0
+		return false, 0
+	}
+
+	// Silence supression happened - sequential packets (no loss), but with a gap in timestamp
+	missedFrames := int(tsDiff) / int(h.samplesPerFrame)
```
milos-lk, line 806: tsDiff is calculated based on expectedTS which is calculated based on samplesPerFrame - does this work for codecs where no fixed packetization is guaranteed (e.g opus)?  
I was worried about this when creating that draft PR - the way I reasoned about it is adding WriteSample method here: https://github.com/livekit/sip/pull/551/files#diff-0e5ab3c9dae3a6f34f94fa0e1a41d2a0fe9e5f27f004d712c8eb9657f1d81e5dR798
(this only works if the HandleRTP->...WriteSample() chain of calls is sync (could probably be made working in async scenarios as well - but thought we don't have that))

pkg/sip/media_port.go
```diff
@@ -727,6 +728,121 @@ func (p *MediaPort) setupOutput(tid traceid.ID) error {
...
+	return true, missedFrames
+}
+
+func (h *silenceFiller) fillWithSilence(framesToFill int) error {
+	for ; framesToFill > 0; framesToFill-- {
+		silence := make(msdk.PCM16Sample, h.samplesPerFrame)
```
milos-lk, line 812: nit: we could probably have it as a struct member and allocate it once to avoid taxing GC

pkg/sip/media_port_test.go
```diff
@@ -542,3 +542,226 @@ func TestMediaTimeout(t *testing.T) {
...
+		}
+	}
+	for _, index := range indexes { // Make sure we're not missing any indexes
+		if index < 0 || index >= expectedSize {
+			t.Fatalf("index %d out of range", index)
+		}
+	}
+}
+
+func TestSilnceSuppressionHandling(t *testing.T) {
```
milos-lk: minor: spelling `Silnce`

pkg/sip/media_port.go
```diff
@@ -727,6 +728,121 @@ func (p *MediaPort) setupOutput(tid traceid.ID) error {
...
+
+func (h *silenceFiller) String() string {
+	return fmt.Sprintf("SilenceFiller(%d) -> %s", h.maxGapSize, h.encodedSink.String())
+}
+
+func (h *silenceFiller) isSilenceSuppression(header *rtp.Header) (bool, int) {
+	currentTS := header.Timestamp
+	currentSeq := header.SequenceNumber
+	expectedSeq := h.lastSeq + 1
+	expectedTS := h.lastTS + uint32(h.samplesPerFrame)
```
milos-lk: though functionally correct  - maybe the following pseudo logic is a bit cleaner:
```
if !h.initialized {
  h.initialized = true
  h.lastSeq = currentSeq
  h.lastTS = currentTS
  return false, 0
}

expectedSeq := h.lastSeq + 1
expectedTS  := h.lastTS + uint32(expectedDelta)

seqDiff := currentSeq - expectedSeq
tsDiff  := currentTS - expectedTS

h.lastSeq = currentSeq
h.lastTS = currentTS
```
this avoid meaningless values for initial expected variables and is potentially less fragile if some extension is needed inside the function in future (as it overwrites last state only at the end). wdyt?

pkg/sip/media_port.go
```diff
@@ -727,6 +728,121 @@ func (p *MediaPort) setupOutput(tid traceid.ID) error {
...
+		silence := make(msdk.PCM16Sample, h.samplesPerFrame)
+		if err := h.pcmSink.WriteSample(silence); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (h *silenceFiller) HandleRTP(header *rtp.Header, payload []byte) error {
+	h.mu.Lock()
```
milos-lk: do we expect multiple goroutines calling this function on the same receiver object at the same time?

pkg/sip/media_port.go
alexlivekit: I thought about this too, but I'm concerned these might not always be treated as "read-only; do not store, buffer, or chain". I am playing it a bit safe here, for sure.

pkg/sip/media_port.go
alexlivekit: No, we're single threaded at this time. Removing.

pkg/sip/media_port.go
alexlivekit: Interesting, my aim was to modify `h.lastTS` and `h.lastSeq` as soon as possible, so that any future `if something { return false, 0 }` doesn't prevent these always staying up to date. So it turned out to be `get data` -> `update last header values` -> `everything else`.

Having said that, I do think that the `if !h.initialized {` block is cleaner for sure, updated.

lmk if that works.

pkg/sip/media_port.go
alexlivekit: A few notes on this.
- In tel, we have to specify a fixed ptimed of either 10/20/40 ms. 
  - There is no option for variable windowing, and sample rates do not change within the same payload type.
  - While it is possible to re-negotiate the same payload type to use `Opus/16000` for example, we currently also do not support SDP re-negotiation at all.
  - Our code also hard codes 20ms ptime (via `rtp.DefFramesPerSec` and others), so that's not a major concern.
- The code you are pointing to has this: `frameDurTS := uint32(uint64(len(sample)) * uint64(g.clockRate) / uint64(g.sampleRate))` is I think also interesting.
  - By and large, `g.clockRate` == `g.sampleRate`, so this in effect is `frameDurTS := uint32(len(sample))`.
  - The one exception is g722, where `g.clockRate == 8000` and `g.sampleRate == 16000`, and you take care of that!
  - This is a cool approach, but one that doesn't quite work for multiple packets being skipped in said variable windowing system.

I'm not sure I have addressed your concerns there, but I'm also not sure what the proposition is. Let me know what you thing here.

milos-lk: ah cool - as long as this stays fixed - I guess we are good.

milos-lk (approved):

PR #558, opened 2026-01-07
Signal logger

alexlivekit:
If we ever want a fancier version (used mostly for VAD, or Voice Activity Detection), where our noise floor is dynamically calculated (this helps indicate speech even in noisier environments), we can look at the implementation in [signal-logger-adaptive](https://github.com/livekit/sip/blob/signal-logger-adaptive/pkg/sip/signal_logger.go).

For now, the purpose of this feature is to output whether there's any signal, voice or noise, for sanity and debugging.

### DIFF
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -23,6 +23,7 @@ import (
 	"math"
 	"net/netip"
 	"slices"
+	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -864,7 +865,7 @@ func (c *inboundCall) handleInvite(ctx context.Context, tid traceid.ID, req *sip
 		disp.RingingTimeout = defaultRingingTimeout
 	}
 	disp.Room.JitterBuf = c.jitterBuf
-	disp.Room.LogSignalChanges = strings.EqualFold(disp.FeatureFlags[signalLoggingFeatureFlag], "true")
+	disp.Room.LogSignalChanges, _ = strconv.ParseBool(disp.FeatureFlags[signalLoggingFeatureFlag])
 	ctx, cancel := context.WithTimeout(ctx, disp.MaxCallDuration)
 	defer cancel()
 	status := CallRinging
@@ -958,7 +959,7 @@ func (c *inboundCall) runMediaConn(tid traceid.ID, offerData []byte, enc livekit
 	}
 
 	logSignalChanges := false
-	logSignalChanges = strings.EqualFold(featureFlags[signalLoggingFeatureFlag], "true")
+	logSignalChanges, _ = strconv.ParseBool(featureFlags[signalLoggingFeatureFlag])
 	mp, err := NewMediaPort(tid, c.log(), c.mon, &MediaOptions{
 		IP:                  c.s.sconf.MediaIP,
 		Ports:               conf.RTPPort,
--- a/pkg/sip/signal_logger.go
+++ b/pkg/sip/signal_logger.go
@@ -24,9 +24,6 @@ import (
 )
 
 const (
-	// int16Max is the maximum absolute value for int16 (32768); used for dBFS reference.
-	int16Max = 1 << 15
-
 	// DefaultInitialNoiseFloorDB is the default noise floor in dBFS.
 	DefaultInitialNoiseFloorDB = -50
 	// DefaultHangoverDuration is how long we stay in "signal" after level drops below exit threshold.
@@ -140,7 +137,7 @@ func (s *SignalLogger) Close() error {
 }
 
 // rmsToDBFS computes RMS of the frame then converts to dBFS: 10*log10(mean(square)/MAX^2).
-// Uses int16Max (32768) as reference. Returns a value <= 0; silence approaches -inf, so we clamp to minDBFS.
+// Uses math.MaxInt16 (32767) as reference. Returns a value <= 0; silence approaches -inf, so we clamp to minDBFS.
 func (s *SignalLogger) rmsToDBFS(sample msdk.PCM16Sample, minDBFS float64) float64 {
 	if len(sample) == 0 {
 		return minDBFS
@@ -151,7 +148,7 @@ func (s *SignalLogger) rmsToDBFS(sample msdk.PCM16Sample, minDBFS float64) float
 		sumSq += x * x
 	}
 	meanSq := float64(sumSq) / float64(len(sample))
-	refSq := float64(int16Max) * float64(int16Max)
+	refSq := float64(math.MaxInt16) * float64(math.MaxInt16)
 	if meanSq <= 0 {
 		return minDBFS
 	}
--- a/pkg/sip/outbound.go
+++ b/pkg/sip/outbound.go
@@ -20,7 +20,7 @@ import (
 	"math"
 	"net"
 	"sort"
-	"strings"
+	"strconv"
 	"sync"
 	"time"
 
@@ -91,7 +91,7 @@ type outboundCall struct {
 }
 
 func (c *Client) newCall(ctx context.Context, tid traceid.ID, conf *config.Config, log logger.Logger, id LocalTag, room RoomConfig, sipConf sipOutboundConfig, state *CallState, projectID string) (*outboundCall, error) {
-	signalLoggingEnabled := strings.EqualFold(sipConf.featureFlags[signalLoggingFeatureFlag], "true")
+	signalLoggingEnabled, _ := strconv.ParseBool(sipConf.featureFlags[signalLoggingFeatureFlag])
 	if sipConf.maxCallDuration <= 0 || sipConf.maxCallDuration > maxCallDuration {
 		sipConf.maxCallDuration = maxCallDuration
 	}


pkg/sip/inbound.go
```diff
@@ -864,6 +864,9 @@ func (c *inboundCall) handleInvite(ctx context.Context, tid traceid.ID, req *sip
 		disp.RingingTimeout = defaultRingingTimeout
 	}
 	disp.Room.JitterBuf = c.jitterBuf
+	if disp.FeatureFlags != nil {
```
dennwc: It is okay to omit this change, map lookup will still work on nil value.

pkg/sip/inbound.go
```diff
@@ -950,12 +953,17 @@ func (c *inboundCall) runMediaConn(tid traceid.ID, offerData []byte, enc livekit
 		return nil, err
 	}
 
+	logSignalChanges := false
+	if featureFlags != nil {
```
dennwc: Same here, no need for a check

pkg/sip/media_port.go
```diff
@@ -722,6 +725,12 @@ func (p *MediaPort) setupOutput(tid traceid.ID) error {
 	if p.stats != nil {
 		audioOut = newMediaWriterCount(audioOut, &p.stats.AudioOutFrames, &p.stats.AudioOutSamples)
 	}
+	if p.logSignalChanges {
+		audioOut, err = NewSignalLogger(p.log, "mixed", audioOut)
+		if err != nil {
+			return err
```
dennwc, line 779: We need to close original `audioOut` here.

pkg/sip/room.go
```diff
@@ -321,7 +322,12 @@ func (r *Room) Connect(conf *config.Config, rconf RoomConfig) error {
 					defer log.Infow("track closed")
 					defer mTrack.Close()
 
-					codec, err := opus.Decode(mTrack, channels, log)
+					var out msdk.PCM16Writer = mTrack
+					if rconf.LogSignalChanges {
+						out, _ = NewSignalLogger(log, track.ID(), out)
```
dennwc: Not even logging an error? This could silenty write nil to `out`.

pkg/sip/signal_logger.go
```diff
@@ -0,0 +1,161 @@
...
+const (
+	DefaultSignalMultiplier = float64(2)      // Singnal needs to be at least 2 times the noise floor to be detected.
+	DefaultNoiseFloor       = float64(25)     // Use a low noise floor; The purpose is to detect any kind of signal, not just voice.
+	DefaultHangoverDuration = 1 * time.Second // Duration of silence that needs to elapse before we no longer consider it a signal.
+)
+
+// Keeps an internal state of whether we're currently transmitting signal (voice or noise), or silence.
+// This implements msdk.PCM16Writer to inspect decoded packet content.
+// Used to log changes betweem those states.
+type SignalLogger struct {
```
dennwc, line 42: We should consider moving it to `media-sdk`, since it's not SIP-specific.

pkg/sip/signal_logger.go
```diff
@@ -0,0 +1,161 @@
...
+
+	now := time.Now()
+	isSignal := currentEnergy > (s.noiseFloor * s.signalMultiplier)
+	if isSignal {
+		s.lastSignalTime = now
+	}
+
+	s.framesProcessed++
+	if s.framesProcessed > 10 { // Don't log any changes at first
+		if isSignal && isSignal != s.lastIsSignal { // silence -> signal: Immediate transition
```
dennwc: This could potentially hit each frame, wouldn't it be too spammy?

pkg/sip/signal_logger.go
```diff
@@ -0,0 +1,161 @@
...
+			totalAbs += int64(-v)
+		} else {
+			totalAbs += int64(v)
+		}
+	}
+	return float64(totalAbs) / float64(len(sample))
+}
+
+func (s *SignalLogger) WriteSample(sample msdk.PCM16Sample) error {
+	currentEnergy := s.MeanAbsoluteDeviation(sample)
```
dennwc: Probably doesn't matter here, but energy is usually computed as a square of amplitude. Maybe rename to `ampMean`?

pkg/sip/signal_logger.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	var totalAbs int64
+	for _, v := range sample {
+		if v < 0 {
+			totalAbs += int64(-v)
+		} else {
+			totalAbs += int64(v)
+		}
+	}
+	return float64(totalAbs) / float64(len(sample))
```
dennwc: We could also consider calculating a median, it will be less affected by sudden spikes.

pkg/sip/signal_logger.go
```diff
@@ -0,0 +1,161 @@
...
+}
+func NewSignalLogger(log logger.Logger, name string, next msdk.PCM16Writer, options ...SignalLoggerOption) (msdk.PCM16Writer, error) {
+	s := &SignalLogger{
+		log:              log,
+		next:             next,
+		name:             name,
+		hangoverDuration: DefaultHangoverDuration,
+		signalMultiplier: DefaultSignalMultiplier,
+		noiseFloor:       DefaultNoiseFloor,
+		lastSignalTime:   time.Time{},
```
dennwc: No need to initialize explicitly.

dennwc (approved):

pkg/sip/media_port.go
hechen-eng: why do we need to close original `audioOut`? It seems `MediaPort.Close()` does it later 🤔

pkg/sip/media_port.go
```diff
@@ -755,6 +764,14 @@ func (p *MediaPort) setupInput() {
 	if p.stats != nil {
 		audioWriter = newMediaWriterCount(audioWriter, &p.stats.AudioInFrames, &p.stats.AudioInSamples)
 	}
+	if p.logSignalChanges {
+		signalLogger, err := NewSignalLogger(p.log, "input", audioWriter)
```
hechen-eng, line 816: curious.. why don't we return error for the `setupInput` path?

pkg/sip/signal_logger.go
alexlivekit: Would still want to use average to not filter out signal in high or low bands, which should count as signal still.

pkg/sip/signal_logger.go
alexlivekit: That's what `hangoverDuration` is about. It can flip to signal immediately, but to flip out of `isSignal` you'd need to wait at least a second of silence. Would that still be too spammy?

pkg/sip/media_port.go
alexlivekit: Good catch.

> It seems MediaPort.Close() does it later

It doesn't since we don't actually link the port to `audioOut` if we error out here.

pkg/sip/signal_logger_test.go
```diff
@@ -0,0 +1,274 @@
...
+		t.Run("signal_to_silence", func(t *testing.T) {
+			synctest.Test(t, func(t *testing.T) {
+				sl := testTransition(t, signalFrames, silenceFrames, 10, 50, true)
+				require.Equal(t, uint64(1), sl.stateChanges)
+			})
+		})
+	})
+
+	t.Run("silence_to_silence_transitions", func(t *testing.T) {
+		sl := testTransition(t, silenceFrames, silenceFrames, 10, 0, false) // Not too long, it will eventually bring noise floor low enough
```
hechen-eng: why 0 second count?

pkg/sip/media_port.go
alexlivekit: didn't want to change too much, and the function currently doesn't support erroring out. Using the original `audioWriter` in that case doesn't actually degrade anything.

pkg/sip/signal_logger_test.go
alexlivekit: because both are silence, and 10 packets is enough. Doesn't really matter either way. The one point here is testing that the tenth packet doesn't trigger any transition.

pkg/sip/media_port.go
hechen-eng: fixed!

pkg/sip/inbound.go
```diff
@@ -864,9 +864,7 @@ func (c *inboundCall) handleInvite(ctx context.Context, tid traceid.ID, req *sip
 		disp.RingingTimeout = defaultRingingTimeout
 	}
 	disp.Room.JitterBuf = c.jitterBuf
-	if disp.FeatureFlags != nil {
-		disp.Room.LogSignalChanges = disp.FeatureFlags[signalLoggingFeatureFlag] == "true"
-	}
+	disp.Room.LogSignalChanges = strings.EqualFold(disp.FeatureFlags[signalLoggingFeatureFlag], "true")
```
alexlivekit: Better yet, use `strconv.ParseBool`

pkg/sip/signal_logger.go
```diff
@@ -0,0 +1,219 @@
...
+	"math"
+	"time"
+
+	msdk "github.com/livekit/media-sdk"
+	"github.com/livekit/protocol/logger"
+)
+
+const (
+	// int16Max is the maximum absolute value for int16 (32768); used for dBFS reference.
+	int16Max = 1 << 15
```
alexlivekit: use `math.MaxUint64` and the like instead.

alexlivekit (commented):
Overall, other than the two nits, this is looking good imo.

Since this is gated by a feature flag, imo this is okay to land, but please test the new changes in staging before enabling for anyone else.

PR #559, opened 2026-01-08
Minor touchups

alexlivekit:
- silence filler into its own file.
- added room print with explicit call / room / participant tie.


pkg/sip/room.go
```diff
@@ -403,6 +403,8 @@ func (r *Room) Connect(conf *config.Config, rconf RoomConfig) error {
 	r.room = room
 	r.p.ID = r.room.LocalParticipant.SID()
 	r.p.Identity = r.room.LocalParticipant.Identity()
+	r.log = r.log.WithValues("room", r.room.Name(), "roomID", r.room.SID(), "participant", r.p.Identity, "pID", r.p.ID)
+	r.log.Infow("joined room")
```
nishadmusthafa: NIT: Should we add something to indicate that it is a sip participant that joined? I think it's not useful now. But later on when we start melding logs across agent, sip and other client types, it might help.

nishadmusthafa (approved):

PR #560, opened 2026-01-08
Adding debug logging for tcp dials

nishadmusthafa:


pkg/sip/outbound.go
```diff
@@ -971,6 +960,27 @@ func (c *sipOutbound) attemptInvite(ctx context.Context, callID sip.CallIDHeader
...
+				dest := req.Destination()
+				if dest != "" {
+					if conn, err := tpl.GetConnection("tcp", dest); err == nil && conn != nil {
+						if localAddr := conn.LocalAddr(); localAddr != nil {
+							if tcpAddr, ok := localAddr.(*net.TCPAddr); ok {
+								c.log.Debugw("TCP connection using port on cloud-sip side", "port", tcpAddr.Port)
+							}
+						}
+					}
+				}
```
```suggestion
				if dest := req.Destination(); dest != "" {
					if conn, err := tpl.GetConnection("tcp", dest); err == nil && conn != nil {
						if tcpAddr, ok := conn.LocalAddr().(*net.TCPAddr); ok && tcpAddr != nil {
								c.log.Debugw("TCP connection using port on cloud-sip side", "port", tcpAddr.Port)
						}
					}
				}
```
It's a bit more compact this way. It's a bit weird to have all these nested ifs, so might be better in a separate function (where it can `return` early).

dennwc (approved):

PR #561, opened 2026-01-09
Brushing up the stats collection

alexlivekit:


pkg/sip/media.go
```diff
@@ -164,6 +190,9 @@ func (r *rtpStatsHandler) String() string {
 func (r *rtpStatsHandler) HandleRTP(h *rtp.Header, payload []byte) error {
 	if r.mon != nil {
 		typ := r.typ
+		if typ == "" && h.PayloadType < 96 {
+			typ, _ = staticPayloadTypes[h.PayloadType]
```
dennwc, line 194: Pion's SDP package has a helper for this, I believe

alexlivekit: I know the `int` [version exists](https://github.com/pion/rtp/blob/1edc72d15cbeb6ffe0fe40f863797273c19083f2/payload_types.go#L10), but wanted this to almost match (with the `auto_` prefix or something) [what we use today](https://ophoenix1b.grafana.production.livekit-internal.net/goto/UDF-wr4DR?orgId=1).

PR #563, opened 2026-01-14
Test ringing duration in E2E tests.

dennwc:


test/lktest/sip.go
```diff
@@ -154,15 +156,30 @@ func TestSIPOutbound(t TB, ctx context.Context, lkOut, lkIn *LiveKit, params SIP
...
+		outRingStart time.Time
 	)
 	defer func() {
+		if params.RingFor > 0 {
+			ringedFor := time.Duration(ringingOut.Load())
+			const (
+				dtmin = 2 * time.Second
+				dtmax = 3 * time.Second
+			)
+			if ringedFor < params.RingFor-dtmin || ringedFor > params.RingFor+dtmax {
```
alexlivekit, line 177: This fudging feels like a lot. Did we see problems with it, or are we planning to trim it down afterwards?

The reason I'm asking is because if this is the actual variability we have in our system we might want to go ahead and address that as a standalone bug.

alexlivekit (approved):
overall lgtm

test/lktest/sip.go
dennwc: We can narrow it down later. Delay is mostly on E2E test side itself, not in our system. Ideally putting a metric here would help, but we need to set a push gateway for it on infra side first.

dennwc: We can narrow it down later. Delay is mostly on E2E test side itself, not in our system. Ideally putting a metric here would help, but we need to set a push gateway for it on infra side first.

PR #569, opened 2026-01-21
adding termination error logging

alexlivekit:


pkg/sip/inbound.go
```diff
@@ -1129,6 +1130,13 @@ func (c *inboundCall) printStats(log logger.Logger) {
 
 // close should only be called from handleInvite.
 func (c *inboundCall) close(ctx context.Context, error bool, status CallStatus, reason string) {
+	go func() {
+		time.Sleep(5 * time.Minute)
+		if !c.terminated.Load() {
+			c.log().Errorw("call failed to terminate after 5 minutes", nil)
```
nishadmusthafa: If the point is to alert, why not add a prom metric ?

alexlivekit: quick and dirty and all that. Can do though.

alexlivekit: Updated, although we do want the log still - we want the IDs.
That's why I was thinking to create a log-based alert, that can also work, and not have to double-count

pkg/sip/inbound.go
```diff
@@ -1129,6 +1130,14 @@ func (c *inboundCall) printStats(log logger.Logger) {
...
 func (c *inboundCall) close(ctx context.Context, error bool, status CallStatus, reason string) {
+	go func() {
+		time.Sleep(5 * time.Minute)
+		if !c.terminated.Load() {
+			c.mon.CallTerminationFailure()
+			c.log().Errorw("call failed to terminate after 5 minutes", nil) // To be able to get call IDs
+		}
+	}()
+
 	ctx = context.WithoutCancel(ctx)
```
dennwc, line 1144:
```suggestion
	ctx = context.WithoutCancel(ctx) // do not timeout
	ctx, cancel := context.WithCancel(ctx)
	defer cancel()
	go func() {
		select {
		case <-ctx.Done():
			return
		case <-time.After(5 * time.Minute):
			if !c.terminated.Load() {
				c.mon.CallTerminationFailure()
				c.log().Errorw("call failed to terminate after 5 minutes", nil) // To be able to get call IDs
			}
		}
	}()
```

We shouldn't block a goroutine unconditionally, it should also `select` on the context of the parent call.

Also, this should be after the CAS on `c.done`, so that we don't run an extra goroutine if `Close` is called twice.

pkg/sip/inbound.go
```diff
@@ -1184,6 +1193,7 @@ func (c *inboundCall) close(ctx context.Context, error bool, status CallStatus,
 	}
 
 	c.cancel()
+	c.terminated.Store(true)
```
dennwc: Maybe `defer` it?

pkg/sip/outbound.go
```diff
@@ -509,8 +512,16 @@ func sipResponse(ctx context.Context, tx sip.ClientTransaction, stop <-chan stru
 }
 
 func (c *outboundCall) stopSIP(ctx context.Context, reason string) {
+	go func() {
+		time.Sleep(5 * time.Minute)
```
dennwc: Same here, we shouldn't block unconditionally.

pkg/stats/monitor.go
```diff
@@ -395,6 +404,10 @@ func (c *CallMonitor) CallTerminate(reason string) {
 	c.m.callsTerminated.With(c.labels(prometheus.Labels{"reason": reason})).Inc()
 }
 
+func (c *CallMonitor) CallTerminationFailure() {
+	c.m.callsTerminationFailures.With(c.labels(prometheus.Labels{})).Inc()
```
dennwc:
```suggestion
	c.m.callsTerminationFailures.With(c.labels(nil)).Inc()
```

pkg/stats/monitor.go
```diff
@@ -170,6 +171,14 @@ func (m *Monitor) Start(conf *config.Config) error {
 		ConstLabels: prometheus.Labels{"node_id": conf.NodeID},
 	}, []string{"dir", "to", "reason"}))
 
+	m.callsTerminationFailures = mustRegister(m, prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace:   "livekit",
+		Subsystem:   "sip",
+		Name:        "calls_termination_failures",
+		Help:        "Number of calls that failed to terminate after 5 minutes",
+		ConstLabels: prometheus.Labels{"node_id": conf.NodeID},
+	}, []string{"dir"}))
```
dennwc, line 180: Calling `c.labels` will add `to` as well, which is not defined here. This will likely panic.

PR #570, opened 2026-01-22
discarding buffered packets at the start of the stream

alexlivekit:


pkg/sip/media_port.go
```diff
@@ -632,7 +632,6 @@ func (p *MediaPort) rtpLoop(tid traceid.ID, sess rtp.Session) {
 		p.stats.Streams.Add(1)
 		p.mediaReceived.Break()
 		log := p.log.WithValues("ssrc", ssrc)
-		log.Infow("accepting RTP stream")
```
alexlivekit, line 635: Moved to include discarded packets number, and first RTP seq

pkg/sip/media_port.go
```diff
@@ -655,6 +657,19 @@ func (p *MediaPort) rtpReadLoop(tid traceid.ID, log logger.Logger, r rtp.ReadStr
 			log.Errorw("read RTP failed", err)
 			return
 		}
+		if catchup {
+			// When a remote writer has been sending data but the local stream didn't start the RTP loop,
+			// such as in outbound calls that take a sec to answer, we have a ton of junk in the stream.
+			// We need to discard all of it.
+			now := time.Now()
+			if lastPacketTime.IsZero() || now.Sub(lastPacketTime) < time.Millisecond {
```
alexlivekit, line 665: Note, for inbound calls, we will discard the first packet as well, even if there isn't anything buffered in there.
We can, of course introduce code to correct for that, but the added complexity was deemed not justified due to non-existing impact.

pkg/sip/media_port.go
```diff
@@ -655,6 +657,19 @@ func (p *MediaPort) rtpReadLoop(tid traceid.ID, log logger.Logger, r rtp.ReadStr
 			log.Errorw("read RTP failed", err)
 			return
 		}
+		if catchup {
+			// When a remote writer has been sending data but the local stream didn't start the RTP loop,
+			// such as in outbound calls that take a sec to answer, we have a ton of junk in the stream.
```
dennwc, line 662: I'm not sure I'm following here.

There's presumably some garbage sent before the call is established. But the code in the PR doesn't even check the RTP headers or the payload. Why are we dropping it unconditionally?

Assuming there's some audio there (silence? ringing?) or even just audio junk, it won't propagate through the media pipeline until it's actually enabled (we have multiple `SwitchWriter`s in the pipeline that can cut the stream).

So why is this necessary?

alexlivekit: The timeline is this:
- Start outbound call.
  - This causes us to start configuring the media port, and bind local sockets.
  - However, we do not yet start rtpLoop.
  - We want to be doing as early as possible to make sure media i ready to go. If we ever add DTLS support for calls, it would be cruicial for this to happen early as possible to give the encryption time to set up.
- Wait for user to answer
  - During this time, incoming packets that are already being sent by the callee to livekit are just sitting in the OS UDP buffer.
  - Some packets will be dropped from the buffer. Annoyingly, it will be newer packets that would be dropped by the OS if the UDP buffer is full.
-  Receive SIP-200
  - Only now do we start rtpLoop
  - This starts reading ALL the packets in the OS buffer, along with any gaps that might have been in there due to packet drops.

PR #575, opened 2026-01-26
feat(tls): allow configuring TLS versions and cipher suites

NikToloknov:
Expose configuration options for TLS MinVersion, MaxVersion, and CipherSuites backed by crypto/tls. This allows LiveKit to interoperate with providers that require non-default or legacy TLS configurations.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -55,8 +55,8 @@ type TLSConfig struct {
 	Certs      []TLSCert `yaml:"certs"`
 	KeyLog     string    `yaml:"key_log"`
 
-	MinVersion string `yaml:"min_version"` // min TLS version, accepts: "1.0", "1.1", "1.2", "1.3"
-	MaxVersion string `yaml:"max_version"` // max TLS version, accepts: "1.0", "1.1", "1.2", "1.3"
+	MinVersion string `yaml:"min_version"` // min TLS version, accepts: "tls1.0", "tls1.1", "tls1.2", "tls1.3"
+	MaxVersion string `yaml:"max_version"` // max TLS version, accepts: "tls1.0", "tls1.1", "tls1.2", "tls1.3"
 
 	// CipherSuites is an optional list of cipher suite names.
 	// If not provided, Go's secure defaults are used.
--- a/pkg/sip/tls.go
+++ b/pkg/sip/tls.go
@@ -60,9 +60,9 @@ func ConfigureTLS(c *tls.Config) {
 	}
 }
 
-// ParseCipherSuites parses cipher suite names to uint16 IDs.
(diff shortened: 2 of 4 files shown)


pkg/config/config.go
```diff
@@ -54,6 +54,14 @@ type TLSConfig struct {
 	ListenPort int       `yaml:"port_listen"` // SIP signaling port to listen on
 	Certs      []TLSCert `yaml:"certs"`
 	KeyLog     string    `yaml:"key_log"`
+
+	MinVersion string `yaml:"min_version"` // min TLS version, accepts: "1.0", "1.1", "1.2", "1.3"
```
dennwc: Values like `1.0` are dangerous in YAML, since they will be interpreted as a number by default. The user must then know to set it to `"1.0"` (a string).

So I'd propose we settle on accepting just `tls1.0` and/or `TLS1.0`.

pkg/sip/tls.go
```diff
@@ -49,3 +59,47 @@ func ConfigureTLS(c *tls.Config) {
 		return nil
 	}
 }
+
+// ParseCipherSuites parses cipher suite names to uint16 IDs.
+// Logs a warning for each insecure cipher suite configured.
+func ParseCipherSuites(suites []string, log logger.Logger) ([]uint16, error) {
```
dennwc:
```suggestion
func ParseCipherSuites(log logger.Logger, suites []string) ([]uint16, error) {
```

dennwc:
```suggestion
func parseCipherSuites(suites []string, log logger.Logger) ([]uint16, error) {
```
Can be unexported, since these are only used it this package.

pkg/sip/tls.go
```diff
@@ -49,3 +59,47 @@ func ConfigureTLS(c *tls.Config) {
...
+			return nil, errors.New("unknown cipher suite: " + suite)
+		}
+	}
+
+	return parsedCipherSuites, nil
+}
+
+// ParseTLSVersion parses a TLS version string to its uint16 constant.
+// Accepts formats: "1.0", "1.1", "1.2", "1.3" or "TLS 1.0", "TLS 1.1", "TLS 1.2", "TLS 1.3".
+func ParseTLSVersion(version string) (uint16, error) {
```
dennwc:
```suggestion
func parseTLSVersion(version string) (uint16, error) {
```
Same here - can be unexported.

pkg/config/config.go
NikToloknov: Good point, thanks. I’ve updated the code to accept only `TLS1.0` and `tls1.0` to avoid YAML numeric parsing issues.

NikToloknov:
@dennwc Updated, thanks! Would appreciate another review.

dennwc (approved):

PR #581, opened 2026-02-05
Modify psrpc.Canceled to psrpc.DeadlineExceeded for requests that tim…

alexlivekit:
…e out


pkg/sip/outbound.go
```diff
@@ -489,10 +489,10 @@ func sipResponse(ctx context.Context, tx sip.ClientTransaction, stop <-chan stru
 		select {
 		case <-ctx.Done():
 			_ = tx.Cancel()
-			return nil, psrpc.NewErrorf(psrpc.Canceled, "canceled")
+			return nil, psrpc.NewErrorf(psrpc.DeadlineExceeded, "timed out")
```
```suggestion
			return nil, psrpc.NewErrorf(psrpc.DeadlineExceeded, "sip request timed out")
```

PR #582, opened 2026-02-06
Fix re-INVITE dialog matching using RFC 3261 compliant To tag lookup

briankwest:
Fixes a critical bug where mid-dialog re-INVITEs from the remote party were incorrectly processed as new inbound calls instead of being recognized as part of an existing dialog.

## Problem
When a SIP provider (e.g., SignalWire) sends a re-INVITE during an active call, LiveKit was treating it as a new call, causing:
- Authentication challenges (401/407)
- "No trunk found" errors (404)
- Call failures and disconnections

This primarily affected outbound calls where From/To headers are "swapped" from LiveKit's perspective (remote is UAC for the re-INVITE).

## Root Cause
The existing dialog matching happened too late (after processInvite) and only checked byCallID, which is unreliable. The code missed the definitive proof: the To tag in the request matches our local tag.

Per RFC 3261 Section 12.2, a dialog is identified by:
- Call-ID
- Local tag (in To header for UAS)
- Remote tag (in From header for UAS)

## Solution
Add early dialog matching in onInvite() before processing as new call:
1. Extract To tag from incoming INVITE
2. Check if tag exists in byLocalTag map
3. If match found, delegate to existing call's handleReinvite()
4. Otherwise, proceed with normal new call processing

New handleReinvite() method on inboundCall:
- Validates CSeq (detect retransmissions vs new re-INVITEs)
- Responds with current SDP for session refresh
- Handles edge cases (old CSeq, missing headers)

## Impact
✅ Fixes outbound call re-INVITEs (broken → working) ✅ Improves inbound call re-INVITEs (performance + reliability) ✅ RFC 3261 compliant dialog identification
✅ Reduced CPU usage (early exit avoids processInvite) ✅ Better logging for debugging

## Testing
Tested with SignalWire re-INVITEs on both inbound and outbound calls. Verified proper 200 OK responses and continued call operation.

### DIFF
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -182,9 +182,12 @@ func (s *Server) handleInviteAuth(tid traceid.ID, log logger.Logger, req *sip.Re
 
 	h := req.GetHeader("Proxy-Authorization")
 	if h == nil {
+		// Generate unique nonce using timestamp + Call-ID hash to avoid collisions
+		// when multiple calls arrive in the same microsecond
+		nonce := fmt.Sprintf("%d-%s", time.Now().UnixMicro(), sipCallID)
 		inviteState.challenge = digest.Challenge{
 			Realm:     UserAgent,
-			Nonce:     fmt.Sprintf("%d", time.Now().UnixMicro()),
+			Nonce:     nonce,
 			Algorithm: "MD5",
 		}
 
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -78,6 +78,13 @@ func hashPassword(password string) string {
 	return hex.EncodeToString(hash[:8]) // Use first 8 bytes for shorter hash
 }
 
+// generateNonce creates a unique nonce for digest authentication challenges.
+// Includes both timestamp and Call-ID to ensure uniqueness even when multiple
+// challenges are generated in the same microsecond.
+func generateNonce(sipCallID string) string {
+	return fmt.Sprintf("%d-%s", time.Now().UnixMicro(), sipCallID)
+}
+
 type inboundCallInfo struct {
 	sync.Mutex
 	cseq        uint32
@@ -182,12 +189,9 @@ func (s *Server) handleInviteAuth(tid traceid.ID, log logger.Logger, req *sip.Re
 
 	h := req.GetHeader("Proxy-Authorization")
 	if h == nil {
-		// Generate unique nonce using timestamp + Call-ID hash to avoid collisions
-		// when multiple calls arrive in the same microsecond
-		nonce := fmt.Sprintf("%d-%s", time.Now().UnixMicro(), sipCallID)
 		inviteState.challenge = digest.Challenge{
 			Realm:     UserAgent,
-			Nonce:     nonce,
+			Nonce:     generateNonce(sipCallID),
 			Algorithm: "MD5",
 		}
 
@@ -389,15 +393,8 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 		return psrpc.NewError(psrpc.InvalidArgument, errors.Wrap(err, "invite validation failed"))
 	}
 
-	s.cmu.RLock()
-	existing := s.byCallID[cc.SIPCallID()]
-	s.cmu.RUnlock()
-	if existing != nil && existing.cc.InviteCSeq() < cc.InviteCSeq() {
-		log.Infow("accepting reinvite", "sipCallID", existing.cc.ID(), "content-type", req.ContentType(), "content-length", req.ContentLength())
-		existing.log().Infow("reinvite", "content-type", req.ContentType(), "content-length", req.ContentLength(), "cseq", cc.InviteCSeq())
-		cc.AcceptAsKeepAlive(existing.cc.OwnSDP())
-		return nil
-	}
+	// Note: Mid-dialog re-INVITEs are now handled earlier in onInvite() via To tag matching (RFC 3261)
+	// This avoids unnecessary processing and uses proper dialog identification
 
 	from, to := cc.From(), cc.To()
 
@@ -1321,6 +1318,8 @@ func (c *inboundCall) handleReinvite(req *sip.Request, tx sip.ServerTransaction)
 			r := sip.NewResponseFromRequest(req, sip.StatusOK, "OK", sdp)
 			r.AppendHeader(&contentTypeHeaderSDP)
 			_ = tx.Respond(r)
+		} else {
+			c.log().Warnw("retransmission: no SDP available, not responding", nil)
 		}
 		return
 	}
@@ -1344,14 +1343,14 @@ func (c *inboundCall) handleReinvite(req *sip.Request, tx sip.ServerTransaction)
 	// TODO: Support SDP renegotiation if needed in the future
 	//       This would require parsing the offer, updating media session, etc.
 
-	r := sip.NewResponseFromRequest(req, sip.StatusOK, "OK", currentSDP)
-	r.AppendHeader(&contentTypeHeaderSDP)
+	// Create a temporary sipInbound for the re-INVITE to properly handle the response
+	// This ensures all headers (Allow, Contact, etc.) are added correctly
+	tr := callTransportFromReq(req)
+	legTr := legTransportFromReq(req)
+	reinviteCC := c.s.newInbound(c.log(), c.cc.tag, c.s.ContactURI(legTr), req, tx, nil)
 
-	err := tx.Respond(r)
-	if err != nil {
-		c.log().Errorw("failed to respond to re-INVITE", err)
-		return
-	}
+	// Use the existing helper to respond with proper headers
+	reinviteCC.AcceptAsKeepAlive(currentSDP)
 
 	c.log().Infow("re-INVITE accepted with current SDP")
 }
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -1348,8 +1348,12 @@ func (c *inboundCall) handleReinvite(req *sip.Request, tx sip.ServerTransaction)
 	r := sip.NewResponseFromRequest(req, sip.StatusOK, "OK", currentSDP)
 	r.AppendHeader(&contentTypeHeaderSDP)
 	r.AppendHeader(sip.NewHeader("Allow", "INVITE, ACK, CANCEL, BYE, NOTIFY, REFER, MESSAGE, OPTIONS, INFO, SUBSCRIBE"))
-	r.AppendHeader(c.cc.contact)
-	c.cc.addExtraHeaders(r)
+	if c.cc.contact != nil {
+		r.AppendHeader(c.cc.contact)
+	}
+	if c.s != nil && c.s.conf != nil {
+		c.cc.addExtraHeaders(r)
+	}
 
 	err := tx.Respond(r)
 	if err != nil {
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -1310,17 +1310,10 @@ func (c *inboundCall) handleReinvite(req *sip.Request, tx sip.ServerTransaction)
 	if cseq.SeqNo == inviteCSeq {
 		// Same CSeq - this is a retransmission of the original INVITE
 		// Resend the original 200 OK if we already accepted
-		c.log().Infow("INVITE retransmission detected",
-			"cseq", cseq.SeqNo)
+		c.log().Infow("INVITE retransmission detected", "cseq", cseq.SeqNo)
 
 		// Re-send 200 OK with our current SDP
-		if sdp := c.cc.OwnSDP(); sdp != nil {
-			r := sip.NewResponseFromRequest(req, sip.StatusOK, "OK", sdp)
-			r.AppendHeader(&contentTypeHeaderSDP)
-			_ = tx.Respond(r)
-		} else {
-			c.log().Warnw("retransmission: no SDP available, not responding", nil)
-		}
+		c.cc.ResendOK(tx, req)
 		return
 	}
 
@@ -1331,37 +1324,10 @@ func (c *inboundCall) handleReinvite(req *sip.Request, tx sip.ServerTransaction)
 		"contentType", req.ContentType(),
 		"contentLength", req.ContentLength())
 
-	// Get current SDP to send in response
-	currentSDP := c.cc.OwnSDP()
-	if currentSDP == nil {
-		c.log().Errorw("no SDP available for re-INVITE response", nil)
-		_ = tx.Respond(sip.NewResponseFromRequest(req, sip.StatusInternalServerError, "No SDP available", nil))
-		return
-	}
-
 	// For now, we respond with the same SDP (session refresh, no renegotiation)
 	// TODO: Support SDP renegotiation if needed in the future
 	//       This would require parsing the offer, updating media session, etc.
-
-	// Respond with 200 OK including all proper SIP headers
-	// Following the same pattern as AcceptAsKeepAlive/respondWithData
-	r := sip.NewResponseFromRequest(req, sip.StatusOK, "OK", currentSDP)
-	r.AppendHeader(&contentTypeHeaderSDP)
-	r.AppendHeader(sip.NewHeader("Allow", "INVITE, ACK, CANCEL, BYE, NOTIFY, REFER, MESSAGE, OPTIONS, INFO, SUBSCRIBE"))
-	if c.cc.contact != nil {
-		r.AppendHeader(c.cc.contact)
-	}
-	if c.s != nil && c.s.conf != nil {
-		c.cc.addExtraHeaders(r)
-	}
-
-	err := tx.Respond(r)
-	if err != nil {
-		c.log().Errorw("failed to respond to re-INVITE", err)
-		return
-	}
-
-	c.log().Infow("re-INVITE accepted with current SDP")
+	c.cc.AcceptAsKeepAlive(tx, req)
 }
 
 func (c *inboundCall) Close() error {
@@ -1631,8 +1597,15 @@ func (c *sipInbound) respondWithData(status sip.StatusCode, reason string, conte
 	if c.inviteTx == nil {
 		return
 	}
+	c.respondWithDataTo(c.inviteTx, c.invite, status, reason, contentType, body)
+}
 
-	r := sip.NewResponseFromRequest(c.invite, status, reason, body)
+func (c *sipInbound) respondWithDataTo(
+	tx sip.ServerTransaction, invite *sip.Request,
+	status sip.StatusCode, reason string,
+	contentType string, body []byte,
+) {
+	r := sip.NewResponseFromRequest(invite, status, reason, body)
 	if typ := sip.ContentTypeHeader(contentType); typ != "" {
 		r.AppendHeader(&typ)
 	}
@@ -1642,7 +1615,7 @@ func (c *sipInbound) respondWithData(status sip.StatusCode, reason string, conte
 		r.AppendHeader(c.contact)
 	}
 	c.addExtraHeaders(r)
-	_ = c.inviteTx.Respond(r)
+	_ = tx.Respond(r)
 }
 
 func (c *sipInbound) RespondAndDrop(status sip.StatusCode, reason string) {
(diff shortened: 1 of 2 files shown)


briankwest:
# Re-INVITE Flow: Before vs After Fix

## 🔴 BEFORE FIX (Broken)

```
┌─────────────────────────────────────────────────────────────────┐
│ SignalWire sends re-INVITE on established outbound call         │
│                                                                 │
│  From: <sip:+16048158957@sw.com>;tag=gK69j24Mj789F              │
│  To: <sip:+14373654058@lk.com>;tag=SCL_4upGMjQw3oCb  ← OUR TAG  │
│  Call-ID: abc123                                                │
│  CSeq: 2 INVITE                                                 │
└─────────────────────────────────────────────────────────────────┘
                            │
                            ▼
                    ┌──────────────┐
                    │  onInvite()  │
                    └──────┬───────┘
                           │
                           ▼
                ┌──────────────────────┐
                │  processInvite()     │  ← Treats as NEW call!
                │  (line 282)          │
                └──────────┬───────────┘
                           │
                           ▼
        ┌──────────────────────────────────────┐
        │ Creates new sipInbound context       │
        │ Generates NEW call ID: xyz789        │
        │ Assigns NEW local tag                │
        └──────────┬───────────────────────────┘
                   │
                   ▼
        ┌──────────────────────────────────────┐
        │ ValidateInvite()                     │
        └──────────┬───────────────────────────┘
                   │
                   ▼
        ┌──────────────────────────────────────┐
        │ ? Check byCallID (line 347)          │  ← Too late!
        │                                      │
        │ existing := s.byCallID["abc123"]     │
        │ if existing != nil {                 │
        │   // This MIGHT work for inbound     │
        │   // but FAILS for outbound!         │
        │ }                                    │
        └──────────┬───────────────────────────┘
                   │
                   ├─── IF MATCHED (lucky) ────┐
                   │                           │
                   │                           ▼
                   │                   ┌─────────────────┐
                   │                   │ Accept as       │
                   │                   │ keep-alive      │
                   │                   └─────────────────┘
                   │
                   └─── IF NOT MATCHED (bug!) ──┐
                                                │
                                                ▼
                        ┌──────────────────────────────────┐
                        │ X GetAuthCredentials()           │
                        │ X Checks trunks/dispatch rules   │
                        │ X Sends 401/407 challenge        │
                        │ X Or 404 No Trunk Found          │
                        └──────────────────────────────────┘
                                     │
                                     ▼
                        ┌──────────────────────────────────┐
                        │ * RE-INVITE FAILS                │
                        │ * Call may drop                  │
                        │ * Provider confused              │
                        └──────────────────────────────────┘
```

## * AFTER FIX (Correct)

```
┌─────────────────────────────────────────────────────────────────┐
│ SignalWire sends re-INVITE on established outbound call         │
│                                                                 │
│  From: <sip:+16048158957@sw.com>;tag=gK69j24Mj789F              │
│  To: <sip:+14373654058@lk.com>;tag=SCL_4upGMjQw3oCb  ← OUR TAG  │
│  Call-ID: abc123                                                │
│  CSeq: 2 INVITE                                                 │
└─────────────────────────────────────────────────────────────────┘
                            │
                            ▼
                    ┌──────────────────┐
                    │  onInvite()      │  ← NEW: Early dialog check
                    └──────┬───────────┘
                           │
                           ▼
        ┌──────────────────────────────────────────────┐
        │ * Extract To tag: "SCL_4upGMjQw3oCb"         │
        │ s.cmu.RLock()                                │
        │ existing := s.byLocalTag[LocalTag(toTag)]    │
        │ s.cmu.RUnlock()                              │
        └──────────┬───────────────────────────────────┘
                   │
                   ├─── To tag NOT in byLocalTag ───┐
                   │    (new call)                  │
                   │                                ▼
                   │                    ┌─────────────────────┐
                   │                    │ processInvite()     │
                   │                    │ (normal flow)       │
                   │                    └─────────────────────┘
                   │
                   └─── To tag FOUND in byLocalTag ──┐
                        (existing dialog!)           │
                                                     ▼
                        ┌──────────────────────────────────────┐
                        │ * existing.handleReinvite(req, tx)   │
                        │                                      │
                        │ LOG: "received mid-dialog re-INVITE" │
                        │      localTag=SCL_4upGMjQw3oCb       │
                        └──────────┬───────────────────────────┘
                                   │
                                   ▼
                        ┌──────────────────────────────────────┐
                        │ Validate CSeq                        │
                        │                                      │
                        │ cseq := req.CSeq()                   │
                        │ inviteCSeq := c.cc.InviteCSeq()      │
                        └──────────┬───────────────────────────┘
                                   │
                ┌──────────────────┼──────────────────┐
                │                  │                  │
                ▼                  ▼                  ▼
        ┌────────────┐   ┌─────────────────┐   ┌──────────────┐
        │ CSeq < old │   │ CSeq == old     │   │ CSeq > old   │
        │ (ignore)   │   │ (retransmit)    │   │ (re-INVITE)  │
        └────────────┘   └────────┬────────┘   └──────┬───────┘
                                  │                   │
                                  ▼                   ▼
                        ┌──────────────────┐   ┌──────────────────┐
                        │ Resend 200 OK    │   │ Accept with      │
                        │ with same SDP    │   │ current SDP      │
                        └──────────────────┘   └──────┬───────────┘
                                                      │
                                                      ▼
                        ┌──────────────────────────────────────┐
                        │ * 200 OK sent                        │
                        │ * Call continues normally            │
                        │ * No new call created                │
                        │ * No authentication challenge        │
                        └──────────────────────────────────────┘
```

## 🔑 Key Differences

### Detection Point
| Before | After |
|--------|-------|
| Line 347 (after processing) | Line 277 (immediately in onInvite) |
| Inside processInvite() | Before processInvite() |

### Matching Strategy
| Before | After |
|--------|-------|
| Only Call-ID + CSeq | **To tag** (definitive proof) |
| Unreliable for outbound calls | Works for all calls |

### Performance
| Before | After |
|--------|-------|
| Full INVITE processing overhead | Early exit on tag match |
| Creates temporary objects | Zero overhead |

### Correctness
| Before | After |
|--------|-------|
| X Fails for outbound re-INVITEs | * RFC 3261 compliant |
| X Sends auth challenges | * Immediate 200 OK |

## 📊 Dialog State Diagram

```
                        ┌─────────────────────┐
                        │   Call Established  │
                        │                     │
                        │  byLocalTag:        │
                        │  [SCL_xyz] → call   │
                        │                     │
                        │  byCallID:          │
                        │  [abc123] → call    │
                        │                     │
                        │  byRemoteTag:       │
                        │  [gK69...] → call   │
                        └──────────┬──────────┘
                                   │
                    ┌──────────────┼──────────────┐
                    │              │              │
         ┌──────────▼────────┐     │    ┌─────────▼─────────┐
         │  New INVITE       │     │    │  re-INVITE        │
         │  (no To tag)      │     │    │  (has To tag)     │
         └──────────┬────────┘     │    └─────────┬─────────┘
                    │              │              │
                    ▼              │              ▼
         ┌──────────────────────┐  │    ┌────────────────────┐
         │ processInvite()      │  │    │ handleReinvite()   │
         │ - New dialog         │  │    │ - Same dialog      │
         │ - New tags           │  │    │ - Same tags        │
         │ - New CallID         │  │    │ - Same CallID      │
         └──────────────────────┘  │    └────────────────────┘
                                   │
                                   ▼
                        ┌─────────────────────┐
                        │   Multiple Dialogs  │
                        │                     │
                        │  byLocalTag:        │
                        │  [SCL_xyz] → call1  │
                        │  [SCL_abc] → call2  │
                        │                     │
                        └─────────────────────┘
```

## 🎭 From/To Direction Examples

### Outbound Call Initial INVITE (LiveKit → Provider)

```
┌────────────────────────────────────────────────┐
│ INVITE sip:+16048158957@provider.com SIP/2.0   │
│                                                │
│ From: <sip:+14373654058@lk>;tag=SCL_xyz   ←────┼── LiveKit (UAC)
│ To: <sip:+16048158957@provider>           ←────┼── Provider (UAS, no tag yet)
│ Call-ID: abc123                                │
│ CSeq: 1 INVITE                                 │
└────────────────────────────────────────────────┘
```

### Provider's 200 OK Response

```
┌────────────────────────────────────────────────┐
│ SIP/2.0 200 OK                                 │
│                                                │
│ From: <sip:+14373654058@lk>;tag=SCL_xyz   ←────┼── Same as INVITE
│ To: <sip:+16048158957@p>;tag=gK69j24      ←────┼── Provider adds its tag
│ Call-ID: abc123                                │
│ CSeq: 1 INVITE                                 │
└────────────────────────────────────────────────┘

Dialog now established:
  - Call-ID: abc123
  - Local tag (LiveKit): SCL_xyz
  - Remote tag (Provider): gK69j24
```

### Provider's Mid-Dialog re-INVITE

```
┌────────────────────────────────────────────────┐
│ INVITE sip:+14373654058@lk SIP/2.0             │
│                                                │
│ From: <sip:+16048158957@p>;tag=gK69j24    ←────┼── Provider (UAC for re-INVITE!)
│ To: <sip:+14373654058@lk>;tag=SCL_xyz     ←────┼── LiveKit (UAS for re-INVITE!)
│ Call-ID: abc123                           ←────┼── Same Call-ID
│ CSeq: 2 INVITE                                 │
└────────────────────────────────────────────────┘

Key observation:
  * To tag = SCL_xyz = LiveKit's tag = PROVES this is our dialog!
  * From/To "swapped" because Provider is now the UAC
  * This is CORRECT per RFC 3261
```

## 🧩 Why byLocalTag Works

```
Server State:
  byLocalTag = {
    "SCL_xyz": call1,  ← Outbound call to Provider
    "LK_abc":  call2,  ← Inbound call from User
    "SCL_def": call3,  ← Another outbound call
  }

Incoming re-INVITE:
  To: <...>;tag=SCL_xyz

Lookup:
  byLocalTag["SCL_xyz"] → call1 * MATCH!

Conclusion:
  This re-INVITE belongs to call1!
  Handle it within that call's context.
```

## 🎯 Summary

The fix is simple but profound:

**Instead of asking**: "Does this Call-ID exist?" (ambiguous)
**We now ask**: "Is this To tag mine?" (definitive)

The To tag is like a session ID that we generated. If we see our own session ID in an incoming request, we KNOW it's for an existing session, not a new one.

This is faster, simpler, and RFC-compliant. *

pkg/sip/inbound.go
```diff
@@ -1243,6 +1288,74 @@ func (c *inboundCall) Bye(reason ReasonHeader) {
...
+		// Resend the original 200 OK if we already accepted
+		c.log().Infow("INVITE retransmission detected",
+			"cseq", cseq.SeqNo)
+
+		// Re-send 200 OK with our current SDP
+		if sdp := c.cc.OwnSDP(); sdp != nil {
+			r := sip.NewResponseFromRequest(req, sip.StatusOK, "OK", sdp)
+			r.AppendHeader(&contentTypeHeaderSDP)
+			_ = tx.Respond(r)
+		}
```
dennwc: SDP can potentially be nil here, we should log it at least.

pkg/sip/inbound_reinvite_test.go
```diff
@@ -0,0 +1,178 @@
...
+	// Two calls with different Call-IDs should get different nonces
+	// even if generated at the same microsecond
+
+	callID1 := "call1-123@test.com"
+	callID2 := "call2-456@test.com"
+
+	// Simulate nonce generation as done in handleInviteAuth
+	// Note: We can't directly test handleInviteAuth without full server setup,
+	// but we verify the nonce format includes the Call-ID
+	nonce1 := callID1 // In actual code: fmt.Sprintf("%d-%s", time.Now().UnixMicro(), callID1)
```
dennwc: Maybe move the nonce generation to a new helper function and call it here? Otherwise this test does nothing to prevent any potential regressions.

pkg/sip/inbound.go
```diff
@@ -1243,6 +1288,74 @@ func (c *inboundCall) Bye(reason ReasonHeader) {
...
+		c.log().Errorw("no SDP available for re-INVITE response", nil)
+		_ = tx.Respond(sip.NewResponseFromRequest(req, sip.StatusInternalServerError, "No SDP available", nil))
+		return
+	}
+
+	// For now, we respond with the same SDP (session refresh, no renegotiation)
+	// TODO: Support SDP renegotiation if needed in the future
+	//       This would require parsing the offer, updating media session, etc.
+
+	r := sip.NewResponseFromRequest(req, sip.StatusOK, "OK", currentSDP)
```
dennwc: We have a `cc.AcceptAsKeepAlive(sdp)` helper for this, and after this change we should remove the old call site then, since it won't trigger.

dennwc (commented):
Thank you @briankwest! The change look good, just a few minor comments:

pkg/sip/inbound.go
briankwest: Push a new commit to address this.

pkg/sip/inbound_reinvite_test.go
briankwest: Push a new commit to address this.

pkg/sip/inbound.go
briankwest: Push a new commit to address this.

alexlivekit (commented):
Interesting.

Today, the existing solution uses `existing := s.byCallID[cc.SIPCallID()]`. Since we are only assigning a new callID, which isn't the same thing as sipCallID (naming can be better), it actually works. It's not per spec, but it works for inbound calls.

This PR actually improves two aspects:
- Using the local tag as the key, per RFC (also safer, sipCallID is set by the requestor).
- Today, looking up by call id coincidentally prevents two identical INVITE requests with different Via/Cseq headers from creating two dialogs. We should probably stop doing that.

The things can be improved:
- Retransmission are handled by sipgo, and there's no need to tinker with them explicitly
- Root cause - today is outbound calls do not register at all in byRemoteTag/byLocalTag/byCallID, which I suspect is why this PR was submitted.

By and large, redirecting the re-invite flow to a separate handling path is nice, but I suspect change could be made far simpler (swap `existing := s.byCallID[cc.SIPCallID()]` with `existing := s.byLocalTag[cc.ID()]`) to achieve both improvement points.

adityaverma-nextdim:
We're actively blocked by this issue in production. We're using LiveKit Cloud with a NetSapiens/Telecloud SIP trunk. When our AI agent makes an outbound call to a call queue (via CreateSIPParticipantRequest), the call connects successfully — but the moment a human agent picks up from the queue, NetSapiens sends a re-INVITE and LiveKit responds with 404 No trunk found, causing the call to drop immediately.

We've confirmed with our SIP provider that the re-INVITE is standard behavior during queue bridging, and the same setup works fine with Twilio and Telnyx trunks (which don't send re-INVITEs in this scenario).

We noticed PR #613 was just opened as a simpler alternative to this fix. Is there an estimated timeline for when either fix will be merged and deployed to LiveKit Cloud? This is the only thing blocking our go-live.

@briankwest Thank you for the great work on this — the RFC 3261 compliant To tag approach is exactly what's needed.

alexlivekit:
@adityaverma-nextdim , either this PR or https://github.com/livekit/sip/pull/613 is expected to land soon, but unfortunately neither will currently solve your problem - outbound calls, created via CreateSIPParticipantRequest, currently do not register for neither `byCallID` nor `byLocalTag`. The fix for this should be coming up in the next week or so, with rollout shortly after.

adityaverma-nextdim:
Hi @alexlivekit , do we have any update on the re-INVITE issue?

alexlivekit:
Closed in favor of https://github.com/livekit/sip/pull/614.

adityaverma-nextdim:
Hi @alexlivekit, 
We’re not seeing any changes in our use case—we’re still encountering the re-INVITE issue. Could you please share a tentative timeline for the resolution on your end?

Thanks.

alexlivekit:
> Could you please share a tentative timeline for the resolution on your end?

@adityaverma-nextdim , should be resolved by the end of the week, and may be resolved as soon as tomorrow.

briankwest:
I don't believe this has been fixed properly, I'm still having customers open issues against this problem with us.

alexlivekit:
> I don't believe this has been fixed properly, I'm still having customers open issues against this problem with us.

Care to share a scenario or an example?

JoeSlain:
Hey I'm having the same problem in prod :

Title: Outbound SIP: bridge stops receiving inbound in-dialog requests mid-call —
carrier re-INVITE and BYE dropped, ~80% of calls reaching 15 min are torn down

Environment
  LiveKit Cloud (project id available privately)
  Carrier: CALLR B2BUA/0.42.1, digest outbound trunk -> eu.sip.callr.com

Summary
On OUTBOUND calls, after ~15 minutes with no SIP traffic on the dialog, the bridge
stops receiving inbound in-dialog requests sent to the Contact it advertised.
The carrier's session-timer re-INVITE and its follow-up BYE are both dropped and
never reach the bridge. Requests sent FROM that same address still arrive at the
carrier. Inbound calls are unaffected. We have packet captures from both sides
that agree exactly.

Reference call
  Room SID     RM_KzXuhzJAqyyf
  SIP Call ID  SCL_gxoSC3RP2nHX
  Bridge Contact  sip:<BRIDGE_IP>:9000     (our project's outbound SIP address)
  Carrier         185.76.22.162:5060
  Answered        2026-08-31 09:35:59.320 CEST   (all times below CEST)

Negotiation
The bridge's INVITE does not advertise "Supported: timer" and its Allow list omits
UPDATE. The carrier's 200 OK returns:
    Session-Expires: 1800;refresher=uas
Per RFC 4028 the carrier becomes the refresher and must refresh at 900s.

Carrier-side capture
  09:50:55.557  INVITE sip:<BRIDGE_IP>:9000
                Session-Expires: 1800;refresher=uac, Min-SE: 120
                correct Call-ID and dialog tags, R-URI = the advertised Contact
                10 retransmissions through 09:51:23.044 — NO response
  09:51:25.555  BYE sip:<BRIDGE_IP>:9000
                10 retransmissions through 09:51:53.044 — NO response
                X-RTP-Statistics: 93060 packets, 0 errors
  09:52:02.993  BYE from <BRIDGE_IP>:9000 (our application hangup) ARRIVES at the
                carrier and is answered 481 Call Does Not Exist

Bridge-side capture (dashboard, Telephony > Calls > PCAP)
  94,910 packets, of which only 10 are SIP — all during setup, last one the ACK at
  09:35:59.328. None of the 20 packets above appear. A raw string scan of the whole
  file gives BYE=0, OPTIONS=0, UPDATE=0, NOTIFY=0, INVITE=2. No ICMP.
  RTP carrier->bridge: 46,757 packets, regular at 20ms, no gap over 0.5s across
  15m35, last packet 09:51:25.609.
  RTP bridge->carrier: 48,143 packets, continues normally until 09:52:02.245.

The failure is directional and outbound-only
Packets leaving <BRIDGE_IP>:9000 arrive and are answered. Packets entering it are
dropped. There was no SIP traffic in either direction for 15m56s before the first
re-INVITE.

Comparing 21 days of call durations makes the scope clear — 1,836 outbound vs
1,693 inbound calls on the same project and the same carrier:

                    INBOUND      OUTBOUND
     600- 660s :      11            13
     660- 720s :      14            19
     720- 780s :       7             6
     780- 840s :       7            11
     840- 900s :       9             9
     900- 960s :      12            37     <-- 4x the neighbouring bands
     960-1020s :       3             3
    1020-1080s :       4             0
    1080-1140s :       3             0
      >= 1140s :      23             6

Inbound decays smoothly and 23 calls run past 19 minutes. Outbound spikes at the
900s refresh point and then goes to exactly zero for two consecutive bands.

Our reading: on inbound calls the bridge's SIP port is permanently listening, so
there is a static path. On outbound the bridge originates from <BRIDGE_IP>:9000 and
the return path appears to rely on a dynamic mapping that expires during a long
silent dialog. The bridge advertises a public Contact but sends no keepalive of any
kind, and does not support session timers, so nothing keeps that path open.

Impact
Roughly 80% of outbound calls that reach 15 minutes are torn down. Users also get a
ghost leg: because the carrier's BYE is dropped too, the SIP participant stays in
the room publishing digital silence and the caller talks to nobody until they hang
up — 36 seconds in the reference call.

Asks
1. Keep the inbound path to the advertised Contact alive on outbound calls — a
   periodic keepalive, or a persistent binding instead of a dynamic one.
2. Advertise "Supported: timer" and handle session refreshes, so this common
   carrier behaviour is handled rather than dropped.
3. Independently: is there any leg-liveness detection that is not media-based?
   SIPMediaConfig.mediaTimeout is currently the only safety net; we raised it to
   600s to avoid false teardowns under carrier silence suppression, so a genuinely
   dead leg now lingers for up to 10 minutes.

Both packet captures, the project id and the bridge address are available privately
to a LiveKit engineer — they contain end-user phone numbers and SIP credentials, so
we are not attaching them here.

PR #584, opened 2026-02-10
E2E enablement: Allowing to specify the full CreateSipParticipant in tests

alexlivekit:
This also adds support for individual room names, enabling concurrent tests.

### DIFF
--- a/test/lktest/sip.go
+++ b/test/lktest/sip.go
@@ -383,13 +383,8 @@ Check logs for call:
 	}
 }
 
-type SIPOutboundRequestTestParams struct {
-	Req *livekit.CreateSIPParticipantRequest
-}
-
 type SIPOutboundRequestTestIDs struct {
 	CallID        string
-	SipCallID     string
 	PatricipantID string
 	RoomName      string
 	TrunkID       string
@@ -403,6 +398,16 @@ func (ids *SIPOutboundRequestTestIDs) SetFromCreateSIPParticipantResponse(resp *
 	ids.CallID = resp.SipCallId
 	ids.PatricipantID = resp.ParticipantId
 	ids.RoomName = resp.RoomName
+}
+
+func (ids *SIPOutboundRequestTestIDs) GetValues() []string {
+	return []string{
+		"callID: " + ids.CallID,
+		"patricipantID: " + ids.PatricipantID,
+		"roomName: " + ids.RoomName,
+		"trunkID: " + ids.TrunkID,
+		"location: " + ids.Location,
+	}
 }
 
 func getInboundTrunksByNumbers(ctx context.Context, lkIn *LiveKit, numbers []string) ([]*livekit.SIPInboundTrunkInfo, error) {
@@ -437,43 +442,48 @@ func getDispatchRulesByTrunks(ctx context.Context, lkIn *LiveKit, trunks []*live
 
 type roomIDFunc func(ctx context.Context, lk *LiveKit, rule *livekit.SIPDispatchRuleInfo, req *livekit.CreateSIPParticipantRequest) (string, error)
 
-func TestSIPOutboundRequest(t TB, ctx context.Context, lkOut, lkIn *LiveKit, params SIPOutboundRequestTestParams) error {
+func secondsSince(start time.Time) int {
+	return int(time.Since(start).Seconds())
+}
+
+func TestCreateSipParticipant(t TB, ctx context.Context, lkOut, lkIn *LiveKit, req *livekit.CreateSIPParticipantRequest) error {
+	start := time.Now()
 	inIDs := SIPOutboundRequestTestIDs{}
 	outIDs := SIPOutboundRequestTestIDs{}
 	defer func() {
-		t.Logf("Onbound IDs: %v", outIDs)
-		t.Logf("Inbound IDs: %v", inIDs)
+		t.Logf("Onbound IDs: %v", outIDs.GetValues())
+		t.Logf("Inbound IDs: %v", inIDs.GetValues())
 	}()
 
-	require.Equal(t, "", params.Req.SipTrunkId, "SipTrunkId must be empty")
-	require.NotNil(t, params.Req.Trunk, "A trunk must be inlined")
+	require.Equal(t, "", req.SipTrunkId, "SipTrunkId must be empty")
+	require.NotNil(t, req.Trunk, "A trunk must be inlined")
 	outIDs.TrunkID = "inline"
-	require.NotEmpty(t, params.Req.Trunk.DestinationCountry, "DestinationCountry must be set")
-	outIDs.Location = params.Req.Trunk.DestinationCountry
-	require.NotEmpty(t, params.Req.Trunk.Hostname, "Hostname must be set")
-	inIDs.Location = params.Req.Trunk.Hostname
-	require.NotEmpty(t, params.Req.SipCallTo, "SipCallTo must be set")
-	require.NotEmpty(t, params.Req.RoomName, "RoomName must be set")
-	outIDs.RoomName = params.Req.RoomName
-	t.Logf("Calling from %s@%s to %s@%s", params.Req.SipNumber, params.Req.Trunk.DestinationCountry, params.Req.SipCallTo, params.Req.Trunk.Hostname)
-	t.Logf("Using outbound room %q", params.Req.RoomName)
+	require.NotEmpty(t, req.Trunk.DestinationCountry, "DestinationCountry must be set")
+	outIDs.Location = req.Trunk.DestinationCountry
+	require.NotEmpty(t, req.Trunk.Hostname, "Hostname must be set")
+	inIDs.Location = req.Trunk.Hostname
+	require.NotEmpty(t, req.SipCallTo, "SipCallTo must be set")
+	require.NotEmpty(t, req.RoomName, "RoomName must be set")
+	outIDs.RoomName = req.RoomName
 	outClosed := make(chan struct{}, 1)
 	inClosed := make(chan struct{}, 1)
 
-	trsIn, err := getInboundTrunksByNumbers(ctx, lkIn, []string{params.Req.SipCallTo})
+	t.Logf("+%ds: Getting inbound trunk", secondsSince(start))
+	trsIn, err := getInboundTrunksByNumbers(ctx, lkIn, []string{req.SipCallTo})
 	if err != nil {
 		return err
 	}
 	trIn := trsIn[0]
-	t.Logf("Expecting call in inbound trunk %q (%s, via number: %s)", trIn.Name, trIn.SipTrunkId, params.Req.SipCallTo)
 	inIDs.TrunkID = trIn.SipTrunkId
+
+	t.Logf("+%ds: Getting dispatch rule", secondsSince(start))
 	rulesIn, err := getDispatchRulesByTrunks(ctx, lkIn, trsIn)
 	if err != nil {
 		return err
 	}
 	ruleIn := rulesIn[0]
-	t.Logf("Expecting call in dispatch rule %q (%s)", ruleIn.Name, ruleIn.SipDispatchRuleId)
 
+	t.Logf("+%ds: Getting room ID function", secondsSince(start))
 	var getRoomID roomIDFunc
 	if _, ok := ruleIn.Rule.Rule.(*livekit.SIPDispatchRule_DispatchRuleIndividual); ok {
 		// Inbound room name is dynamic: e2e_{SipNumber}_{guid}. SipNumber is unique per test, so we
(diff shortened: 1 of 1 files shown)


test/lktest/sip.go
```diff
@@ -381,3 +382,296 @@ Check logs for call:
...
+
+	// 3. Find inbound room by prefix (poll until it appears)
+	roomIn, err := getRoomID(ctx, lkIn, ruleIn, req)
+	if err != nil {
+		return outIDs, nil, fmt.Errorf("find inbound room: %w", err)
+	}
+	t.Logf("DEBUG: Found inbound room: %s", roomIn)
+
+	t.Cleanup(func() {
+		_, _ = lkOut.Rooms.DeleteRoom(context.Background(), &livekit.DeleteRoomRequest{Room: roomOut})
```
dennwc: Cleanup is similar to defer, it should be called as soon as the object is created. In this case, for example, outbound room is not cleaned up when we test for encryption failure. So it should be called right after we connect to outbound room instead.

test/lktest/sip.go
```diff
@@ -381,3 +382,296 @@ Check logs for call:
...
+	}
+	if resp == nil || len(resp.Items) == 0 {
+		return nil, fmt.Errorf("no dispatch rules found for trunks: %v", ids)
+	}
+	return resp.Items, nil
+}
+
+type roomIDFunc func(ctx context.Context, lk *LiveKit, rule *livekit.SIPDispatchRuleInfo, req *livekit.CreateSIPParticipantRequest) (string, error)
+
+func TestSIPOutboundRequest(t TB, ctx context.Context, lkOut, lkIn *LiveKit, params SIPOutboundRequestTestParams) (outIDs, inIDs *SIPOutboundRequestTestIDs, err error) {
```
dennwc: Would be great to call it as a part of E2E in OSS as well, as we do with the old test.

test/lktest/sip.go
```diff
@@ -381,3 +382,346 @@ Check logs for call:
...
+	}
+	if resp == nil || len(resp.Items) == 0 {
+		return nil, fmt.Errorf("no dispatch rules found for trunks: %v", ids)
+	}
+	return resp.Items, nil
+}
+
+type roomIDFunc func(ctx context.Context, lk *LiveKit, rule *livekit.SIPDispatchRuleInfo, req *livekit.CreateSIPParticipantRequest) (string, error)
+
+func TestSIPOutboundRequest(t TB, ctx context.Context, lkOut, lkIn *LiveKit, params SIPOutboundRequestTestParams) error {
```
hechen-eng: Shall we rename it? The existing test is `TestSIPOutbound` and this one is `TestSIPOutboundRequest`, which is a bit confusing at first sight

pkg/sip/media_port.go
```diff
@@ -604,7 +604,7 @@ func (p *MediaPort) SetAnswer(offer *sdp.Offer, answerData []byte, enc sdp.Encry
 	}
 	mc, err := answer.Apply(offer, enc)
 	if err != nil {
-		return nil, err
+		return nil, SDPError{Err: err}
```
hechen-eng, line 607: this is just a fix not related to e2e, right?

test/lktest/sip.go
```diff
@@ -381,3 +382,346 @@ Check logs for call:
 		CheckAudioForParticipants(t, ctx, pOut, pIn)
 	}
 }
+
+type SIPOutboundRequestTestParams struct {
```
hechen-eng: why do we need this struct?

test/lktest/sip.go
```diff
@@ -381,3 +382,346 @@ Check logs for call:
...
+	if err != nil {
+		return err
+	}
+	ruleIn := rulesIn[0]
+	t.Logf("Expecting call in dispatch rule %q (%s)", ruleIn.Name, ruleIn.SipDispatchRuleId)
+
+	var getRoomID roomIDFunc
+	if _, ok := ruleIn.Rule.Rule.(*livekit.SIPDispatchRule_DispatchRuleIndividual); ok {
+		// Inbound room name is dynamic: e2e_{SipNumber}_{guid}. SipNumber is unique per test, so we
+		// create the outbound call first, then poll for a room whose name has prefix "e2e_"+SipNumber+"_".
```
hechen-eng, line 490: I wonder do we need to test `SIPDispatchRule_DispatchRuleIndividual` at all for this E2E. Is `SIPDispatchRule_DispatchRuleDirect` sufficient for our purposes?

test/lktest/sip.go
```diff
@@ -381,3 +382,346 @@ Check logs for call:
...
+
+	if req.MediaEncryption == livekit.SIPMediaEncryption_SIP_MEDIA_ENCRYPT_DISABLE && trIn.MediaEncryption == livekit.SIPMediaEncryption_SIP_MEDIA_ENCRYPT_REQUIRE {
+		// CreateSipParticipant request ddisables encryption, that is required by trunk
+		req.WaitUntilAnswered = true // We expect this to get rejected
+		ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+		defer cancel()
+		resp, err := lkOut.SIP.CreateSIPParticipant(ctx, req)
+		outIDs.SetFromCreateSIPParticipantResponse(resp)
+		if err == nil {
+			_, _ = lkOut.Rooms.RemoveParticipant(context.Background(), &livekit.RoomParticipantIdentity{
```
hechen-eng, line 561: isn't outbound SIP participant deleted automatically by CreateSIPPartition when it disconnects LiveKit server?

test/lktest/sip.go
```diff
@@ -381,3 +382,346 @@ Check logs for call:
...
+	}
+	const outIdentity = "siptest_outbound"
+	const outName = "Outbound Call"
+	const outMeta = `{"test":true, "dir": "out"}`
+	req.ParticipantIdentity = outIdentity
+	req.ParticipantName = outName
+	req.ParticipantMetadata = outMeta
+
+	if req.MediaEncryption == livekit.SIPMediaEncryption_SIP_MEDIA_ENCRYPT_DISABLE && trIn.MediaEncryption == livekit.SIPMediaEncryption_SIP_MEDIA_ENCRYPT_REQUIRE {
+		// CreateSipParticipant request ddisables encryption, that is required by trunk
```
hechen-eng:
```suggestion
		// CreateSipParticipant request disables encryption, that is required by trunk
```

test/lktest/sip.go
```diff
@@ -381,3 +382,346 @@ Check logs for call:
...
+	r := lkOut.CreateSIPParticipant(t, req) // Also adds cleanup!
+	outIDs.SetFromCreateSIPParticipantResponse(r)
+
+	t.Logf("Finding inbound room")
+
+	// 3. Find inbound room by prefix (poll until it appears)
+	subCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+	start := time.Now()
+	roomIn, err := getRoomID(subCtx, lkIn, ruleIn, req) // Takes about 5 seconds to propagate
```
hechen-eng: 5 seconds to be very long for an RTC infrastructure, but I guess it is what it is...

test/lktest/sip.go
```diff
@@ -381,3 +382,346 @@ Check logs for call:
...
+		return err
+	}
+	trIn := trsIn[0]
+	t.Logf("Expecting call in inbound trunk %q (%s, via number: %s)", trIn.Name, trIn.SipTrunkId, params.Req.SipCallTo)
+	inIDs.TrunkID = trIn.SipTrunkId
+	rulesIn, err := getDispatchRulesByTrunks(ctx, lkIn, trsIn)
+	if err != nil {
+		return err
+	}
+	ruleIn := rulesIn[0]
```
hechen-eng, line 484: curious, why do we choose the first trunk and first rule, i.e. `trIn := trsIn[0]` and `rulesIn[0]`?

test/lktest/sip.go
```diff
@@ -381,3 +382,346 @@ Check logs for call:
...
+					if rp.Identity() != identityTest {
+						close(inClosed)
+					}
+				},
+			},
+		})
+	}()
+	readyIn.Wait()
+
+	if req.MediaEncryption == livekit.SIPMediaEncryption_SIP_MEDIA_ENCRYPT_REQUIRE && trIn.MediaEncryption == livekit.SIPMediaEncryption_SIP_MEDIA_ENCRYPT_DISABLE {
```
hechen-eng: Don't we fail when `CreateSIPParticipant` in this case?

test/lktest/sip.go
```diff
@@ -381,3 +382,346 @@ Check logs for call:
...
+		// FIXME
+		// We should be able to reject calls immediately when the SDP mismatches (on the inbound call),
+		// but today this is delayed until after attempting to answer the call.
+		// At this point both calls should be dead or dying. Verify.
+		t.Logf("%v DEBUG: Expecting call failure due to cryptography requirement mismatch", time.Now())
+		subCtx, cancel = context.WithTimeout(ctx, 10*time.Second)
```
hechen-eng: recommend to increase to 30s to reduce noise..

hechen-eng (approved):
minor comments...

test/lktest/sip.go
alexlivekit: It's a leftover from beefier params in the past that I've been using to pass debug params locally.
Good catch, will drop.

test/lktest/sip.go
alexlivekit: `SIPDispatchRule_DispatchRuleDirect` would not be able to support concurrent cluster tests (e.g. the same room is going to be used when testing `ochicago1b` and `oashburn1b`). That's a no-go.

test/lktest/sip.go
alexlivekit: Not sure I follow here. Once a call has been set up, `CreateSipParticipant` is complete and will not be doing anything on its own.

So in this case, if no error occurred (meaning something, somehow, answered), we would need to manually hang up the call before failing the test, and we do that here via `RemoveParticipant`.

test/lktest/sip.go
alexlivekit: You're not going to like my update. We're seeing propagation times of over 15 seconds in prod in some extreme cases.

test/lktest/sip.go
alexlivekit: Because of how we configured the projects - there's only one trunk per destination number, and only one dispatch rule total.

test/lktest/sip.go
alexlivekit: No, because it is an async call. We don't specify `req.WaitUntilAnswered = true`, so it starts the outbound call and returns success. Furthermore, we don't even realize this doesn't have a chance to succeed (encryption mismatch) until we're trying to answer internally the call on the other end.

pkg/sip/media_port.go
alexlivekit: Yup. 
This bit of code then attempts to figure out what went wrong:
```go
	if err := c.dialSIP(ctx, tid); err != nil {
		c.log.Infow("SIP call failed", "error", err)

		reportErr := err
		status, desc, reason := callDropped, "invite-failed", livekit.DisconnectReason_UNKNOWN_REASON
		var e *livekit.SIPStatus
		if errors.As(err, &e) {
			switch int(e.Code) {
			case int(sip.StatusTemporarilyUnavailable):
				status, desc, reason = callUnavailable, "unavailable", livekit.DisconnectReason_USER_UNAVAILABLE
				reportErr = nil
			case int(sip.StatusBusyHere):
				status, desc, reason = callRejected, "busy", livekit.DisconnectReason_USER_REJECTED
				reportErr = nil
			}
		} else if errors.Is(err, sdp.ErrNoCommonCrypto) {
			status, desc, reason = callRejected, "encryption-required", livekit.DisconnectReason_MEDIA_FAILURE
			reportErr = nil
		}
		c.close(ctx, reportErr, status, desc, reason)
		return err
	}
```
The thing is, `errors.Is(err, sdp.ErrNoCommonCrypto)` doesn't actually check if `err == sdp.ErrNoCommonCrypto`, so we wrap it. It's a little silly, but:
1. We already do it in `runMediaConn()` on inbound calls.
2. It allows any intermediate function to actually wrap this error without braking. So much more maintainable.

test/lktest/sip.go
hechen-eng: at line 147, the old comment says: `// Make sure we delete inbound SIP participant. Outbound is deleted automatically by CreateSIPParticipant.` and it didn't remove participant for outbound. Am I missing something here 🤔

alexlivekit: Ah, thanks for clarifying the question.
`lkOut.CreateSIPParticipant(ctx, req)` Automatically adds cleanup for the participant.
`lkOut.SIP.CreateSIPParticipant(ctx, req)`, as we use here, does not. That's why this error path manually does that.

PR #595, opened 2026-02-20
Resolve domain names in NAT1To1IP and MediaNAT1To1IP config

jaylim95:
Previously, these fields only accepted literal IP addresses via netip.ParseAddr. Added support for DNS resolution when SIP LB does not have a static IP but has a stable domain name.

Added resolveAddrOrHost helper that first tries parsing as a literal IP, then falls back to DNS resolution via net.LookupHost, returning the first IPv4 result.

### DIFF
--- a/pkg/sip/config.go
+++ b/pkg/sip/config.go
@@ -38,7 +38,7 @@ func GetServiceConfig(conf *config.Config) (*ServiceConfig, error) {
 			return nil, err
 		}
 	} else if conf.NAT1To1IP != "" {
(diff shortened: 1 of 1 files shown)


dennwc (approved):
Thank you! LGTM after CLA is signed!

jaylim95:
I tried clicking on sign CLA for like 20 times in the last 3 weeks but the status never update...

dennwc:
Let me check. I guess this is because the commit is authored by Claude, not your email.

PR #599, opened 2026-02-25
TEL-372: log feature flag

hechen-eng:


pkg/sip/inbound.go
```diff
@@ -698,6 +698,12 @@ func (c *inboundCall) handleInvite(ctx context.Context, tid traceid.ID, req *sip
 	if disp.DispatchRuleID != "" {
 		c.appendLogValues("sipRule", disp.DispatchRuleID)
 	}
+	for key, value := range disp.FeatureFlags {
+		c.appendLogValues("featureFlag_"+key, value)
+	}
+	for _, feature := range disp.EnabledFeatures {
+		c.appendLogValues(feature.String(), "true")
+	}
```
alexlivekit, line 706: Not currently necessary imo - we already have logs of all requests coming into the system.

PR #607, opened 2026-03-09
TEL-331: add start ringing metric

hechen-eng:


dennwc:
Sorry, missed that this one already exists. I filed https://github.com/livekit/sip/pull/609 which might cover more cases.

nishadmusthafa:
we can close this right @hechen-eng ?

PR #609, opened 2026-03-10
Instrument inbound call stages.

dennwc:


hechen-eng:
I was wondering whether datadog is a better place for events logging for this purpose 🤔

There are two aspects here. One is that we want get alerted and the corresponding metrics should be in prometheus and grafana. The other is that we may want to do data analysis (e.g. see which part takes the most time or in which region) or trouble shooting and it's better in datadog.

For inbound alerting, is it sufficient to just alert on `invite --> 200 OK`? If it's over some threshold, we get alerts and then we dive deeper into each step in datadog logs.

PR #612, opened 2026-03-11
TEL-420: use psrpc error if possible

hechen-eng:


pkg/sip/service.go
```diff
@@ -451,7 +451,7 @@ func (s *Service) validateCallProvider(state *CallState) error {
 
 	// Check if provider is internal and prevent transfer is enabled
 	if state.callInfo.ProviderInfo.Type == livekit.ProviderType_PROVIDER_TYPE_INTERNAL && state.callInfo.ProviderInfo.PreventTransfer {
-		return fmt.Errorf("we don't yet support transfers for this phone number type")
+		return psrpc.NewErrorf(psrpc.InvalidArgument, "we don't yet support transfers for this phone number type")
```
nishadmusthafa: I think we can use https://github.com/livekit/psrpc/blob/eeeec9735242b93037a4ef413ca078e8e1ebbdfe/errors.go#L314 (Unimplemented) here

pkg/sip/outbound.go
```diff
@@ -393,9 +393,14 @@ func (c *outboundCall) connectSIP(ctx context.Context, tid traceid.ID) error {
 				status, desc, reason = callRejected, "busy", livekit.DisconnectReason_USER_REJECTED
 				reportErr = nil
 			}
+		} else if errors.Is(err, sdp.ErrNoCommonMedia) {
+			status, desc, reason = callRejected, "no-common-codec", livekit.DisconnectReason_MEDIA_FAILURE
+			reportErr = nil
+			err = psrpc.NewError(psrpc.InvalidArgument, err)
```
nishadmusthafa: FailedPrecondition is a better option here

pkg/sip/outbound.go
```diff
@@ -159,7 +159,7 @@ func (c *Client) newCall(ctx context.Context, tid traceid.ID, conf *config.Confi
 	call.media.DisableOut() // disabled until we get 200
 	if err := call.connectToRoom(ctx, room, c.getRoom); err != nil {
 		call.close(ctx, errors.Wrap(err, "room join failed"), callDropped, "join-failed", livekit.DisconnectReason_UNKNOWN_REASON)
-		return nil, fmt.Errorf("update room failed: %w", err)
+		return nil, psrpc.NewError(psrpc.Internal, fmt.Errorf("update room failed: %w", err))
```
nishadmusthafa, line 162: Just clarifying.. this will still be a 500 right ? I agree that it should be the case.

pkg/sip/outbound.go
```diff
@@ -393,9 +393,14 @@ func (c *outboundCall) connectSIP(ctx context.Context, tid traceid.ID) error {
...
 				reportErr = nil
 			}
+		} else if errors.Is(err, sdp.ErrNoCommonMedia) {
+			status, desc, reason = callRejected, "no-common-codec", livekit.DisconnectReason_MEDIA_FAILURE
+			reportErr = nil
+			err = psrpc.NewError(psrpc.InvalidArgument, err)
 		} else if errors.Is(err, sdp.ErrNoCommonCrypto) {
 			status, desc, reason = callRejected, "encryption-required", livekit.DisconnectReason_MEDIA_FAILURE
 			reportErr = nil
+			err = psrpc.NewError(psrpc.InvalidArgument, err)
```
nishadmusthafa: Should this be FailedPrecondition too?

pkg/sip/outbound.go
hechen-eng: yeah, it'll be a 5xx 👍

pkg/sip/outbound.go
hechen-eng: agreed 👍

PR #613, opened 2026-03-12
Adding reinvite test and rekeying by local tag

alexlivekit:
Alternative option for what was proposed in https://github.com/livekit/sip/pull/582
This is only part 1, and should be followed up by adding outbound calls to active maps.
Part 3 should be whatever is needed to get the outbound calls to also be mapped.


alexlivekit:
Replaced by #614

PR #616, opened 2026-03-13
fix: detach dial context from rpc deadline when wait_until_answered is true

rahul20110:
When `WaitUntilAnswered=true`, `call.Dial(ctx)` uses the Twirp/psrpc request
  context directly. When the RPC deadline fires, the context cancels, which
  triggers `tx.Cancel()` sending a SIP CANCEL to the provider — dropping the
  call and causing a retry loop.

  `DialAsync` already handles this correctly by calling
  `context.WithoutCancel(ctx)` before dialing. This applies the same fix to the
  synchronous `Dial` path.


alexlivekit:
The idea of `WaitUntilAnswered=true` isn't to turn the API call into an async call once we hit the deadline. Us originating the cancel is done with intention.

The retry loop, however, needs to be fixed. If you have recent examples of this behavior, please share them with the team and we will happily look into those.

As is, this isn't a direction we would want to move in.

PR #622, opened 2026-03-16
Adding basic outbound reINVITE support

alexlivekit:
Depends on https://github.com/livekit/media-sdk/pull/48

### DIFF
--- a/pkg/sip/inbound_process_invite_test.go
+++ b/pkg/sip/inbound_process_invite_test.go
@@ -155,7 +155,7 @@ func NewInboundTest(t *testing.T) *InboundTest {
 
 // RegisterOutboundCallForReinvite registers a fake outbound call so that a re-INVITE
 // with the given localTag (To tag) is accepted as outbound reinvite and answered with sdpOffer.
-func (it *InboundTest) RegisterOutboundCallForReinvite(t *testing.T, localTag LocalTag) (offer, answer, localSDP []byte) {
+func (it *InboundTest) RegisterOutboundCallForReinvite(t *testing.T, localTag LocalTag) (offer, answer, localSDPBytes []byte) {
 	t.Helper()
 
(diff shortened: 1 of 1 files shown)


pkg/sip/inbound.go
```diff
@@ -338,11 +338,23 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
...
-		log.Infow("accepting reinvite", "content-type", req.ContentType(), "content-length", req.ContentLength())
 		existing.log().Infow("reinvite", "content-type", req.ContentType(), "content-length", req.ContentLength(), "cseq", cc.InviteCSeq())
 		cc.AcceptAsKeepAlive(existing.cc.OwnSDP())
 		return nil
 	}
+	if s.cli != nil { // Process reinvite for existing outbound calls
+		oc := s.cli.GetActiveCall(cc.ID())
+		newCSeq := cc.InviteCSeq()
+		if oc != nil && oc.cc != nil && oc.cc.InviteCSeq() < newCSeq {
+			sdp := oc.cc.LocalSDP()
```
hechen-eng, line 349: This assumes that we don't need to re-negotiate SDP. What's the effort to actually re-negotiate it during re-invite?

alexlivekit: Not supported atm - we would need to (potentially) change up media pipeline wiring.
It's not all that scary, but out of scope here.

pkg/sip/client.go
```diff
@@ -168,6 +168,12 @@ func (c *Client) CreateSIPParticipant(ctx context.Context, req *rpc.InternalCrea
 	return c.createSIPParticipant(ctx, req)
 }
 
+func (c *Client) GetActiveCall(tag LocalTag) *outboundCall {
```
dennwc: This should ideally return an interface, instead of a private/unexported type. Alternatively, you could unexport the method and keep it as-is.

hechen-eng (approved):

PR #625, opened 2026-03-19
Fixing up TransferSipParticipant when receiving BYE after REFER

alexlivekit:


pkg/sip/inbound.go
```diff
@@ -1974,7 +1974,7 @@ func (c *sipInbound) newReferReq(transferTo string, headers map[string]string) (
 	return req, nil
 }
 
-func (c *sipInbound) TransferCall(ctx context.Context, transferTo string, headers map[string]string) error {
+func (c *sipInbound) TransferCall(ctx context.Context, transferTo string, headers map[string]string, ctxDone <-chan struct{}) error {
```
dennwc:
```suggestion
func (c *sipInbound) TransferCall(ctx context.Context, transferTo string, headers map[string]string, callDone <-chan struct{}) error {
```

PR #629, opened 2026-03-23
reduce connect timeout to 1s

davidzhao:
also bumped server-sdk-go and mediatransportutil


test/lktest/livekit.go
```diff
@@ -115,12 +115,13 @@ func (lk *LiveKit) CreateSIPParticipantSync(t TB, req *livekit.CreateSIPParticip
...
 func (lk *LiveKit) Connect(t TB, room, identity string, cb *lksdk.RoomCallback) *lksdk.Room {
 	r := lksdk.NewRoom(cb)
+	// faster connection timeout since they should be in the same DC
 	err := r.Join(lk.WsUrl, lksdk.ConnectInfo{
 		APIKey:              lk.ApiKey,
 		APISecret:           lk.ApiSecret,
 		RoomName:            room,
 		ParticipantIdentity: identity,
-	})
+	}, lksdk.WithConnectTimeout(1*time.Second))
```
hechen-eng, line 124: Is it only needed in test? Meanwhile, do we have data showing join room is fast in the same DC?  We recently found join room p99 latency is not so great..

alexlivekit: +1 , this might be a source of some flaky pages, but in all honesty if we have bandwidth to look into that we should

PR #630, opened 2026-03-23
Refactoring reinvite tests, adding transfer tests

alexlivekit:


pkg/sip/outbound.go
```diff
@@ -282,7 +282,6 @@ func (c *outboundCall) Disconnected() <-chan struct{} {
 }
 
 func (c *outboundCall) Close(ctx context.Context) error {
-	c.closing.Break()
```
hechen-eng: this change is probably not needed since `core.Fuse` handles mutex itself.

pkg/sip/signaling_test.go
```diff
@@ -0,0 +1,908 @@
...
+}
+
+func (s *sipUATest) UnregisterSink(localTag string, method string) bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	dialogMethods, ok := s.sinks[localTag]
+	if !ok {
+		return false
+	}
+	sink, ok := dialogMethods[localTag]
```
hechen-eng: should it be?

```suggestion
	sink, ok := dialogMethods[method]
```

pkg/sip/signaling_test.go
```diff
@@ -0,0 +1,908 @@
...
+	if !ok {
+		return false
+	}
+	close(sink)
+	delete(dialogMethods, localTag)
```
hechen-eng: should it be?

```suggestion
	delete(dialogMethods, method)
```

PR #633, opened 2026-03-25
Modify Record-Route handling

alexlivekit:
Building on top of the tests added in https://github.com/livekit/sip/pull/635


pkg/siptest/client.go
```diff
@@ -432,10 +432,7 @@ func (c *Client) sendBye() {
 	select {
 	case <-c.ack:
 	case <-tx.Done():
-	case r := <-tx.Responses():
-		if r.StatusCode == 200 {
-			_ = c.sipClient.WriteRequest(sip.NewAckRequest(req, r, nil))
-		}
```
alexlivekit, line 438: AFAICT, we should not be sending ACK for BYEs.
That goes for the others here too...

hechen-eng: +1 that we should not ack bye. Created a linear a while back for this https://linear.app/livekit/issue/TEL-391/do-not-ack-to-200-ok-bye

hechen-eng (approved):
Alt Text

amr-imad:
Hi, hope this is useful — while investigating a transfer failure on our end, we noticed this fix appears to have been reverted a week later by 0b27a4b2 (#643 ), whose commit message reads Revert "Modify Record-Route handling (#633 )". No release was tagged in between, so it looks like it never shipped, and current HEAD still has the pre-fix logic in NewReferRequest (protocol.go) and swapSrcDst (inbound.go).

We're hitting what looks like this exact issue in production — details and captures in #771 . Would really appreciate it if this could be looked at and re-applied when you get a chance.

cc @alexlivekit  @hechen-eng — thank you!

amr-imad:
Hi — following up from #633 : that fix appears to have been reverted a week later by #643  (commit 0b27a4b2, message Revert "Modify Record-Route handling (#633 )"), and it looks like it went unnoticed partly because these tests still pass on current HEAD despite the revert.

Digging into why: the reverted code path plus the leftover header-conversion step in swapSrcDst apply the Record-Route reversal twice, and for the two-proxy case here those two reversals cancel out, landing on the correct order by coincidence. So this suite doesn't currently catch the regression.

We're seeing what looks like the real-world version of this in production — details in #771 . Might be worth a case closer to that (more hops, or with add_record_route enabled per #642 ) to actually catch it going forward.

cc @alexlivekit @hechen-eng — thanks for the original fix, hope this helps get it re-applied!

hechen-eng:
> Hi — following up from #633 : that fix appears to have been reverted a week later by #643 (commit [0b27a4b](https://github.com/livekit/sip/commit/0b27a4b24ebfd5227566bd3f53c5fee90e814f63), message Revert "Modify Record-Route handling (#633 )"), and it looks like it went unnoticed partly because these tests still pass on current HEAD despite the revert.
> 
> Digging into why: the reverted code path plus the leftover header-conversion step in swapSrcDst apply the Record-Route reversal twice, and for the two-proxy case here those two reversals cancel out, landing on the correct order by coincidence. So this suite doesn't currently catch the regression.
> 
> We're seeing what looks like the real-world version of this in production — details in #771 . Might be worth a case closer to that (more hops, or with add_record_route enabled per #642 ) to actually catch it going forward.
> 
> cc @alexlivekit @hechen-eng — thanks for the original fix, hope this helps get it re-applied!

@amr-imad thanks for raising this. It's on my to-do list 👍

PR #651, opened 2026-04-12
chore: pin GHA by commit, switch to golangci-lint

davidzhao:
also fixed lint issues


pkg/sip/silence_filler_test.go
```diff
@@ -299,7 +299,7 @@ func TestSilenceSuppressionDifferentCodecs(t *testing.T) {
 			// Create a gap of 2 frames by incrementing timestamp
 			missingFrameCount := 2
 			nextTimestamp += uint32(missingFrameCount) * tsPerFrame
-			nextSeq, nextTimestamp, err = tester.SendSignalFrames(5, nextSeq, nextTimestamp)
+			_, _, err = tester.SendSignalFrames(5, nextSeq, nextTimestamp)
```
dennwc: Looks like a false positive

pkg/sip/inbound.go
```diff
@@ -1772,7 +1772,7 @@ func (c *sipInbound) OwnSDP() []byte {
 }
 
 func (c *sipInbound) Accept(ctx context.Context, sdpData []byte, headers map[string]string) error {
-	ctx, span := Tracer.Start(ctx, "sip.inbound.Accept")
+	_, span := Tracer.Start(ctx, "sip.inbound.Accept")
```
dennwc: It's better not to change these. Context should be replaced when creating a new span. And someone will definitely forget to undo this change when passing context to the next function.

dennwc (changes requested):

pkg/sip/silence_filler_test.go
davidzhao: hmm how so?

pkg/sip/inbound.go
davidzhao: that's fair.. this is flagged by the CI because the modified value wasn't used.. I suppose we could disable that rule

dennwc (approved):

PR #653, opened 2026-04-14
TEL-464: reduce resampling in krisp

hechen-eng:


pkg/sip/server.go
```diff
@@ -120,7 +120,7 @@ type CallIdentifier struct {
 type Handler interface {
 	GetAuthCredentials(ctx context.Context, call *rpc.SIPCall) (AuthInfo, error)
 	DispatchCall(ctx context.Context, info *CallInfo) CallDispatch
-	GetMediaProcessor(features []livekit.SIPFeature, featureFlags map[string]string, callID string) msdk.PCM16Processor
+	GetMediaProcessor(features []livekit.SIPFeature, featureFlags map[string]string, callID string, inputSampleRate int) msdk.PCM16Processor
```
dennwc, line 123: I think we should stop introducing new arguments here and instead gather optional ones in a single `struct`.

In this case, the only required one would be the new  `inputSampleRate`, and everything else would be in a new `MediaProcessorOpts` struct.

hechen-eng: ~aren't all of them required?~ ~or all of them can be optional and we can handle missing parameters. `inputSampleRate` is only required for krisp, so technically it's optional.~

~I don't see much value of putting all parameters in a struct though, there are only four of them..~

I misunderstood the concern earlier. Now it's fixed with `MediaProcessorOpts` 👍

alexlivekit (approved):
Nice.

I might have combined the call id, sample rate, (and possibly others) into something like a call descriptor, but this is not a showstopper.

PR #655, opened 2026-04-16
add sip.verbose_call_logging feature flag

nishadmusthafa:
Move "changing media source/destination" and "call statistics" logs from Info to Debug by default. When the sip.verbose_call_logging project flag is enabled, these logs are emitted at Info level. Also move "BYE for non-existent call" to Debug unconditionally.

### DIFF
--- a/pkg/sip/client.go
+++ b/pkg/sip/client.go
@@ -345,7 +345,7 @@ func (c *Client) onBye(req *sip.Request, tx sip.ServerTransaction) bool {
 	c.cmu.Unlock()
 	if call == nil {
 		if tag != "" {
-			c.log.Infow("BYE for non-existent call", "sipTag", tag)
+			c.log.Debugw("BYE for non-existent call", "sipTag", tag)
 		}
 		_ = tx.Respond(sip.NewResponseFromRequest(req, sip.StatusCallTransactionDoesNotExists, "Call does not exist", nil))
 		return false
--- a/pkg/sip/media.go
+++ b/pkg/sip/media.go
@@ -113,17 +113,22 @@ func (s *Stats) Load() StatsSnapshot {
 	}
 }
 
-func (s *Stats) Log(log logger.Logger, callStart time.Time) {
+func (s *Stats) Log(log logger.Logger, callStart time.Time, verbose bool) {
 	const expectedSampleRate = RoomSampleRate
 	st := s.Load()
-	log.Infow("call statistics",
+	args := []any{
 		"stats", st,
 		"durMin", int(time.Since(callStart).Minutes()),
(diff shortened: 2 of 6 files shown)


hechen-eng:
it'll hurt debuggability without call statistics, especially for audio quality related issues. Shall we at least log at the info at the end of the call?

nishadmusthafa:
The log lines that I converted to debug with the feature flag, how often do we use that for debugging ?

hechen-eng:
> The log lines that I converted to debug with the feature flag, how often do we use that for debugging ?

Depends on how often we see audio quality issues. Also, there are media latency metrics in it. We still haven't enabled jitter buffer and call statistics will be needed when we do that

pkg/sip/client.go
```diff
@@ -345,7 +345,7 @@ func (c *Client) onBye(req *sip.Request, tx sip.ServerTransaction) bool {
 	c.cmu.Unlock()
 	if call == nil {
 		if tag != "" {
-			c.log.Infow("BYE for non-existent call", "sipTag", tag)
```
alexlivekit, line 348: This was already removed, we're double logging as it is.

pkg/sip/media.go
```diff
@@ -113,17 +113,22 @@ func (s *Stats) Load() StatsSnapshot {
...
+	args := []any{
 		"stats", st,
 		"durMin", int(time.Since(callStart).Minutes()),
 		"sip_rx_ppm", ratePPM(st.Port.AudioRX, expectedSampleRate),
 		"sip_tx_ppm", ratePPM(st.Port.AudioTX, expectedSampleRate),
 		"lk_publish_ppm", ratePPM(st.Room.PublishTX, expectedSampleRate),
 		"expected_pcm_hz", expectedSampleRate,
-	)
+	}
+	if verbose {
```
alexlivekit, line 127:
```suggestion
	if verbose || st.Closed {
```
Please keep the one-per call, on termination print. This is invaluable.

For the rest, I'm with He. It will be next-to-impossible to find stuck calls after this change. I actually rely on this one periodic print quite a lot. How do you find stuck calls today?

pkg/sip/media_port.go
```diff
@@ -249,6 +250,14 @@ func (c *udpConn) Read(b []byte) (n int, err error) {
 	return n, err
 }
 
+func (c *udpConn) logMediaChange(msg string, args ...any) {
```
alexlivekit, line 253: Not sure how to feel about this either. We should fix the underlying cause imo, and then suppress this.

nishadmusthafa:
@alexlivekit @hechen-eng Agreed on the alternative approach with 
1. reducing spam calls causing large number of logs
2. eventually use cloud-observability to reduce some of the logs
3. aggregating logs in multiple places under one log so we can still debug. Closing this one out.

PR #662, opened 2026-04-23
TEL-535: bad SDP should result in 400 error

hechen-eng:


pkg/sip/inbound.go
```diff
@@ -991,6 +997,11 @@ func (c *inboundCall) runMediaConn(tid traceid.ID, offerData []byte, enc livekit
 		return nil, err
 	}
 
+	offer, err := sdp.ParseOffer(offerData)
+	if err != nil {
+		return nil, SDPOfferParseError{Err: err}
```
alexlivekit, line 1002: NIT: Use the existing `SDPError`.

alexlivekit (approved):
Other than not needing to create a new error class, looks good!

pkg/sip/inbound.go
hechen-eng: we currently treat `SDPError` as internal error...

alexlivekit: Hmm. In what use case would a 5xx be accurate when receiving an SDPError?

alexlivekit: Yeah, today we get SDPError only if:
1. There's no audio
2. There are no matching crypto settings (either to local config or to None/Allow/Require per-call configuration)
3. Parsing issue
IMO all these are 4xx rather than 5xx. We should be able to safely reuse this.

PR #664, opened 2026-04-23
TEL-535: Clasisfying sdp errors as 4xx rather than 5xx

alexlivekit:


pkg/sip/client.go
```diff
@@ -344,10 +344,6 @@ func (c *Client) onBye(req *sip.Request, tx sip.ServerTransaction) bool {
 	call := c.activeCalls[tag]
 	c.cmu.Unlock()
 	if call == nil {
-		if tag != "" {
-			c.log.Infow("BYE for non-existent call", "sipTag", tag)
-		}
-		_ = tx.Respond(sip.NewResponseFromRequest(req, sip.StatusCallTransactionDoesNotExists, "Call does not exist", nil))
 		return false
```
alexlivekit, line 351: Unrelated: When we `return false` here, [the caller](https://github.com/livekit/sip/blob/e5fedb46a2511e74383db3ae6cc1b00ac927722b/pkg/sip/inbound.go#L535-L541) also does this exact flow.

hechen-eng: The client also receives BYE. Shouldn't we respond `StatusCallTransactionDoesNotExists` if the call doesn't exist 🤔

hechen-eng (approved):

pkg/sip/client.go
alexlivekit: Both do respond with `StatusCallTransactionDoesNotExists` today. Not sure I follow

PR #667, opened 2026-04-28
TEL-455: instrument call termination

hechen-eng:
We don't have a clear way to measure SIP call reliability today. There is no call completion rate metric. This change classifies every termination at the source — the close site is where someone has the context to decide whether a failure is our fault, the customer's fault, or a successful completion. After this change, the SLI is a single metric query with a stable label, and adding a new termination path forces a classification decision via the constructor functions (Success / ServerError / ClientError) rather than a free-form string.

Changes
- New type `stats.Termination{Result, Reason}` with `Success/ServerError/ClientError` constructors. Every termination must pick one.
- `livekit_sip_calls_terminated` counter gains a `result` label (`success | server_error | client_error`) alongside existing `dir/to/reason`.
- Pre-INVITE rejections (`auth-error, flood, no-rule, quota-exceeded, no-trunk, unauthorized`) now also write to `calls_terminated` via `InviteErrorShort`, so the SLI denominator is a single metric. Legacy `invite_error` counter retained for back-compatability.
- `inboundCall.close` / `outboundCall.close` signatures take `Termination` instead of a bare `reason` string. `error bool` dropped from inbound, where `Result == ServerError` covers it.

### DIFF
--- a/pkg/sip/client.go
+++ b/pkg/sip/client.go
@@ -349,7 +349,7 @@ func (c *Client) onBye(req *sip.Request, tx sip.ServerTransaction) bool {
 	call.log.Infow("BYE from remote")
 	go func(call *outboundCall) {
 		call.cc.AcceptBye(req, tx)
-		call.CloseWithReason(ctx, CallHangup, "bye", livekit.DisconnectReason_CLIENT_INITIATED)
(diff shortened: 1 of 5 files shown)


alexlivekit (approved):
nice!

PR #668, opened 2026-04-28
TEL-455: instrument inbound call termination due to shutdown

hechen-eng:


alexlivekit (commented):
wouldn't this cause all active calls to be immediately dropped whenever a pod wants to go down?

hechen-eng:
> wouldn't this cause all active calls to be immediately dropped whenever a pod wants to go down?

when a pod wants to go down, it'll go through draining process, i.e. wait for all active calls to finish. If the pod is killed, e.g. `syscall.SIGINT`, it'll shutdown. This is the same behavior as is today. This PR merely tracks the impacted calls when a pod is killed 👍

alexlivekit:
> when a pod wants to go down, it'll go through draining process, i.e. wait for all active calls to finish. If the pod is killed, e.g. syscall.SIGINT, it'll shutdown.

How exactly does it start draining? 
IIRC, a pod would receive ~~SIGINT~~ SIGTERM to enter the terminating state. That, in our system (and in k8s in general) is considered "draining". From there on out it has ~3 hours (our configured grace period) to wait for calls to terminate. If after this time the pod is still up, it'll get killed by `SIGKILL`, which simply force-kills it.
Sounds like I'm missing something here?

> This is the same behavior as is today.

So I am confused. Today, this piece of code calls `sipInbound.Close()`, which does `sipInbound.cancel()`.
There are two places that listen to this context terminnating:
- `TransferCall()` for inbound calls, I'm guessing this wasn't your primary target.
- `inboundCall.handleInvite()`, but this is only for calls that are being established.
So the code today,  AFAICT, does not really affect calls that are already running.

The PR is calling `sipInbound.close()`, which will begin teardown immediately. So if I have this right, not quite the same thing.

hechen-eng:
> How exactly does it start draining?

either controlled by "restart_after" configuration or `SIGTERM` or `SIGQUIT`.

> The PR is calling sipInbound.close(), which will begin teardown immediately. So if I have this right, not quite the same thing.

With `Close()` -> `cancel()`, it will closeWithHangup. With `Shutdown()`, it'll `close(..., "shutdown")`. So both will impact calls that are already running?

alexlivekit:
> IIRC, a pod would receive ~~SIGINT~~ SIGTERM to enter the terminating state. ...

True, but after the edit irrelevant. This code only runs after the system has drained following a SIGTERM, which would normally only be issued locally and in our system means a hard-stop.

So, all good I think.

alexlivekit (approved):

PR #670, opened 2026-04-29
TEL-553: fix noisy srtp error log

hechen-eng:
When a SIP call ends, MediaPort.Close() closes the underlying UDP conn, and the read goroutine wakes up with net.ErrClosed ("use of closed network connection"), which is a perfectly normal teardown, but logged as ERROR.

### DIFF
--- a/pkg/sip/srtp_conn.go
+++ b/pkg/sip/srtp_conn.go
@@ -1,0 +1,47 @@
+// Copyright 2026 LiveKit, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


pkg/sip/srtp_conn.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"errors"
+	"io"
+	"net"
+	"sync/atomic"
+)
+
+// srtpConn wraps the UDP conn passed to a pion SRTP session and translates
+// net.ErrClosed from Read into io.EOF after Close has been called. pion/srtp's
+// read goroutine logs every non-EOF read error at ERROR level (see pion/srtp/v3
```
dennwc, line 26: Maybe we could fix upstream instead? Or could keep it as-is for now.

dennwc (approved):

pkg/sip/srtp_conn.go
hechen-eng: not sure about fixing it upstream since `net.ErrClosed` could an error path. It's just that it's a valid case in our system

PR #672, opened 2026-05-04
Allow specifying codecs for the SDP.

dennwc:
Allow specifying codec list for the SDP. All codecs supported by our media SDK can be enabled, even if they are not enabled by default.

Disabling codecs works by setting `OnlyListedCodecs` in the `SIPMediaConfig` and specifying the exact codec set. It will override any enable/disable flags we set. 

For now there are no new supported codecs. The first planned additions are OPUS and AMR-WB which will follow shortly in separate PRs.

Requires https://github.com/livekit/protocol/pull/1530


hechen-eng (approved):
Alt Text

PR #674, opened 2026-05-05
Tagging outbound logs with sipCallID

alexlivekit:


pkg/sip/outbound.go
```diff
@@ -108,6 +108,12 @@ func (c *Client) newCall(ctx context.Context, tid traceid.ID, conf *config.Confi
 	if sipConf.host == "" {
 		sipConf.host = contact.GetHost()
 	}
+	fromURI := URI{
+		User:      sipConf.from,
+		Host:      sipConf.host,
+		Addr:      contact.Addr,
+		Transport: tr,
+	}
```
alexlivekit, line 116: Moved, as-is, from function call

pkg/sip/outbound.go
```diff
@@ -175,6 +166,17 @@ func (c *Client) newCall(ctx context.Context, tid traceid.ID, conf *config.Confi
...
+func (c *outboundCall) setAttrsToHeaders(headers map[string]string) map[string]string {
+	if len(c.sipConf.attrsToHeaders) == 0 {
+		return headers
+	}
+	r := c.lkRoom.Room()
+	if r == nil {
+		return headers
+	}
+	return AttrsToHeaders(r.LocalParticipant.Attributes(), c.sipConf.attrsToHeaders, headers)
+}
```
alexlivekit, line 178: Moved, as-is, from function call

pkg/sip/outbound.go
```diff
@@ -779,6 +781,7 @@ func (c *Client) newOutbound(log logger.Logger, id LocalTag, from, contact URI,
 		log:        log,
 		c:          c,
 		id:         id,
+		callID:     guid.HashedID(string(id)),
```
alexlivekit, line 784: Note: We previously did
```go
c.callID = guid.HashedID(fmt.Sprintf("%s-%s", string(c.id), toHeader.Address.String()))
```
However, given `c.id` which is our local tag is already basically guaranteed to be unique per call attempt, I've found that to be a bit of an overkill.

pkg/sip/outbound.go
```diff
@@ -121,23 +127,8 @@ func (c *Client) newCall(ctx context.Context, tid traceid.ID, conf *config.Confi
...
-			return headers
-		}
-		r := c.lkRoom.Room()
-		if r == nil {
-			return headers
-		}
-		return AttrsToHeaders(r.LocalParticipant.Attributes(), c.sipConf.attrsToHeaders, headers)
-	})
+	call.cc = c.newOutbound(log, id, fromURI, contact, sipConf.displayName, call.setAttrsToHeaders)
+	call.log = call.log.WithValues("jitterBuf", call.jitterBuf, "sipCallID", call.cc.callID)
```
hechen-eng, line 131: thanks for doing this!

hechen-eng (approved):

PR #675, opened 2026-05-05
TEL-562: make media timeout detection more deterministic

hechen-eng:


pkg/sip/media_port.go
```diff
@@ -366,7 +372,7 @@ func NewMediaPortWith(tid traceid.ID, log logger.Logger, mon *stats.CallMonitor,
 	}
 	p.timeoutInitial.Store(&opts.MediaTimeoutInitial)
 	p.timeoutGeneral.Store(&opts.MediaTimeout)
-	go p.timeoutLoop(tid, func() {
```
dennwc, line 369: Please keep the `tid`. It helps find stuck media loops for specific calls.

pkg/sip/media_port.go
```diff
@@ -315,6 +316,9 @@ type MediaOptions struct {
 	NoInputResample     bool
 	IgnorePreanswerData bool
 	LogSignalChanges    bool
+
+	// timeoutCheckInterval overrides the timeout polling interval.
+	timeoutCheckInterval time.Duration
```
dennwc: Any reason to make it unexported?

pkg/sip/media_port.go
```diff
@@ -458,101 +460,79 @@ func (p *MediaPort) SetTimeout(initial, general time.Duration) {
...
+					}
 				}
 			}
 
-			// Ticker is allowed to fire earlier than the full timeout interval. Skip if it's not a full timeout yet.
-			if since+timeout/10 < timeout {
-				if verbose {
-					log.Infow("too early to trigger", "since", since, "timeout", timeout)
+			if verbose {
+				fields := []any{
```
dennwc:
```
log := log.With(...)
```
It will define the `log` in the `if` scope, so you can use it on all the cases below.

pkg/sip/media_port.go
```diff
@@ -458,101 +460,79 @@ func (p *MediaPort) SetTimeout(initial, general time.Duration) {
...
 				verbose = true
 				lastLog = now
-				log = log.WithValues(
-					"startPackets", startPackets,
-					"packets", curPackets,
-					"lastPackets", lastPackets,
-					"sinceLast", time.Since(lastTime),
-					"sinceStart", time.Since(startTime),
-				)
-				if curPackets == startPackets {
```
dennwc, line 501: Better keep this one, it helped find quite a few random deadlocks on the media path.

pkg/sip/media_port.go
hechen-eng: I don't see `tid` is actually used in the old code. Do we still need it 🤔

pkg/sip/media_port.go
hechen-eng: changed to exported to be consistent with the rest of the code 👍

pkg/sip/media_port.go
hechen-eng: done 👍

pkg/sip/media_port.go
hechen-eng: we now check whether it's `isInitial` using `lastPacketTime.Before(startTime)`, which should be sufficient? logs are also preserved from line 520 to 525

hechen-eng:
made the change to reset the timer by comparing with lastPacketTime 👍

alexlivekit (approved):
nice!

PR #677, opened 2026-05-06
Consolidating track-related logs

alexlivekit:


alexlivekit (commented):
- Added subscribedAt attribute to per-track logging.
- Added closedAt attribute for the "track closed" message.
- Dropped "mixing track". We can now get this info from "track closed". Forthermore, we already have the library-emitted "track subscribed", which we should ideally drop as well.
- Left the OnTrackSubscribed as a stack function for the sake of review, but this can now be easily moved to a dedicated function if we wanted to.

hechen-eng (approved):
rebase and fix the CI?

PR #680, opened 2026-05-08
TEL-575: do not transition state for 407

hechen-eng:


pkg/sip/inbound.go
```diff
@@ -457,12 +463,19 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
...
 			s.cmu.Lock()
 			s.provisionalInvites.Add([2]string{cc.SIPCallID(), string(cc.Tag())}, cc.ID())
 			s.cmu.Unlock()
 			cmon.InviteErrorShort(stats.ClientError("unauthorized"))
+			if challenge {
+				// We just sent the initial 407 challenge; the client is expected to
+				// retry with digest credentials. This is part of the normal SIP auth
+				// handshake, not a finalized error, so suppress the deferred state
+				// transition to avoid recording a 0-duration call.
+				state = nil
```
nishadmusthafa, line 505: I think this is fine for now. But if we don't get the second invite, as its designed currently, we won't be logging the auth failure or showing it in the dashboard.. right ? Any way around that ?

hechen-eng: hmm, I thought that's part of the long-term plan? We don't need to log since sipfe will log the 407 response. 

With current short-term fix, if there is a second invite, we are all good. If not, we'll have a hanging INCOMING record. As far as sip, 407 is a final response, the SIP transaction is complete from the server's perspective. Are we okay with this for now?

nishadmusthafa: The long term was more about logging an event for the first auth challenge. We still need to handle the case of a second invite not coming. If 407 is the final response in that case, how will billing know to end the call ?

hechen-eng: from billing's perspective, the call never started since the call is never active, so we should be fine in terms of billing.

It's bad user experience though (a hanging INCOMING call status). Maybe we can use a timeout to guard this 🤔

nishadmusthafa: So we send them the UpdateSIPCallStateRequest with callStatus set to `SCS_INCOMING` before all this happens don't we ? If that's not the case, then we are good.

nishadmusthafa: If we are sending SCS_INCOMING, we should also send either SCS_DISCONNECT or SCS_ERROR. As you suggested a timeout for the second invite might be a good approach then

pkg/sip/inbound.go
```diff
@@ -158,7 +159,31 @@ func (s *Server) getInvite(sipCallID string) *inProgressInvite {
...
+	time.AfterFunc(authChallengeTimeout, func() {
+		if !i.authResolved.CompareAndSwap(false, true) {
+			return
+		}
+		log.Infow("auth challenge timed out without authenticated retry; finalizing call as error",
+			"sipCallID", i.sipCallID, "timeout", authChallengeTimeout)
+		st.Update(context.Background(), func(info *livekit.SIPCallInfo) {
+			info.CallStatus = livekit.SIPCallStatus_SCS_ERROR
+			info.Error = "auth challenge issued, no authenticated retry received"
+			info.EndedAtNs = time.Now().UnixNano()
```
nishadmusthafa: Should `EndedAtNs` be `now()` minus the timeout ?

hechen-eng: I think it should be now to be consistent with the error message `"auth challenge issued, no authenticated retry received"`. There shouldn't be any billing impact because the call was never active 🤔

nishadmusthafa: Oh I just meant the accuracy of the EndedAtNs timestamp. Technically, the 407 was sent 30 seconds prior to this and that is when the call ended right as opposed to now()

hechen-eng: cool, fixed!

nishadmusthafa (approved):
Thank you @hechen-eng !

PR #684, opened 2026-05-12
Eliminate CANCEL race, fix flaky test

alexlivekit:


pkg/sip/inbound.go
```diff
@@ -1271,8 +1271,8 @@ func (c *inboundCall) close(ctx context.Context, status CallStatus, t stats.Term
 	// See: https://github.com/livekit/sip/issues/404
 	c.cc.CloseWithStatus(ctx, sipCode, sipStatus)
 	c.closeMedia()
-	if c.callDur != nil {
-		c.callDur()
```
alexlivekit, line 1275: In a certain edge case this would race. Made sure we're only reading once. No lock needed.

pkg/sip/inbound.go
```diff
@@ -1324,7 +1324,7 @@ func (c *inboundCall) closeWithCancelled(ctx context.Context) {
 	if p := c.closeReason.Load(); p != nil {
 		reason = *p
 	}
-	c.closeWithReason(ctx, CallHangup, stats.Success("cancelled"), reason)
```
alexlivekit, line 1327: Added a dedicated `CallCancelled` status here, same as `CallHangup` in all respects other than the resulting sip status - now 487 instead of 486.

pkg/sip/inbound.go
```diff
@@ -1795,9 +1795,8 @@ func (c *sipInbound) StartRinging() {
 			case <-stop:
 				return
 			case r := <-cancels:
-				close(c.cancelled)
+				close(c.cancelled) // Other goroutines will respond to the primary INVITE
 				_ = tx.Respond(sip.NewResponseFromRequest(r, sip.StatusOK, "OK", nil))
-				c.RespondAndDrop(sip.StatusRequestTerminated, "Request Terminated")
```
alexlivekit, line 1800: This was the thing responding 9/10 times, and it used the correct code. But it would race with `close(c.cancelled)` that would result in other goroutines calling `closeWithCancelled()` that would do cleanup and then respond with 486 which is a little much imo.

Elected to only respond to the CANCEL here, so that we don't have to split cleanup and INVITE response.

hechen-eng (approved):
LGTM 👍

PR #687, opened 2026-05-13
Allow ignoring local IP in SDP and enable symmetric mode

### LINKED ISSUE #611: Feature Request: When an INVITE includes a private IP address in the SDP, send the response to the source IP

dennwc:
Automatically enable symmetric RTP if the other peer send a local IP in their SDP. Fixes #611

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -106,10 +106,11 @@ type Config struct {
 	MediaUseExternalIP bool   `yaml:"media_use_external_ip"`
 	MediaNAT1To1IP     string `yaml:"media_nat_1_to_1_ip"`
 
-	MediaTimeout        time.Duration   `yaml:"media_timeout"`
-	MediaTimeoutInitial time.Duration   `yaml:"media_timeout_initial"`
-	SymmetricRTP        bool            `yaml:"symmetric_rtp"`
-	Codecs              map[string]bool `yaml:"codecs"`
+	MediaTimeout         time.Duration   `yaml:"media_timeout"`
+	MediaTimeoutInitial  time.Duration   `yaml:"media_timeout_initial"`
+	SymmetricRTP         bool            `yaml:"symmetric_rtp"`
+	IgnoreLocalAddrInSDP bool            `yaml:"ignore_local_addr_in_sdp"` // enable symmetric RTP if local IP is specified in SDP
+	Codecs               map[string]bool `yaml:"codecs"`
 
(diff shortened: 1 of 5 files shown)


pkg/config/config.go
```diff
@@ -101,10 +101,11 @@ type Config struct {
...
 	MediaNAT1To1IP     string `yaml:"media_nat_1_to_1_ip"`
 
-	MediaTimeout        time.Duration   `yaml:"media_timeout"`
-	MediaTimeoutInitial time.Duration   `yaml:"media_timeout_initial"`
-	SymmetricRTP        bool            `yaml:"symmetric_rtp"`
-	Codecs              map[string]bool `yaml:"codecs"`
+	MediaTimeout         time.Duration   `yaml:"media_timeout"`
+	MediaTimeoutInitial  time.Duration   `yaml:"media_timeout_initial"`
+	SymmetricRTP         bool            `yaml:"symmetric_rtp"`
+	IgnoreLocalAddrInSDP bool            `yaml:"ignore_local_addr_in_sdp"` // enable symmetric RTP if local IP is specified in SDP
```
hechen-eng, line 112: the comment is a bit confusing.. `ignore_local_addr_in_sdp` indicates ignoring local addr when figuring out whether symmetric RTP should be enabled?

hechen-eng (approved):

pkg/config/config.go
dennwc: It is exactly what it says though - if we see local IP in SDP, we will ignore it and send traffic to the address we receive RTP from (aka symmetric RTP).

PR #692, opened 2026-05-19
TEL-588: better categorization of invite-failed for outbound calls

hechen-eng:
A lot of invite-failed are client errors.

### DIFF
--- a/pkg/sip/errors.go
+++ b/pkg/sip/errors.go
@@ -6,6 +6,9 @@ import (
 	"fmt"
 	"net"
 
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
 	"github.com/livekit/media-sdk/sdp"
 	"github.com/livekit/protocol/livekit"
 	"github.com/livekit/psrpc"
@@ -42,12 +45,18 @@ var _ inviteClassifier = SDPError{}
 func (e SDPError) Error() string { return e.Err.Error() }
 func (e SDPError) Unwrap() error { return e.Err }
 
+// GRPCStatus lets psrpc.GetErrorCode (and any gRPC-aware caller) extract the
+// failed-precondition code without manual psrpc.NewError wrapping.
+func (e SDPError) GRPCStatus() *status.Status {
+	return status.New(codes.FailedPrecondition, e.Err.Error())
+}
+
 func (e SDPError) ClassifyInvite() inviteFailure {
 	res := inviteFailure{
 		status:    callRejected,
 		reason:    livekit.DisconnectReason_MEDIA_FAILURE,
 		reportErr: e.Err,
-		returnErr: psrpc.NewError(psrpc.FailedPrecondition, e.Err),
+		returnErr: e,
 	}
 	switch {
 	case errors.Is(e.Err, sdp.ErrNoCommonMedia):
--- a/pkg/sip/errors.go
+++ b/pkg/sip/errors.go
@@ -2,10 +2,9 @@ package sip
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
-
-	"github.com/pkg/errors"
 
 	"github.com/livekit/media-sdk/sdp"
 	"github.com/livekit/protocol/livekit"
@@ -69,8 +68,7 @@ func classifyInviteError(err error) inviteErrorClassification {
 		returnErr: err,
 	}
 
-	var sipStatus *livekit.SIPStatus
-	if errors.As(err, &sipStatus) {
+	if sipStatus, ok := errors.AsType[*livekit.SIPStatus](err); ok {
 		code := int(sipStatus.Code)
 		switch code {
 		case int(sip.StatusUnauthorized), int(sip.StatusProxyAuthRequired):
@@ -110,8 +108,7 @@ func classifyInviteError(err error) inviteErrorClassification {
 		return res
 	}
 
-	var sdpErr SDPError
-	if errors.As(err, &sdpErr) {
+	if sdpErr, ok := errors.AsType[SDPError](err); ok {
 		res.status, res.reason = callRejected, livekit.DisconnectReason_MEDIA_FAILURE
 		res.reportErr = nil
 		res.returnErr = psrpc.NewError(psrpc.FailedPrecondition, sdpErr.Err)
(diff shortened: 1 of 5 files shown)


pkg/sip/errors.go
```diff
@@ -11,3 +26,152 @@ func (e SDPError) Error() string {
...
+	if errors.Is(err, context.Canceled) {
+		res.status, res.term, res.reason = callRejected, stats.ClientError("canceled"), livekit.DisconnectReason_USER_UNAVAILABLE
+		res.reportErr = nil
+		res.returnErr = psrpc.NewError(psrpc.Canceled, err)
+		return res
+	}
+
+	// Specific net error types before *net.OpError (which wraps them).
+	var addrErr *net.AddrError
+	if errors.As(err, &addrErr) {
```
dennwc: Prefer using `errors.AsType`, it's easier to get right and doesn't require a temp variable.

pkg/sip/errors.go
```diff
@@ -11,3 +26,152 @@ func (e SDPError) Error() string {
 func (e SDPError) Unwrap() error {
 	return e.Err
 }
+
+// inviteErrorClassification captures everything connectSIP needs after a
+// dialSIP failure: how to record the call, how to bucket the SLI, and which
+// error (if any) to surface back to the caller and the call record.
+type inviteErrorClassification struct {
```
dennwc: Looks good! However I was hoping we could consolidate this a bit.

Assuming we are not considering generic SIP events, we usually get some error that happens first. So we usually don't have to populate all of these fields at the same time, and the type of error itself could determine how it should be interpreted in a given context.

If we get `CallStatus`, that type itself (or a small wrapper over it) could implement a method that returns associated disconnect reason and reported error.

If we get a room disconnect afterwards, that's a `DisconnectReason`, which we can wrap into something that also gets us extra info using optional methods. For example, we might want to send a `BYE` with a particular reason after a room disconnect. This is one extra method. Everything else that doesn't implement that method will get a default reason. 

Having this in mind, could you try to come up with an interface that can then expand into this set of info? It could extend `error` and introduce extra methods that convert that error to some specific status/reason in a given context.

Unfortunately I don't have a specific example that does that, except maybe how SIPStatus [implements](https://github.com/livekit/protocol/blob/main/livekit/sip.go#L26) methods for converting to both Twirp and gRPC errors. You may want to look into how `errors.As` works internally, as it allows you to chain a few errors from different contexts and then getting the right representation at the call site.

dennwc (approved):

pkg/sip/errors.go
hechen-eng: just to double confirm I understand this correctly, so.. we want an interface-based approach, with each error type knows how to interpret itself, implementing optional methods that return the disconnect reason, termination, status, etc, right? but `sip` doesn't control  `*livekit.SIPStatus`, `*net.OpError`, `*net.AddrError`, `*net.DNSError`, plus `context.Canceled/DeadlineExceeded`, so we can't hang methods on those directly, and we would still need the classifier

hechen-eng: added something here https://github.com/livekit/sip/pull/692/changes/953f89d65f5d5f149d197021bbac2f58aeaba50f, but I don't feel like it's simplifying things. Let me know if this aligns with your thinking or we should revert that commit.

pkg/sip/errors.go
```diff
@@ -1,13 +1,163 @@
...
 type SDPError struct {
 	Err error
 }
 
-func (e SDPError) Error() string {
-	return e.Err.Error()
+func (e SDPError) Error() string { return e.Err.Error() }
+func (e SDPError) Unwrap() error { return e.Err }
+
+func (e SDPError) ClassifyInvite() inviteFailure {
```
dennwc, line 54: To ensure that a specific error implements the interface you could add static compile assertion:
```
var _ inviteClassifier = SDPError{}
```
This way it will fail if the interface changes.

pkg/sip/errors.go
dennwc: This looks better to me, as it at least splits the large error classification function into smaller code blocks related to a particular error class.

One extra thing we could do potentially, is to split the other things into methods too. For example, `Status` could be a separate optional method. If errors does not implement this interface (`SIPStatus() livekit.SIPStatus`), the call site will assume a default (e.g. `callRejected`). So most errors won't have to have the code for it at all.

But that's a larger change, I assume.

pkg/sip/errors.go
```diff
@@ -1,13 +1,163 @@
...
+	res := inviteFailure{
+		status:    callRejected,
+		reason:    livekit.DisconnectReason_MEDIA_FAILURE,
+		reportErr: nil,
```
dennwc: Why are we not reporting this, btw?

pkg/sip/errors.go
```diff
@@ -1,13 +1,163 @@
...
+		returnErr: psrpc.NewError(psrpc.FailedPrecondition, e.Err),
```
dennwc: Hmm, his is kind of similar to the standard `GRPCStatus` interface. I think the error could just implement that or a PSRPC-specific one.

For the context, [here's how](https://github.com/grpc/grpc-go/blob/master/status/status.go#L100) gRPC checks all errors and gets their statuses. And [here's](https://github.com/livekit/psrpc/blob/main/errors.go#L101) the way how PSRPC checks for both it's own interface and the gRPC one.

pkg/sip/errors.go
```diff
@@ -1,13 +1,163 @@
...
+	returnErr error
+}
+
+// inviteClassifier lets an error describe its own verdict instead of
+// classifyInviteError carrying a switch over every variant. Source sites
+// wrap raw errors in these types so the classification lives next to where
+// the failure happened.
+type inviteClassifier interface {
+	error
+	ClassifyInvite() inviteFailure
```
dennwc, line 36: If you expect other packages to implement this, we might need to export the `inviteFailure` type. But if not, you could unexport `ClassifyInvite()` instead.

dennwc (approved):

pkg/sip/errors.go
hechen-eng: it's existing behavior 🤔

pkg/sip/errors.go
hechen-eng: I'll keep it unexported for now 👍

pkg/sip/errors.go
hechen-eng: `livekit.DisconnectReason_MEDIA_FAILURE` is shown to the customer, which should be sufficient?

hechen-eng: after a second thought, I think it's beneficial for the customer to see the reportErr for SDPError, so that they can fix it by themselves. This is a behavior change though. `SCS_DISCONNECTED` -> `SCS_ERROR` after this change

pkg/sip/errors.go
hechen-eng: > But that's a larger change, I assume.

I'll postpone this. Maybe revisit this after there are more self-classifying types in the future 👍

pkg/sip/errors.go
hechen-eng: sounds good. Done!

PR #693, opened 2026-05-20
Support AMR-WB codec (optional).

dennwc:
Add support for AMR-WB codec. It can be enabled by specifying `MediaConfig` with `AMR-WB/16000` codec in the dispatch rule or `CreateSIPParticipant`.

### DIFF
--- a/pkg/sip/media_codecs.go
+++ b/pkg/sip/media_codecs.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	_ "github.com/livekit/media-sdk/all"
+	"github.com/livekit/media-sdk/amrwb"
 	"github.com/livekit/media-sdk/dtmf"
 	"github.com/livekit/media-sdk/g711"
 	"github.com/livekit/media-sdk/g722"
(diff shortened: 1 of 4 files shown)


hechen-eng:
The CI failure seems to suggest a missing resampling. I believe this needs to be fixed here https://github.com/livekit/media-sdk/blob/main/amrwb/amrwb.go#L58?

davidzhao (commented):
is the idea to enable this by default at some point?

dennwc:
@hechen-eng fixed, also improved the test. I had to swap the resampler for the rest duration, because it's output is non-deterministic (random seed that we do not control).

@davidzhao yes, we could enable it by default, but probably for specific providers only. The reason is that adding anything extra to SDP may push UDP INVITE above MTU and break some existing customers (if their network doesn't handle fragmented UDP well).

pkg/sip/media_codecs.go
```diff
@@ -79,7 +81,25 @@ func codecSet(m *livekit.SIPMediaConfig) (*msdk.CodecSet, error) {
...
+			return nil, errors.New("no codec name specified")
+		}
+		rate := codec.Rate
+		if rate == 0 {
+			// Set default rate
+			switch name {
+			case g711.ALawSDPName, g711.ULawSDPName:
+				rate = 8000
+			case g722.SDPName:
+				rate = 8000 // actually 16000, it's a know bug in the spec
```
hechen-eng, line 95: lol, a bug that becomes a feature 😂

hechen-eng (approved):
Alt Text

PR #694, opened 2026-05-21
TEL-392: cache the drop and respond to retried failed calls

hechen-eng:


dennwc (approved):
We might need a config option for it. LGTM otherwise.

hechen-eng:
> We might need a config option for it. LGTM otherwise.

done 👍

PR #699, opened 2026-05-25
Add participant RPC for getting remote headers.

dennwc:
Implement a room participant RPC method `lk.sip.GetRemoteHeaders` for synchronously getting the full list of remote headers (from `INVITE` for inbound and `200 OK` for outbound).

### DIFF
--- a/pkg/sip/rpc.go
+++ b/pkg/sip/rpc.go
@@ -1,0 +1,101 @@
+package sip
+
+import (
+	"encoding/json"
+	"strings"
+
+	lksdk "github.com/livekit/server-sdk-go/v2"
+)
+
+func registerRPCMethodJSON[Req any, Resp any](r RoomInterface, name string, fnc func(r Req) (Resp, error)) error {
+	return r.RegisterRPC(name, func(data lksdk.RpcInvocationData) (string, error) {
+		var req Req
+		if len(data.Payload) != 0 {
+			if err := json.Unmarshal([]byte(data.Payload), &req); err != nil {
+				return "", err
+			}
+		}
+		resp, err := fnc(req)
+		if err != nil {
+			return "", err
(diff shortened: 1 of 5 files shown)


pkg/sip/rpc.go
```diff
@@ -0,0 +1,101 @@
...
+	out := make(map[string]string)
+	for _, h := range cc.RemoteHeaders() {
+		name := strings.ToLower(h.Name())
+		if _, ok := getHeadersIgnore[name]; ok {
+			continue
+		}
+		if _, ok := skip[name]; ok {
+			continue
+		}
+		if _, ok := only[name]; !ok && only != nil {
```
alexlivekit, line 83: nit: useless `&& only != nil` happens after access to `only[name]`

pkg/sip/rpc.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+		out, err := json.Marshal(&resp)
+		return string(out), err
+	})
+}
+
+type getHeadersReq struct {
+	Include []string `json:"include,omitempty"` // list of headers to include
+	Exclude []string `json:"exclude,omitempty"` // list of headers to exclude
+}
```
alexlivekit, line 30: I'm assuming this has to be a list, right?

alexlivekit (approved):
overall lgtm!

pkg/sip/rpc.go
dennwc: That is intentional, I'm using the `only != nil` as a check that this map is populated and "does not exist" should mean "header must be skipped".

pkg/sip/rpc.go
dennwc: It's easier to specify that way, I think.

PR #700, opened 2026-05-25
Support PCMU and PCMA room codecs

dennwc:
Currently, LiveKit SIP can only subscribe to room tracks that use Opus. This change adds support for subscribing to PCMU or PCMA tracks. The published track will still use Opus.

### DIFF
--- a/pkg/sip/room.go
+++ b/pkg/sip/room.go
@@ -16,19 +16,20 @@ package sip
 
 import (
 	"context"
+	"errors"
 	"io"
 	"math"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
-
-	"errors"
 
 	"github.com/frostbyte73/core"
 	"github.com/pion/webrtc/v4"
 
 	msdk "github.com/livekit/media-sdk"
 	"github.com/livekit/media-sdk/dtmf"
+	"github.com/livekit/media-sdk/g711"
 	"github.com/livekit/media-sdk/jitter"
 	"github.com/livekit/media-sdk/rtp"
 	"github.com/livekit/protocol/livekit"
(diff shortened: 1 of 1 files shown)


alexlivekit (approved):

pkg/sip/room.go
```diff
@@ -361,14 +362,29 @@ func (r *Room) Connect(ctx context.Context, conf *config.Config, rconf RoomConfi
...
+						}
+						defer cw.Close()
+
+						rh = rtp.NewMediaStreamIn(cw)
+					case "pcmu":
+						cw := g711.DecodeULaw(out)
+						rh = rtp.NewMediaStreamIn(cw)
+					case "pcma":
+						cw := g711.DecodeALaw(out)
+						rh = rtp.NewMediaStreamIn(cw)
```
alexlivekit, line 383: Nit: This, `rh = rtp.NewMediaStreamIn(cw)`, can be taken out of the switch, right?

pkg/sip/room.go
alexlivekit, line 346: IIUC, this plugs directly to the mixer, without resampling. Where does resampling happen (now that we might need it)?

alexlivekit (changes requested):
Yeah, I think this may be broken w/r/t resampling.
We resample today on r.out, where we define `r.out` to be using `RoomSampleRate` (hard coded to 48k), then define the mixer, also expecting 48k.
Then, when defining new tracks, `r.mix.NewInput()` expect the sample rate to match.

Marking this as changes required for now. Please feel free to ignore if I'm missing something.

pkg/sip/room.go
dennwc: That won't work, I think. `rtp.NewMediaStreamIn` is generic and depends on the exact frame type.

pkg/sip/room.go
dennwc: This will work, actually. All these methods like `DecodeULaw` internally check if the destination sample rate is the same as the codec expects, and if not, it adds a resampler into the chain automatically. So the mixer will run at 48k, but all tracks that are PCMU will be resampled up to match.

alexlivekit: Ah, you're right - didn't notice that!
So the pipeline is going to be (from room to tel):
- sfu sending g711 codecs to sip
- sip decodes, resamples it to the hard-coded 48k, and sends it to the mixer.
- mixer does its job, then we resample to 8k.

I guess this is fine for an initial support, but I would like to see explicit callouts/TODOs to alleviate this.
My initial thought it we'd want to remove the hard-coded assumption 48k in mixer, and add a manual resampler gate in front of it (to catch everything, not just g711).

alexlivekit (approved):
lgtm, with the one ask to add some TODOs/comments that explicitly document the double resampling we'd be doing in these cases.

pkg/sip/room.go
dennwc: Note that this is not a useless work that we could easily remove. There might be multiple participants in the room, and only one emitting 8k, everyone else could still use 48k.

PR #702, opened 2026-05-26
Support specifying To/From SIP headers for outbound.

dennwc:


pkg/sip/outbound.go
```diff
@@ -104,15 +102,7 @@ func (c *Client) newCall(ctx context.Context, tid traceid.ID, conf *config.Confi
...
-	if sipConf.host == "" {
-		sipConf.host = contact.GetHost()
-	}
-	fromURI := URI{
-		User:      sipConf.from,
-		Host:      sipConf.host,
-		Addr:      contact.Addr,
-		Transport: tr,
-	}
+
```
hechen-eng, line 105:
```suggestion
```

pkg/sip/client.go
```diff
@@ -175,32 +180,173 @@ func (c *Client) getActiveCall(tag LocalTag) *outboundCall {
...
+		p.Host = addr
+	}
+	return p, nil
+}
+
+func buildRawURI(raw string, tr livekit.SIPTransport) (*sip.Uri, error) {
+	p := &sip.Uri{Scheme: "sip"}
+	setUriTransport(p, tr)
+	if err := esip.ParseUri(raw, p); err != nil {
+		return nil, psrpc.NewErrorf(psrpc.InvalidArgument, "invalid request URI")
```
hechen-eng: shall we return a plain `fmt.Errorf`? it'll be wrapped as psrpc error in the caller

pkg/sip/client.go
```diff
@@ -16,15 +16,20 @@ package sip
...
 	"net/netip"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/frostbyte73/core"
 	"golang.org/x/exp/maps"
 
+	esip "github.com/emiago/sipgo/sip"
```
hechen-eng, line 32: we are using both emiago sipgo and livekit sipgo. Do we plan to only rely on one of them?

pkg/sip/client.go
```diff
@@ -175,32 +180,173 @@ func (c *Client) getActiveCall(tag LocalTag) *outboundCall {
...
+	} else {
+		p.Host = addr
+	}
+	return p, nil
+}
+
+func buildRawURI(raw string, tr livekit.SIPTransport) (*sip.Uri, error) {
+	p := &sip.Uri{Scheme: "sip"}
+	setUriTransport(p, tr)
+	if err := esip.ParseUri(raw, p); err != nil {
```
hechen-eng, line 227: this seems to be a bug since it'll drop what's set in `setUriTransport`?

hechen-eng (commented):
it'll be great if we can add some tests..

pkg/sip/client.go
dennwc: LK fork just aliases all types/functions we use to upstream ones. The only reason I import upstream here is that we don't have alias for that function/type.

dennwc:
@hechen-eng should be good now. Added a couple of tests and fixed the transports issue.

hechen-eng (approved):

PR #705, opened 2026-05-27
Rate limiting change of source and destination prints

alexlivekit:


hechen-eng (approved):

pkg/sip/media_port.go
```diff
@@ -246,9 +257,15 @@ func (c *udpConn) Read(b []byte) (n int, err error) {
 	n, addr, err := c.ReadFromUDPAddrPort(b)
 	prev := c.src.Swap(&addr)
 	if prev == nil || !prev.IsValid() {
-		c.log.Infow("setting media source", "addr", addr.String())
+		c.log.Infow("setting media source", "prev", prev, "addr", addr.String())
 	} else if *prev != addr {
-		c.log.Infow("changing media source", "addr", addr.String())
```
hechen-eng, line 251: are we seeing a lot of logs for this?

alexlivekit: Yes. I have not managed to figure out why this happens, but we're essentially getting media from twp sources at the same time (for the same call), so we're printing this 100 times/sec in these cases for however long the call is.

PR #706, opened 2026-05-29
fix symmetric rtp

alexlivekit:
I think the [original implementation](https://github.com/livekit/sip/commit/afd7dd4edf05edfeb4a7553b8f5341ad88ee6c44) broken now that we [introduced IgnoreLocalAddrInSDP](https://github.com/livekit/sip/commit/bab77e1593b8018866b01a9afa05f2873f212da5). I don't see anywhere we actually use SymmetricRTP.

### DIFF
--- a/pkg/sip/media_port.go
+++ b/pkg/sip/media_port.go
@@ -714,7 +714,7 @@ func (p *MediaPort) SetConfig(c *MediaConf) error {
 		"srtp", crypto,
 	)
 
-	symmetric := p.opts.IgnoreLocalAddrInSDP && c.Remote.Addr().IsPrivate()
+	symmetric := p.opts.SymmetricRTP || (p.opts.IgnoreLocalAddrInSDP && c.Remote.Addr().IsPrivate())
 	p.port.SetDst(c.Remote)
 	if symmetric {
 		p.port.SetSymmetric(true)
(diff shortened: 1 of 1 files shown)


pkg/sip/media_port.go
```diff
@@ -739,9 +739,6 @@ func (p *MediaPort) SetConfig(c *MediaConf) error {
 
 	p.mu.Lock()
 	defer p.mu.Unlock()
-	if !symmetric {
-		p.port.SetDst(c.Remote)
-	}
```
alexlivekit, line 744: Is already being set by line 718 && 720.

dennwc: Maybe lets keep it? I think there was a reason why it was done this way. But I do not recall.

dennwc (approved):

pkg/sip/media_port.go
alexlivekit: When `symmetric == false`, `udpConn.SetDst()` is called from one other place - line 718 earlier in this function. The other call is in `udpConn.Read()`, but that only works when `symmetric` is on, so it's irrelevant here.

The only benefit of having this called here as well is if `c.Remote` may change, we'd be updating it here, and that can't happen because it's not stored anywhere outside the stack of this one goroutine (created just before calling `MediaPort.SetConfig()`)

alexlivekit: I think we might have had a reason to do this in the past, but now that reason is gone IIUC.

PR #711, opened 2026-06-05
TEL-682: use even RTP port

hechen-eng:
need to bump media-sdk after https://github.com/livekit/media-sdk/pull/64


hechen-eng:
will it be a concern that only half of the ports can be used? or should we do prefer even ports and fallback to odd ports 🤔

hechen-eng:
> will it be a concern that only half of the ports can be used? or should we do prefer even ports and fallback to odd ports 🤔

shouldn't be a problem

dennwc (approved):

PR #712, opened 2026-06-08
Add optional Opus codec support for SIP

kannanmaran:
## Status

**Ready for review — but opening to get maintainer direction on architecture
before a larger refactor.** Opus is **disabled by default** and changes nothing
when off, so this is safe to land as-is; however, if maintainers prefer the
alternatives below (especially moving the codec into `media-sdk`), I'm happy to
refactor and do not consider the current shape final.

## Open questions for maintainers

1. **Codec location:** Should Opus encode/decode and codec registration live in
   `livekit/media-sdk/opus` (alongside G711/G722/AMR-WB) instead of in
   `livekit/sip`? `media-sdk/opus` already exists for the WebRTC side and its
   decoder is already channel-adaptive; I kept this PR in `livekit/sip` to scope
   the first cut, but media-sdk may be the right home.
2. **Enablement:** Should Opus be enabled through the existing codec
   config / per-trunk codec selection (like AMR-WB) instead of the new
   `enable_opus` flag? (Today the per-trunk path can't name `opus/48000/2` and
   the global `codecs:` map doesn't reach the per-call set — see Q4.)
3. **Encoder-options API:** What shape do you prefer for bitrate / complexity /
   FEC configuration (e.g. a package-level setter in `media-sdk/opus`, an
   options field on the codec, or service-level config)?
4. **Codec-set plumbing:** Should this PR also fix the existing `GlobalCodecs`
   vs `defaultCodecs` inconsistency (the YAML `codecs:` map writes to
   `GlobalCodecs`, but offers are built from `defaultCodecs`), or should that be
   a separate change?
5. **Priority:** What default priority should Opus use relative to G722/G711?
   (Currently `10`, i.e. preferred over all legacy codecs when enabled.)
6. **Interop:** Which SIP stacks do you want validated before considering Opus
   GA (e.g. FreeSWITCH, Asterisk, Twilio/Telnyx, specific SBCs)?

## Summary

Adds optional Opus (`opus/48000/2`) support for SIP media, behind a new
`enable_opus` config flag. **Disabled by default** — when off, SDP offers and the
existing PCMU/PCMA/G722/DTMF behavior are completely unchanged.

When enabled, Opus is preferred **when answering an inbound offer** (selection is
priority-based). In outbound offers, codecs are currently listed static-first, so
a peer selecting by offer order may still pick G722/G711 (see Q5). Peers that
don't support Opus fall back transparently to G722, then G711.

SIP media is encoded/decoded as mono (advertised as `opus/48000/2` per RFC 7587;
the decoder adapts to the peer's actual channel count). Encoder
bitrate/complexity/FEC are configurable via an optional `opus:` block.

The LiveKit/WebRTC-side Opus path (`pkg/media/opus/opus.go`) is untouched — all
new SIP-side logic lives in a separate, gated file reachable only when Opus is
negotiated.

## Config

```yaml
enable_opus: true            # default false
opus:
  bitrate: 24000             # 0 = libopus default
  complexity: 10             # 0 = libopus default
  fec: true
  packet_loss_percent: 5
```

## Tests

- SDP negotiation: registration, dynamic PT (96–127), peer-PT echo (e.g. 111),
  offer/answer, Opus-only accept/reject.
- Preference & fallback: Opus preferred when enabled; G711/G722 fallback;
  PCMU/PCMA/G722 inbound by payload type; DTMF negotiation; graceful
  no-common-codec.
- Codec: encode/decode round-trip, encoder options, long-frame (60 ms) decode,
  DTX, in-band FEC.
- RTP: full MediaPort round-trip with timestamp-cadence assertion and
  silence-fill recovery. The round-trip test paces writes by polling received
  samples (no fixed sleeps); it is deterministic and `-race` clean.
- `go build ./...`, `go test`, `go vet`, and golangci-lint (staticcheck) clean.

## Limitations (Opus is opt-in; validate before GA)

- **Live SIP interop validation is still required** (FreeSWITCH/Asterisk/cloud
  trunk) before relying on Opus in production — see Q6. CI here covers
  unit/negotiation/RTP, not real endpoints.
- **No mid-call codec renegotiation:** the negotiated codec is fixed for the
  call; re-INVITEs are treated as keep-alives that replay the original SDP.
- **No Opus `fmtp` attributes** (`useinbandfec`, `minptime`, `stereo`) are
  signaled yet; peers use RFC 7587 defaults.
- **No PLC/FEC loss recovery for local packet loss:** missing/suppressed frames
  are silence-filled (same as the other codecs).

## Out of scope (intentionally, pending direction)

No `media-sdk` changes, no forced payload type, Opus not enabled by default,
existing codec and WebRTC behavior unchanged. Items in the open questions above
(media-sdk move, config-mechanism unification, `fmtp`, GlobalCodecs/defaultCodecs
reconciliation) are deferred until maintainers weigh in.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dennwc:
Most of this PR should definitely be in the `media-sdk`, not here. Please file a PR against that repo first.

PR #714, opened 2026-06-09
sip: route state updates through a public StateHandler

paulwe:
## Summary
- `CallState` now flows outgoing updates through a public `StateHandler` interface. The default `rpcStateHandler` forwards `UpdateSIPCallState` / `RecordCallContext` to the upstream client, preserving current behavior for non-cloud builds.
- `Handler.OnSessionEnd` takes `*CallState` instead of `*livekit.SIPCallInfo` so post-call writes (e.g. PCAP append after S3 upload) can go through `state.RecordCallContext(mutate)` instead of the raw RPC client.
- `Service` now takes a `GetStateHandler` factory so cloud builds can inject their own (legacy RPC + local observability fork).
- `internalProvidergetSIPTrunkAuthentication` parallelizes the auth/observability lookups (single query each via `sync.WaitGroup.Go`) and threads `SIPCallObservability` through the dispatch wrappers.

## Test plan
- [ ] `go test ./...` clean
- [ ] cloud-sip builds against this version and the StateHandler-based observability fork wires through correctly (separate PR)


pkg/sip/analytics.go
```diff
@@ -78,98 +127,84 @@ func (s *CallState) DeferUpdate(update func(info *livekit.SIPCallInfo)) {
...
-		s.dirty = false
+// RecordCallContext applies an optional mutation to the canonical callInfo
+// and signals the handler that the post-call context has been recorded —
+// e.g. late-arriving PCAP links published after the call has ended. Does not
+// touch the dirty bit: this is a terminal post-call signal, not a regular
+// flush.
+func (s *CallState) RecordCallContext(mutate func(info *livekit.SIPCallInfo)) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if mutate != nil {
```
nishadmusthafa: NIT: seems a bit obfuscated to me. Should we use something more obvious like `appendPcapInfo` instead of mutate ?

nishadmusthafa (approved):

PR #723, opened 2026-06-12
Adding error fan out to all waiting instances of TransferSIPParticipant

alexlivekit:


pkg/sip/service.go
```diff
@@ -374,6 +376,7 @@ func (s *Service) transferSIPParticipant(ctx context.Context, req *rpc.InternalT
 			default:
 				s.log.Errorw("pending transfer channel is full, dropping error", err, "callID", req.SipCallId, "transferTo", req.TransferTo)
```
hechen-eng: shall we reword the log? we are not dropping error now 👍

hechen-eng (approved):

PR #725, opened 2026-06-15
Implement end call participant RPC

dennwc:
This allows ending the call and sending specific headers. Our existing mechanism based on participant attributes is unreliable due to race condition between participant disconnecting and attribute propagation.


pkg/sip/protocol.go
```diff
@@ -40,6 +41,29 @@ var (
...
+		r.Status = sipStatus(r.Code)
+	}
+	return sip.NewResponseFromRequest(req, r.Code, r.Status, nil)
+}
+
+type EndCall struct {
+	Report  error      // reported to LiveKit analytics
+	Status  CallStatus // TODO: legacy
+	Term    stats.Termination
+	Room    livekit.DisconnectReason // disconnect reason for LiveKit participant
```
hechen-eng: should we name it `Reason`?

hechen-eng (approved):
This looks good!

PR #727, opened 2026-06-19
Make the SIP audio mixer input buffer size configurable

dshah1333:
## Summary
Adds a `mixer_input_buffer_frames` config option to size the SIP mixer's per-input buffer. Default unchanged (media-sdk's 5 frames / 100 ms) — opt-in, existing deployments unaffected.

## Why
The bridge mixes each input through media-sdk's mixer, whose input buffer defaults to 5 frames (100 ms). That can be too small for bursty sources: chunks longer than the buffer overflow on arrival and underrun between them, so the mixer restarts and emits silence — choppy outbound audio. `mixer.WithInputBufferFrames` already exists in media-sdk; the bridge just never passed it.

Driving the real mixer offline with identical bursty input, only the buffer changed:

| input buffer | mixer restarts | input frames dropped | injected silence |
|---|---|---|---|
| 5 frames (100 ms) | 85 | 172 / 602 | ~3.4 s / 12 s |
| 15 frames (300 ms) | 0 | 0 | 0 |
| steady input, either size | 0 | 0 | 0 |

## Change
- New `mixer_input_buffer_frames` (int; `0` = library default).
- Passes `mixer.WithInputBufferFrames(n)` to the room and DTMF mixers only when `n > 0`.
- Unit test + a doc comment on the latency trade-off.

## Trade-off
A deeper buffer adds up to `n × 20 ms` of per-input latency (15 frames raises the rebuffer threshold from 60 ms to 160 ms), so the default stays at 5 and operators raise it only when they hit choppy realtime-over-SIP audio. The same 5→15 value is hardcoded in a fork (Bigly-Sales/livekit-sip#6); this keeps it configurable instead. Related: livekit/media-sdk#38 tuned mixer timing but left the input buffer at 5.

### DIFF
--- a/pkg/sip/room.go
+++ b/pkg/sip/room.go
@@ -172,6 +172,23 @@ func DefaultGetRoomFunc(log logger.Logger, st *RoomStats) RoomInterface {
 	return NewRoom(log, st)
 }
 
+func getRoomFuncForConfig(conf *config.Config) GetRoomFunc {
+	if conf == nil || conf.MixerInputBufferFrames <= 0 {
+		return DefaultGetRoomFunc
+	}
+	mixerInputBufferFrames := conf.MixerInputBufferFrames
+	return func(log logger.Logger, st *RoomStats) RoomInterface {
+		return NewRoom(log, st, WithRoomMixerInputBufferFrames(mixerInputBufferFrames))
+	}
+}
+
+func withMixerInputBufferFrames(frames int, opts ...mixer.MixerOptions) []mixer.MixerOptions {
+	if frames > 0 {
+		opts = append(opts, mixer.WithInputBufferFrames(frames))
+	}
+	return opts
+}
+
 type Room struct {
 	log        logger.Logger
 	roomLog    logger.Logger // deferred logger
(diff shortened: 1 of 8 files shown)


dennwc (approved):
LGTM! @alexlivekit could you have another look?

pkg/sip/room.go
```diff
@@ -172,6 +172,23 @@ func DefaultGetRoomFunc(log logger.Logger, st *RoomStats) RoomInterface {
...
 
+func getRoomFuncForConfig(conf *config.Config) GetRoomFunc {
+	if conf == nil || conf.MixerInputBufferFrames <= 0 {
+		return DefaultGetRoomFunc
+	}
+	mixerInputBufferFrames := conf.MixerInputBufferFrames
+	return func(log logger.Logger, st *RoomStats) RoomInterface {
+		return NewRoom(log, st, WithRoomMixerInputBufferFrames(mixerInputBufferFrames))
+	}
+}
```
alexlivekit, line 183: Would probably want to see something more generic here than the tailor-made WithRoomMixerInputBufferFrames() option, but this can be iterated upon at a later time.

alexlivekit (approved):

abhiin1947:
@dennwc @alexlivekit Bumping to see if there's anything left to do here to merge this in. We ran into this recently and would love to get this working.

PR #728, opened 2026-06-21
sip: redirect RTP destination on re-INVITE port change

### LINKED ISSUE #661: SIP service does not redirect RTP to new remote port after re-INVITE changes m= line
Carrier sent a re-INVITE mid-call that changed the remote RTP port from 25540 to 25446 (codec offer changed PCMU → G.722). LiveKit sent a 200 OK with its own SDP unchanged, but continued transmitting RTP to the old port for the remainder of the call. Per RFC 3264 §8, the answerer must send media to the new port in the updated offer.

[SCL_VGXD6nQxzwks_sip_trace.txt](https://github.com/user-attachments/files/26993471/SCL_VGXD6nQxzwks_sip_trace.txt)

VictorUvarov:
## Problem

When a SIP carrier sends a mid-call re-INVITE with a new RTP port (e.g. after a NAT rebind or media relay failover), LiveKit SIP was responding 200 OK but continuing to send RTP to the stale address — violating [RFC 3264 §8](https://www.rfc-editor.org/rfc/rfc3264#section-8), which requires the answerer to direct media to the address/port in the new offer.

Closes #661

## What changed

**`pkg/sip/media_port.go`**: two new methods on `MediaPort`:
- `UpdateRemote`: atomically redirects the RTP destination.
- `RemoteAddr`: read accessor, used in tests.

**`pkg/sip/inbound.go`** : both re-INVITE paths now parse the SDP body and call `UpdateRemote` before responding 200 OK.

## Tests

`pkg/sip/media_port_test.go` — `TestMediaPortUpdateRemote` covers thehappy path, the zero-value no-op, and the unspecified-addr no-op.  `pkg/sip/signaling_test.go` — `TestReinvite` extended:

- `inbound/normal`: asserts `RemoteAddr()` changes after a re-INVITE.
- `outbound/normal`: same for the outbound path.
- `inbound/no_body`: asserts a body-less re-INVITE leaves `RemoteAddr()` unchanged.

---

### Unit tests
```bash
o test ./pkg/sip/... -run 'TestMediaPortUpdateRemote|TestReinvite/inbound/normal|TestReinvite/inbound/no_body|TestReinvite/outbound' -v  -count=1
```

### Integration tests
```bash
go test ./pkg/sip/... -count=1
```

---
  
#### Thought process summary
  
The root cause was a one-liner gap: the re-INVITE handler called AcceptAsKeepAlive (200 OK) but never forwarded the new c=/m= address to the RTP layer. The SDP body already contained the new addr+port as desc.Addr after ParseWith — nothing needed to be renegotiated, just the destination pointer swapped.

The tricky edge case was [RFC 3264 §8.4](https://www.rfc-editor.org/info/rfc3264/#section-8.4) call-hold, where c=0.0.0.0 signals hold and `netip.MustParseAddrPort("0.0.0.0:N").IsValid()` returns `true`. A plain `IsValid()` guard would have let hold re-INVITEs overwrite the real destination with the zero address. The `!addr.Addr().IsUnspecified()` check inside UpdateRemote blocks that.

### DIFF
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -312,6 +312,14 @@ func (s *Server) handleInviteAuth(tid traceid.ID, log logger.Logger, req *sip.Re
 	return true, false
 }
 
+func sdpBodyFromRequest(req *sip.Request) []byte {
+	ct := req.ContentType()
+	if ct != nil && ct.Value() != "application/sdp" {
+		return nil
+	}
+	return req.Body()
+}
+
 func updateRemoteFromSDP(media *MediaPort, log logger.Logger, body []byte) {
 	if len(body) == 0 || media == nil {
 		return
@@ -393,7 +401,7 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 	s.cmu.RUnlock()
 	if existing != nil && existing.cc.InviteCSeq() < cc.InviteCSeq() {
 		existing.log().Infow("reinvite", "content-type", req.ContentType(), "content-length", req.ContentLength(), "cseq", cc.InviteCSeq())
-		updateRemoteFromSDP(existing.media, existing.log(), req.Body())
+		updateRemoteFromSDP(existing.media, existing.log(), sdpBodyFromRequest(req))
 		cc.AcceptAsKeepAlive(existing.cc.OwnSDP())
 		return nil
 	}
@@ -404,7 +412,7 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 			localSDP := oc.cc.LocalSDP()
 			if len(localSDP) != 0 {
 				oc.log.Infow("accepting reinvite", "content-type", req.ContentType(), "content-length", req.ContentLength(), "cseq", cc.InviteCSeq())
-				updateRemoteFromSDP(oc.media, oc.log, req.Body())
+				updateRemoteFromSDP(oc.media, oc.log, sdpBodyFromRequest(req))
 				oc.cc.RecordInvite(newCSeq)
 				cc.AcceptAsKeepAlive(localSDP)
 				return nil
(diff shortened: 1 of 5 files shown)


pkg/sip/inbound.go
```diff
@@ -380,19 +400,21 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
...
 		oc := s.cli.getActiveCall(cc.ID())
 		newCSeq := cc.InviteCSeq()
 		if oc != nil && oc.cc != nil && oc.cc.InviteCSeq() < newCSeq {
-			sdp := oc.cc.LocalSDP()
-			if len(sdp) != 0 {
-				oc.log.Infow("accepting reinvite", "content-type", req.ContentType(), "content-length", req.ContentLength(), "cseq", cc.InviteCSeq())
+			localSDP := oc.cc.LocalSDP()
+			if len(localSDP) != 0 {
+				oc.log.Infow("accepting reinvite", "content-length", req.ContentLength(), "cseq", cc.InviteCSeq())
+				updateRemoteFromSDP(oc.media, oc.log, sdpBodyFromRequest(req))
```
dennwc: I think it might be better to make a separate method on both inbound and outbound call types that call `updateRemoteFromSDP` internally. The reason is that here it is unclear if we can safely access `media` field without holding any mutexes.

pkg/sip/inbound.go
```diff
@@ -312,6 +312,26 @@ func (s *Server) handleInviteAuth(tid traceid.ID, log logger.Logger, req *sip.Re
...
+		return nil
+	}
+	return req.Body()
+}
+
+func updateRemoteFromSDP(media *MediaPort, log logger.Logger, body []byte) {
+	if len(body) == 0 || media == nil {
+		return
+	}
+	desc, err := sdp.ParseWith(msdk.GlobalCodecs(), body)
```
dennwc: `msdk.GlobalCodecs()` is incorrect in this context. It should use the CodecSet for that particular call.

dennwc (commented):
Looks good in general, thank you! A few small changes needed:

dennwc (approved):
LGTM! @alexlivekit would you mind taking another look?

alexlivekit (approved):

PR #734, opened 2026-06-26
Clearing cached destination before sending ACK or other request

alexlivekit:
Scenario:
- Inivite is created, no destination set.
- We send the invite on its way. It passes through sipgo.
- In sipgo, when sending the request out, [we calculate the destination](https://github.com/livekit/sipgo/blob/a5b4a38b6ceb491fbc36c2ef8db6e1bb3e35c32f/transport/layer.go#L271-L299).
- If it's an IP address, all is well.
- If it is a hostname, we resolve it, and fix the destination to that now-resolved IP using `req.SetDestination()`.

Then, when the 200 arrives:
- We [update c.invite.Recipient with contact](https://github.com/livekit/sip/blob/ab002ab5e66bdf2117292585a8198b5e4db30cd1/pkg/sip/outbound.go#L1017-L1032), but do not flush the destination stored in `req.MessageData.Destination()`.
- Construct an ACK to send it back.
- While constructing the ACK, we copy the updated recepient, but crucially we also `ackRequest.SetDestination(inviteRequest.Destination())` [here](https://github.com/livekit/sipgo/blob/a5b4a38b6ceb491fbc36c2ef8db6e1bb3e35c32f/sip/request.go#L112). `inviteRequest.Destination()` returns the fixed destination, and `ackRequest.SetDestination()` fixes it for the ACK response to the same thing. This [takes precedence](https://github.com/emiago/sipgo/blob/3dffa2770b218ace086a06d5b69bb093f413e5ec/sip/request.go#L207-L232) over `Route` headers, `Recepient`, and the `Contact` headr that might have been updated.

### DIFF
--- a/pkg/sip/outbound.go
+++ b/pkg/sip/outbound.go
@@ -22,6 +22,7 @@ import (
 	"net"
 	"sort"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


hechen-eng (approved):
Alt Text

PR #735, opened 2026-06-29
TEL-588: move no-ack to indeterminate

hechen-eng:
media timeout happens and there is no-ack received earlier, so most likely not server_error.


alexlivekit:
Um... calls should have ACKs. Why do we want to reclassify this as not-an-issue?

hechen-eng:
> Um... calls should have ACKs. Why do we want to reclassify this as not-an-issue?

this is an issue, but not likely a server error..

btw, we don't wait for ACK for inbound calls as of today (`c.s.conf.Experimental.InboundWaitACK` is false)

alexlivekit (approved):

PR #737, opened 2026-07-02
sip: classify Twilio carrier fraud blocks distinctly from callee decl…

nishadmusthafa:
…ines

Twilio surfaces a permanent fraud/regulatory block (X-Twilio-Error 32203) as a generic SIP 603 Decline, which was indistinguishable from an ordinary callee rejection: both classified as USER_REJECTED / global-decline-603. Callers that want to suppress a permanently-blocked number couldn't tell the two apart on a failed CreateSIPParticipant.

Detect the X-Twilio-Error code on a non-2xx INVITE response and, for known permanent-block codes, wrap it in a self-classifying carrierBlockedError. It classifies as a distinct ClientError("carrier-blocked") SLI bucket and carries the provider code in the reported error, while preserving the underlying SIPStatus so CallStatusCode and the RPC boundary are unchanged. SIPStatus.Status (the SIP reason phrase) is not overwritten.

Lives alongside the existing Twilio rate-limit handling in classifyInviteError.

### DIFF
--- a/pkg/sip/errors.go
+++ b/pkg/sip/errors.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"strconv"
 	"strings"
 
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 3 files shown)


pkg/sip/errors.go
```diff
@@ -100,6 +101,86 @@ func (e transactionTimeoutError) ClassifyInvite() inviteFailure {
...
 
+// twilioPermanentBlockCodes are Twilio X-Twilio-Error codes for calls a carrier
+// permanently blocks for fraud/regulatory reasons. Twilio surfaces these as a
+// generic SIP 603 Decline, which is indistinguishable from an ordinary callee
+// rejection, so we key off the proprietary error code instead. The distinct
+// classification lets callers suppress the destination rather than retrying it.
+//
+// This lives alongside the Twilio rate-limit handling in classifyInviteError.
+var twilioPermanentBlockCodes = map[int]struct{}{
+	32203: {}, // call blocked as potential fraud / regulatory block
```
hechen-eng, line 112: This is a map. Does it mean there are other permanent block codes? should we add all of them now?

hechen-eng (approved):
LGTM

pkg/sip/errors.go
nishadmusthafa: Good question. I think not given the discussion we are having on where we want to do this. This is going to block an immediate customer.

PR #741, opened 2026-07-10
Adding RTP port draining to the system

alexlivekit:


pkg/config/config.go
```diff
@@ -44,6 +44,13 @@ var (
 	DefaultRTPPortRange = rtcconfig.PortRange{Start: 10000, End: 20000}
 )
 
+const (
+	// After a call closes we keep its RTP port bound and draining so a freshly
+	// allocated call can't inherit a port a peer is still sending stale media to.
```
hechen-eng, line 49: what happens when a new call inherits a port that a peer is still sending stale packets to 🤔

pkg/sip/draining_port.go
```diff
@@ -0,0 +1,75 @@
...
+	"net"
+	"os"
+	"time"
+
+	"github.com/livekit/protocol/logger"
+)
+
+func DrainPort(log logger.Logger, conn *net.UDPConn, idleTimeout, maxDuration time.Duration) {
+	defer conn.Close()
+	if idleTimeout <= 0 && maxDuration <= 0 {
```
hechen-eng, line 32:
```suggestion
	if idleTimeout <= 0 || maxDuration <= 0 {
```

pkg/sip/draining_port.go
```diff
@@ -0,0 +1,75 @@
...
+		return
+	}
+	log = log.WithValues("local", conn.LocalAddr(), "idleTimeout", idleTimeout, "maxDuration", maxDuration)
+	if idleTimeout <= 0 {
```
hechen-eng, line 36: if it's 0, it means it's disabled, no?

hechen-eng (commented):
fix the CI?

pkg/config/config.go
```diff
@@ -44,6 +44,13 @@ var (
...
+	DefaultRTPDrainingIdleTimeout = 30 * time.Second // release once no packets arrive for this long
+	DefaultRTPDrainingDuration    = 10 * time.Minute // hard cap on how long a port stays draining
```
hechen-eng, line 51: We have media timeout 30s. I don't think we should set a 10mins cap here 🤔

hechen-eng:
we probably need production data to see the impact of this since a high rate of short calls can keep a lot ports in the draining state simultaneously, at which port new calls can't allocate a port

pkg/sip/draining_port.go
alexlivekit: hmm? I think we're okay if only one is specified.

pkg/sip/draining_port.go
alexlivekit: either one can be on or off

pkg/config/config.go
alexlivekit: What we have doesn't matter here. The titular case was seeing 5+ minutes of RTP coming to us from the remote (non-livekit endpoint). 

This isn't about closing the call leg if there's no media, it's about making sure the port isn't polluted by another RTP stream, and that is not up to us.

pkg/config/config.go
alexlivekit: That's the whole point of this commit - We do not release the port until we're sure the port is not receiving garbage from the peer. Hence the Drain port goroutine keeping previously-used ports allocated, and hence the extra-long max wait duration, that's an order of magnitude larger than our media timeout.

hechen-eng (approved):

PR #742, opened 2026-07-14
Updating DTMF detection to work based off of timestamp rather than first marker

alexlivekit:


pkg/sip/media_port.go
```diff
@@ -1015,9 +1017,20 @@ func (p *MediaPort) dtmfHandler(h *rtp.Header, payload []byte) error {
 		return nil
 	}
 	fnc := *ptr
-	if ev, ok := dtmf.DecodeRTP(h, payload); ok && fnc != nil {
-		fnc(ev)
+	if fnc == nil {
+		return nil
+	}
+	// RFC 4733 requires a given digit to identical timestamps across packets.
```
hechen-eng:
```suggestion
	// RFC 4733 requires all packets of a given digit to share identical timestamps.
```

pkg/sip/media_port.go
```diff
@@ -1015,9 +1017,20 @@ func (p *MediaPort) dtmfHandler(h *rtp.Header, payload []byte) error {
...
+	// The marker bit can be used, is prone to occasional loss.
```
hechen-eng:
```suggestion
	// The marker bit could be used instead, but it is prone to occasional loss.
```

hechen-eng (approved):

PR #750, opened 2026-07-22
Ignore duplicate CreateSIPParticipant request from sdk retry

cnderrauber:


hechen-eng (changes requested):
We don't dedup for CreateSIPParticipant today and I think it's a gap we need to fill. There are a couple problems with this PR though:
1. not tested.. e.g. is req.Token enough for GetParticipant?
2. what if WsUrl is missing?
3. do we want to create a new LiveKitAPI per call?
4. no timeouts
5. what if a retry arrives before we can see the participant?
6. waitUntilAnswered may be broken since it returns immeidately
7. ...

I think dedup should be handled by sip itself, not querying the ws url to get participant. They are two separate systems and there are consistency issues.

alexlivekit:
> do we want to create a new LiveKitAPI per call?

+1. This is the big one IMO, and goes against being untethered to centralized access on a per-call basis.
Moreover, this has the potential to significantly increase call setup times in certain cases.

PR #752, opened 2026-07-22
reduce log level of a few low-value logs

davidzhao:


pkg/sip/media_port.go
```diff
@@ -758,7 +758,7 @@ func (p *MediaPort) SetConfig(c *MediaConf) error {
 	if c.Crypto != nil {
 		crypto = c.Crypto.Profile.String()
 	}
-	p.log.Infow("using codecs",
+	p.log.Debugw("using codecs",
```
dennwc: This is the only place that records the codecs in logs. Given that we started expanding the list of supported codecs recently, I'd rather keep it as info for now.

dennwc (approved):

pkg/sip/media_port.go
```diff
@@ -813,7 +813,7 @@ func (p *MediaPort) rtpLoop(tid traceid.ID, sess rtp.Session) {
 		p.stats.Streams.Add(1)
 		p.mediaReceived.Break()
 		log := p.log.WithValues("ssrc", ssrc)
-		log.Infow("accepting RTP stream")
```
alexlivekit, line 816: This one helped us find both [these](https://linear.app/livekit/issue/TEL-389/calls-spamming-addr-changes) [issues](https://linear.app/livekit/issue/TEL-835/calls-spamming-ssrc-change) out, but once we handle the second one as well this can be a number, no need to log every time.

PR #756, opened 2026-07-24
TEL-717: ack/bye to 200 ok after cancel

hechen-eng:
cancel and 200 ok is a common race. If 200 ok arrives after cancel, we should ACK and BYE.

### DIFF
--- a/pkg/sip/outbound.go
+++ b/pkg/sip/outbound.go
@@ -558,6 +558,46 @@ func sipResponse(ctx context.Context, tx sip.ClientTransaction, stop <-chan stru
 	}
 }
 
+// cancelResponseGrace bounds how long a cancelled INVITE is watched for a 2xx.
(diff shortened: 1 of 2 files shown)


hechen-eng:
not related to this change, but it seems we have some flaky tests https://github.com/livekit/sip/actions/runs/30114193566/attempts/1?pr=756

alexlivekit (changes requested):
> If 200 ok arrives after cancel, we should ACK and BYE.

~~If this is an outbound call, only we can be sending CANCEL requests, and these should be rejected/dropped. If the remote party wants to reject an incoming (lk -> remote) call, they can simply respond with a >=400 status code.~~ 

~~What is this trying to fix exactly?~~

alexlivekit (approved):
I stand corrected, we absolutely need to ACK & BYE the call if we want to abort, but receive a 200.

PR #762, opened 2026-07-28
Fix bug where max call duration was not being respected for outbound calls.

alexfish8:
Also in this PR:
- A couple of nil checks in outbound.go
  - Add a nil check against `*outboundCall.media` in `outboundCall.close` (if the call to `NewMediaPort` in `newCall` fails, then we'd hit this nil deref).
  - Add a nil check against the response returned from selecting from `tx.Responses` in `sipResponse.
- Make one of the outbound test utility channels buffered to avoid a race in the tests.

### DIFF
--- a/pkg/sip/client.go
+++ b/pkg/sip/client.go
@@ -432,6 +432,7 @@ func (c *Client) createSIPParticipant(ctx context.Context, req *rpc.InternalCrea
 		mediaConfig:     mconf,
 	}
 	log.Infow("Creating SIP participant")
+	deadline := time.Now().Add(sipConf.maxCallDuration)
 	call, err := c.newCall(ctx, tid, c.conf, log, LocalTag(req.SipCallId), roomConf, sipConf, state, req.ProjectId)
 	if err != nil {
 		return nil, err
@@ -448,10 +449,19 @@ func (c *Client) createSIPParticipant(ctx context.Context, req *rpc.InternalCrea
 		call.DialAsync(ctx)
 		return info, nil
 	}
-	if err := call.Dial(ctx); err != nil {
+
+	dialCtx, dialCancel := context.WithDeadline(ctx, deadline)
+	defer dialCancel()
+
+	if err := call.Dial(dialCtx); err != nil {
 		return nil, err
 	}
-	go call.WaitClose(context.WithoutCancel(ctx))
+
+	// Detach this context from the context of the RPC so that the RPC returning
+	// doesn't cause the call to end prematurely.
+	detachedCtx, detatchedCancel := context.WithDeadline(ctx, deadline)
+	defer detatchedCancel()
+	go call.WaitClose(detachedCtx)
 	return info, nil
 }
 
--- a/pkg/sip/client.go
+++ b/pkg/sip/client.go
@@ -458,11 +458,11 @@ func (c *Client) createSIPParticipant(ctx context.Context, req *rpc.InternalCrea
 		return nil, err
 	}
 
-	// Detach this context from the context of the RPC so that the RPC returning
-	// doesn't cause the call to end prematurely.
+	// Detach this context from that of the RPC so that when the RPC returns,
+	// the call doesn't end prematurely.
 	go func() {
-		detachedCtx, detatchedCancel := context.WithDeadline(context.WithoutCancel(ctx), deadline)
-		defer detatchedCancel()
+		detachedCtx, detachedCancel := context.WithDeadline(context.WithoutCancel(ctx), deadline)
+		defer detachedCancel()
 		call.WaitClose(detachedCtx)
 	}()
 	return info, nil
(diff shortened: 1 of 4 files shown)


pkg/sip/outbound.go
```diff
@@ -244,6 +244,9 @@ func (c *outboundCall) waitClose(ctx context.Context, tid traceid.ID) error {
 	ctx = context.WithoutCancel(ctx)
 	defer c.ensureClosed(ctx)
 
+	ctx, cancel := context.WithTimeout(ctx, c.sipConf.maxCallDuration)
```
hechen-eng: `Dial` applies `WithTimeout(maxCallDuration)`, so the total wall-clock is therefore dial time + maxCallDuration 🤔

pkg/sip/outbound.go
```diff
@@ -270,6 +273,13 @@ func (c *outboundCall) waitClose(ctx context.Context, tid traceid.ID) error {
 			err := psrpc.NewErrorf(psrpc.DeadlineExceeded, "media timeout")
 			c.setErrStatus(ctx, err)
 			return err
+		case <-ctx.Done():
+			c.CloseWith(ctx, EndCall{
+				Status: CallHangup,
+				Term:   stats.Success("max-call-duration"),
```
hechen-eng: what does inbound `Term` look like if it's max-call-duration exceeded? shall we make outbound to be consistent with that?

pkg/sip/outbound.go
```diff
@@ -270,6 +273,13 @@ func (c *outboundCall) waitClose(ctx context.Context, tid traceid.ID) error {
...
+				Reason: livekit.DisconnectReason_CLIENT_INITIATED,
```
alexlivekit, line 274: hmmm.. Would `DisconnectReason_CONNECTION_TIMEOUT` be more accurate here?

pkg/sip/outbound.go
alexfish8: It uses `stats.Success("hangup")`. Updated outbound to be consistent with this.

pkg/sip/outbound.go
alexfish8: When the inbound path times out due to max call duration being exceeded, the reason is given as `CLIENT_INITIATED`. Consistency aside, `CLIENT_INITIATED` feels more appropriate, as, ultimately, it was the client's request's configuration that caused the call to end early (but, I don't feel super strongly here. I can change both the inbound / outbound paths to give `CONNECTION_TIMEOUT` as the reason if you think that'd be more appropriate.)

pkg/sip/outbound.go
alexfish8: Good call, reorganized the code so that we use a single deadline for both the dial and waitClose.

alexfish8 (commented):
Thanks for the reviews!

pkg/sip/client.go
```diff
@@ -448,10 +450,21 @@ func (c *Client) createSIPParticipant(ctx context.Context, req *rpc.InternalCrea
...
+
+	if err := call.Dial(dialCtx); err != nil {
 		return nil, err
 	}
-	go call.WaitClose(context.WithoutCancel(ctx))
+
+	// Detach this context from the context of the RPC so that the RPC returning
+	// doesn't cause the call to end prematurely.
+	go func() {
+		detachedCtx, detatchedCancel := context.WithDeadline(context.WithoutCancel(ctx), deadline)
```
hechen-eng:
```suggestion
		detachedCtx, detachedCancel := context.WithDeadline(context.WithoutCancel(ctx), deadline)
```

pkg/sip/client.go
```diff
@@ -448,10 +450,21 @@ func (c *Client) createSIPParticipant(ctx context.Context, req *rpc.InternalCrea
...
+		defer detatchedCancel()
```
hechen-eng:
```suggestion
		defer detachedCancel()
```

pkg/sip/outbound.go
```diff
@@ -270,6 +267,13 @@ func (c *outboundCall) waitClose(ctx context.Context, tid traceid.ID) error {
 			err := psrpc.NewErrorf(psrpc.DeadlineExceeded, "media timeout")
 			c.setErrStatus(ctx, err)
 			return err
+		case <-ctx.Done():
+			c.CloseWith(ctx, EndCall{
+				Status: CallHangup,
+				Term:   stats.Success("hangup"),
```
hechen-eng, line 273: out of scope of this PR, but I think it's beneficial to differentiate max-duration reached vs normal hangup. Feel free to create a follow up ticket for this 👍

hechen-eng (approved):

pkg/sip/outbound.go
alexlivekit: Yeah not a strong opinion here either. looks good, we can always tweak later.

alexlivekit (approved):

pkg/sip/outbound.go
alexfish8: Good point, created https://linear.app/livekit/issue/TEL-832/monitoring-differentiate-between-hangups-caused-by-max-duration-limits.

pkg/sip/client.go
alexfish8: Good catch!

PR #767, opened 2026-08-03
fix: reject codec-changing re-INVITE with 488

### LINKED ISSUE #766: Codec-changing re-INVITE gets 200 OK with stale SDP
Hi, we ran into this on an outbound carrier call and I think the current
re-INVITE handling has a gap.

The initial exchange selected PCMU. Right after the ACK, the carrier sent an
in-dialog re-INVITE offering only PCMA. LiveKit answered `200 OK` with the
cached PCMU SDP from the initial exchange and kept the media pipeline on PCMU.
That left no common audio codec between the new offer and the answer, and the
call had no usable return audio.

Simplified SDP sequence:

```sdp
# Initial LiveKit offer
m=audio 10820 RTP/AVP 9 0 8 101
a=rtpmap:0 PCMU/8000
a=rtpmap:8 PCMA/8000

# Initial peer answer
m=audio 45832 RTP/AVP 0 101
a=rtpmap:0 PCMU/8000

# Peer re-INVITE
m=audio 45832 RTP/AVP 8 96
a=rtpmap:8 PCMA/8000

# LiveKit 200 OK
m=audio 10820 RTP/AVP 0 101
a=rtpmap:0 PCMU/8000
```

The last answer contains PCMU even though the re-INVITE offered only PCMA.
RFC 3264 section 6.1 requires a `sendrecv` answer to contain at least one codec
from the offer:

https://www.rfc-editor.org/rfc/rfc3264.html#section-6.1

I checked current `main` at
`179b2574d8e6b97306855cf0c8628b17f53c3e92`. The re-INVITE path still parses the
new SDP only to update the RTP address, then returns the old local SDP:

- https://github.com/livekit/sip/blob/179b2574d8e6b97306855cf0c8628b17f53c3e92/pkg/sip/inbound.go#L323-L333
- https://github.com/livekit/sip/blob/179b2574d8e6b97306855cf0c8628b17f53c3e92/pkg/sip/inbound.go#L408-L418

I also added a small regression test locally. Before the guard, LiveKit
returned `200`; after checking that the currently negotiated codec and payload
types are still offered, the same re-INVITE returns `488 Not Acceptable Here`
and leaves the existing RTP destination untouched.

I think either behavior below would be valid:

1. Fully renegotiate the codec and update the media pipeline before answering
   with matching SDP; or
2. Reject a codec/payload change with `488` and keep the established session.

What should not happen is a successful answer containing a codec that was not
in the offer.

This looks related to #661 and #728. The RTP destination update from #728 works,
but the codec change mentioned in #661 still appears to be unhandled.

If you want I can also fix it myself and create a Pull Request for this.

lixuanqun:
When an in-dialog re-INVITE offers SDP that no longer includes the currently negotiated audio codec/payload type, answer 488 Not Acceptable Here and leave the RTP destination unchanged.

Previously LiveKit returned 200 OK with the cached local SDP, which could advertise a codec that was not in the offer (RFC 3264 §6.1) and leave the call with no usable media. Address-only re-INVITEs that keep the negotiated codec continue to update the RTP destination as before.

Adds inbound/outbound regression coverage for codec mismatch.

Fixes livekit/sip#766

### DIFF
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -320,16 +320,61 @@ func sdpBodyFromRequest(req *sip.Request) []byte {
 	return req.Body()
 }
 
-func updateRemoteFromSDP(media *MediaPort, log logger.Logger, codecs *msdk.CodecSet, body []byte) {
+// applyReinviteSDP updates the RTP destination from a re-INVITE SDP body when
+// the currently negotiated audio codec and payload type remain offered.
+//
+// An empty body is treated as a session refresh (ok=true, no media change).
+// Parse failures leave the destination unchanged and still return ok=true so
+// the dialog can be refreshed with the existing local SDP (historical behavior).
+//
+// When the offer removes the negotiated audio codec/payload, ok=false and the
+// caller must reject with 488 without changing the media path. Answering 200
+// with the cached local SDP would advertise a codec that was not in the offer
+// (see livekit/sip#766 / RFC 3264 §6.1).
+func applyReinviteSDP(media *MediaPort, log logger.Logger, codecs *msdk.CodecSet, body []byte) (ok bool) {
 	if len(body) == 0 || media == nil {
-		return
+		return true
 	}
 	desc, err := sdp.ParseWith(codecs, body)
 	if err != nil {
 		log.Warnw("failed to parse re-INVITE SDP, RTP destination not updated", err)
-		return
+		return true
+	}
+	if !reinviteOffersCurrentAudio(media, desc) {
+		cur := media.Config()
+		var curType byte
+		var curName string
+		if cur != nil && cur.Audio.Codec != nil {
+			curType = cur.Audio.Type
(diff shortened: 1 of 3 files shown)


lixuanqun:
Rebased onto latest main (`f75e9ce`) and resolved conflicts against the media port rework (#802) and the newer re-INVITE handling.

**Heads-up on test expectations:** main's `TestReinvite/incompatible_codec` (added after this PR was opened) asserted the old behavior — codec-changing re-INVITEs get `200 OK` and the RTP destination is updated. Since rejecting with 488 is the point of this PR (RFC 3264 §6.1: answering 200 with the cached local SDP would advertise a codec that is no longer in the offer, see #766), I updated both the inbound and outbound `incompatible_codec` subtests to expect `488` and an unchanged RTP destination. Happy to split that test change out or adjust if you'd rather keep the current semantics.

PR #773, opened 2026-08-03
TEL-764 Add new metrics for sdp negotiation

genseric-ghiro:
## Summary
- Adding two new metrics (`sdpParsed`, `codecOffered`) to record when an SDP negotiation takes place, and what codecs are offered. This is only for inbound calls as the set of offered codecs for outbound calls is always the same (and the answer only contains what the user picked, and not what they support)
- Adding the provider label to these two metrics. The goal is to extend this tag to all the existing metrics (where applicable in a follow-up).

## Testing
- New tests in media_port, media_codecs and inbound

### DIFF
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -320,6 +320,27 @@ func sdpBodyFromRequest(req *sip.Request) []byte {
 	return req.Body()
 }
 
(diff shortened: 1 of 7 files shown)


dennwc (approved):
LGTM!

Issue #524, opened 2025-11-15
SAD REGISTER BOYS


AL-Amin-code:
Thread 4 ppl willing to have REGISTER x-mark period in LK sip component nor only send INVITES.
At least provide us timings when it gonna be implemented. 

Thanks.

alexvahter:
Any updates on adding REGISTER support?

8thgencore:
If there are professionals here, how difficult is it to implement the registration option?

infinitisno:
I am using kamailio as a workaround, and used asterisk as well, configuration for both were relatively simple for my usecase.

Here is my configuration for kamailio (it just solves my specific use case, does not handle all edge cases, etc):
```
#!KAMAILIO
#!defenv LISTEN_PORT
#!defenv REMOTE_TRUNK_HOST
#!defenv REMOTE_TRUNK_PORT
#!defenv REMOTE_TRUNK_USER
#!defenv REMOTE_TRUNK_PASSWORD
#!defenv REMOTE_TRUNK_REALM
#!defenv PUBLIC_HOST
#!defenv PUBLIC_PORT
#!defenv MYSQL_USER
#!defenv MYSQL_PASSWORD
#!defenv MYSQL_HOST
#!defenv MYSQL_PORT
#!defenv MYSQL_DATABASE
#!substdef "/_LISTEN_PORT/$env(LISTEN_PORT)/"
#!substdef "/_MYSQL_USER/$env(MYSQL_USER)/"
#!substdef "/_MYSQL_PASSWORD/$env(MYSQL_PASSWORD)/"
#!substdef "/_MYSQL_HOST/$env(MYSQL_HOST)/"
#!substdef "/_MYSQL_PORT/$env(MYSQL_PORT)/"
#!substdef "/_MYSQL_DATABASE/$env(MYSQL_DATABASE)/"
#!substdef "/_PUBLIC_HOST/$env(PUBLIC_HOST)/"
#!substdef "/_PUBLIC_PORT/$env(PUBLIC_PORT)/"
#!substdef "/_TRUNK_HOST/$env(REMOTE_TRUNK_HOST)/"
#!substdef "/_TRUNK_PORT/$env(REMOTE_TRUNK_PORT)/"
#!substdef "/_TRUNK_USER/$env(REMOTE_TRUNK_USER)/"
#!substdef "/_TRUNK_PASS/$env(REMOTE_TRUNK_PASSWORD)/"
[...]

PR #774, opened 2026-08-04
feat: outbound SIP registration (REGISTER)

skyzer:
## Problem

LiveKit SIP can only be an INVITE-only endpoint: a provider has to know a stable address to send inbound INVITEs to. Providers commonly offer a second inbound mode for endpoints that do not have one — the endpoint sends `REGISTER`, and calls are delivered to the resulting binding. DIDWW, for example, offers it explicitly for a PBX or SBC that "has a dynamic IP address, operates behind NAT, or when you prefer registration-based routing". Without `REGISTER`, a self-hosted deployment on a dynamic IP, behind NAT, or on CGNAT cannot accept inbound calls at all.

There is nothing partial to build on today: the registered method handlers in `pkg/sip/server.go` are `OnOptions`/`OnInvite`/`OnAck`/`OnBye`/`OnNotify`/`OnNoRoute`, and `pkg/config/config.go` has no registration keys.

## Why not just run a proxy

That is the workaround in #524, and it does work — Kamailio `uac_reg`, OpenSIPS `uac_registrant`, or Asterisk PJSIP in front of LiveKit SIP on localhost. We run one downstream and have taken it end to end on a live PSTN call through a real NAT. It costs a second SIP element, with its own config, container and failure modes, whose only job is to send REGISTER — in front of a service that already owns the trunk state and already knows its own signaling address. It also rewrites the source of every inbound INVITE to the proxy, so LiveKit SIP no longer sees which carrier a call came from, and the NAT details below have to be solved again in the proxy's config language.

## Design

Config-file only. No `livekit/protocol` change, so this is reviewable and mergeable in one repo.

```yaml
sip_registrations:
  - registrar: sip:sip.provider.example   # or host[:port]
    username: 1000
    password: secret
    # auth_username: 1000                 # digest username, if issued separately
    # domain: provider.example            # AOR host, defaults to the registrar host
    # expiry: 10m
    # keepalive: 25s                      # negative to disable
```

I did consider putting this on `SIPOutboundTrunkInfo`/`SIPInboundTrunkInfo`, which is where @lixuanqun's design comment on #524 proposes it. That is arguably the better product surface, but it is a cross-repo change, and it forces the multi-node question immediately: a trunk-level flag implies every node holding that trunk registers, so it needs owner election before it is safe. A registration is really a property of one node's connectivity to a provider rather than of a project's trunk record, so per-node config is both the smaller change and, I think, the more accurate model. The two compose — trunk-level fields can be layered on later without changing anything here. If you would rather have the protobuf version, I am glad to write that instead; I would just rather agree on it first.

Registration is a prerequisite for inbound calls, not a path for them. Once the provider has a binding, its INVITEs arrive on the existing inbound path and are authenticated and dispatched by trunk exactly as they are today. `inbound.go` is untouched.

## What it does

`pkg/sip/register.go` adds one `registrant` per configured account, started after the server is listening and stopped before the client closes.

- REGISTER on startup, digest auth on 401 and 407, refresh ahead of expiry, un-REGISTER with `Expires: 0` on shutdown, exponential backoff on failure.
- One Call-ID and a monotonic CSeq for the life of the registration (RFC 3261 §10.2).
- The digest URI is the Request-URI, which for REGISTER is the registrar, not the AOR in `To`.
- `423 Interval Too Brief` negotiation, and the negotiated value is kept for later refreshes instead of being re-negotiated every time. (DIDWW answers 423 with `Min-Expires: 600`.)
- A rejected credential (a repeated non-stale nonce, or 403) drops straight to the maximum backoff and logs at Error, rather than retrying every two seconds indefinitely.
- Registrations are withdrawn when shutdown starts, before the drain loop, so the provider stops routing new calls to a node that is going away. That needed a small hook in `pkg/service`; `service.NewService`'s signature is unchanged.

Two NAT details this has to get right, both learned the hard way downstream:

**`rport`.** Behind NAT our Via carries an address the registrar cannot reply to, so it has to ask the registrar to answer the source address and port it actually observed (RFC 3581). sipgo only adds a Via when the request has none, and the one it adds has no `rport`, so the registrant builds its own. Without this, registration never completes at all — the registrar answers into the void. The `received`/`rport` it echoes back is logged when it differs from the Contact we sent, which is the single most useful line for someone who is registered but not getting calls.

**Keepalive.** Providers commonly enforce a registration interval far longer than a NAT UDP mapping survives — 600s against a mapping usually dropped after 30s of silence. Without traffic, the mapping the provider recorded stops being reachable and inbound calls stop, while the registration still looks healthy. That is a worse failure than not registering, so I did not want to leave it purely as an operator concern: a bounded OPTIONS keepalive (25s by default, configurable, negative to disable) holds the mapping open, and an unanswered one is logged. It keeps running while a refresh is failing, because that is exactly when the mapping the provider is still calling needs to be held.

REGISTER also has to leave from the signaling socket, since the provider delivers to the mapping the REGISTER created. sipgo reuses the server listener for outbound UDP, but only once that listener is registered with the transport layer; sending earlier binds a separate socket which is then pooled for the registrar address permanently. So `startRegistrations` waits for the listener and fails startup if it never appears.

Registrations are per-node, and the README says to point one node at a given account. `livekit_sip_registrations_active` is exported per registrar, because losing a registration silently stops inbound calls.

## Tests

`pkg/sip/register_test.go` runs a real sipgo UAS as the registrar over UDP, and drives a real `registrant` through a real `sipgo.Client` on its own signaling socket. Covered: the digest handshake and that the digest URI is the Request-URI; `rport` on the Via and that the REGISTER leaves from the signaling port; a separate `auth_username`; the 401 and the 407 paths; a stale-nonce re-challenge; Min-Expires negotiation and its persistence across a refresh; refresh with a stable Call-ID and increasing CSeq; the granted expiry read from the returned Contact, the Expires header, or the shortest listed binding; un-REGISTER on shutdown, including a challenged one; keepalives continuing through a refresh failure and stopping once the binding lapses; a rejected password not looping; and `Stop` aborting a REGISTER in flight rather than waiting out the transaction timeout. `pkg/config/config_test.go` covers the config bounds.

`go test -race ./pkg/...` passes.

## Validating against a real carrier

Offered rather than claimed: this Go code has been tested against a local UAS, not yet against a carrier. We have DIDWW trunks in both Static Endpoint and SIP Registration mode, and we have already taken the equivalent path end to end downstream through a real NAT — REGISTER out, the carrier delivering the INVITE to the NAT-mapped port rather than the Contact literal, the call answered, two-way audio published into the room. If this is worth pursuing, I will run that same validation against this branch and post the trace.

Addresses #524. Scope is outbound registration only; accepting REGISTER as a registrar is not part of this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ Currently, the following features are supported:
 - Dialing Out (Sending INVITEs)
 - Dialing In (Accepting INVITEs)
 - Digest Authentication
+- Registering with a provider (Sending REGISTERs)
 - Touch Tone (Sending and Reading DTMF)
 
 ## Documentation
@@ -65,6 +66,40 @@ rtp_port: port to listen and send RTP traffic (default 10000-20000)
 ```
 
 The config file can be added to a mounted volume with its location passed in the SIP_CONFIG_FILE env var, or its body can be passed in the SIP_CONFIG_BODY env var.
+
+### Registering with a provider
+
+Providers usually deliver inbound calls to an address you give them, which requires the SIP
+service to be reachable at a stable IP. Where that is not possible, most of them offer a
+registration-based trunk instead: the service registers, and calls are delivered to the
+resulting binding. Configure one entry per account:
+
+```yaml
+sip_registrations:
+  - registrar: sip:sip.provider.example      # registrar address, or host[:port]
+    username: 1000                           # address-of-record user
+    password: secret
+    # auth_username: 1000                    # digest username, if issued separately
+    # domain: provider.example               # AOR host, defaults to the registrar host
+    # expiry: 10m                            # requested lifetime; the registrar may grant less
+    # keepalive: 25s                         # OPTIONS interval to hold the NAT binding open
+```
+
+Each node registers its own Contact, so point one node at a given account. Registration only
+tells the provider where to send calls: inbound INVITEs still arrive on the normal inbound path
+and are authenticated and dispatched by trunk as usual.
+
+The Contact is this node's signaling address, which behind NAT is a private one. That works
+with providers that route to the source address they observe, which is what the REGISTER asks
+them to do. If yours routes to the Contact instead, set `nat_1_to_1_ip` to the address it
+should use.
+
+`keepalive` matters because providers commonly refuse a registration interval short enough to
+refresh a NAT mapping on its own, and a mapping that expires stops inbound calls while the
+registration still looks healthy. Set it to a negative value to disable.
+
+Because losing a registration silently stops inbound calls, each one is also exported as
+`livekit_sip_registrations_active`, labeled by registrar.
 
 ### Using the SIP service
 
--- a/cmd/livekit-sip/main.go
+++ b/cmd/livekit-sip/main.go
@@ -104,6 +104,7 @@ func runService(ctx context.Context, c *cli.Command) error {
 		return err
 	}
 	svc := service.NewService(conf, log, sipsrv, sipsrv.Stop, sipsrv.ActiveCalls, psrpcClient, bus, mon)
+	svc.SetSIPServiceDrain(sipsrv.StopRegistrations)
 	sipsrv.SetHandler(svc)
 
 	if err = sipsrv.Start(); err != nil {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -51,6 +51,52 @@ const (
 	DefaultRTPDrainingDuration    = 10 * time.Minute // hard cap on how long a port stays draining
 )
 
+const (
+	// DefaultSIPRegistrationExpiry is the registration lifetime requested from the registrar.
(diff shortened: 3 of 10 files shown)


skyzer:
Thanks for the automated review — both findings were right and are fixed in 4eaf040:

- `registrations_active` is now labeled by address-of-record as well as registrar, so two accounts on one provider no longer share a series and mask each other.
- `refreshAfter` never returns less than half the granted lifetime. Subtracting a fixed 10s margin alone meant an 11s grant refreshed at 1s while a 10s grant refreshed at 5s; `TestRefreshAfter` now asserts the interval is monotonic in the lifetime.

Luskendeilder:
Hey @skyzer, this looks great — any idea if a maintainer's likely to review it soon? Been quiet for a few weeks and we're trying to decide whether to keep running it as a fork patch or wait it out.

PR #775, opened 2026-08-04
fix: cache participant attrs for BYE attributes_to_headers (#404)

### LINKED ISSUE #404: LiveKit attributes are not mapped to SIP X-* headers when sending BYE
I have configured an inbound trunk with an attributes_to_headers mapping. My goal is to set some custom headers in the BYE request when the call is ended. However if I look at the BYE headers from livekit-sip it does not contain the custom one specified by me. 

I have added some debugging code in inbound.go and it looks like the headers are not set by the following callback because the room is nil. 
```
cc := s.newInbound(LocalTag(callID), s.ContactURI(tr), req, tx, func(headers map[string]string) map[string]string {
        c := call
        if c == nil || len(c.attrsToHdr) == 0 {
                return headers
        }
        r := c.lkRoom.Room()
        if r == nil {
                return headers
        }
        headers = AttrsToHeaders(r.LocalParticipant.Attributes(), c.attrsToHdr, headers)
        fmt.Println("headers", headers)
        return headers
})
```

The livekit agent is hanging up by deleting the room. Is that the issue? Shouldn't the custom header be set anyway in the BYE request even if the room has been destroyed?

lixuanqun:
When an agent hangs up by deleting the room, LocalParticipant is gone before SIP sends BYE, so attributes_to_headers mapping was skipped and custom X-* headers were missing.

Cache the last-seen participant attributes on join/update and reuse them when building BYE/REFER headers if the room is already nil. Applies to inbound and outbound calls.

Fixes livekit/sip#404

### DIFF
--- a/pkg/sip/attrs_headers_test.go
+++ b/pkg/sip/attrs_headers_test.go
@@ -1,0 +1,79 @@
+// Copyright 2026 LiveKit, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sip
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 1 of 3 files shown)


genseric-ghiro:
Thanks for the fix @lixuanqun. I think it'd be a good idea to implement devin's suggestion for both inbound and outbound calls

lixuanqun:
Thanks @genseric-ghiro. Implemented Devin's finding on both inbound and outbound in `84eab15`: `snapshotParticipantAttrs` now skips the cache overwrite when the live read comes back empty — so a teardown-time empty read (room/participant already gone) no longer erases the values the cache exists to preserve for BYE. Regression tests added in `attrs_headers_test.go` for both call directions. CI re-running.

PR #777, opened 2026-08-06
Allow SIPServer to be configured with interceptors.

alexfish8:
Currently, the SIP server is configured so that each handler is automatically wrapped with a panic recovery interceptor. This PR removes the panic interceptors and allows SIP servers to be configured with arbitrary interceptors (so as to provide greater flexibility to users). Also, allow update `NewService` so that callers might specify options to be passed to the underlying server.


alexlivekit (approved):
lgtm

PR #780, opened 2026-08-07
fix(sip): answer re-INVITE with the RFC 3264 direction, not the cached SDP

### LINKED ISSUE #779: Hold re-INVITE (a=sendonly) is answered a=sendrecv, violating RFC 3264 §6.1 — carrier BYEs the dialog
# Hold re-INVITE (`a=sendonly`) is answered `a=sendrecv`, violating RFC 3264 §6.1 — carrier BYEs the dialog

## Summary

When a carrier places a call on hold it sends a re-INVITE offering `a=sendonly`. Both re-INVITE
paths in `pkg/sip/inbound.go` answer by replaying the **cached** local SDP verbatim, which still
carries `a=sendrecv`. RFC 3264 §6.1 requires a `sendonly` offer to be answered `recvonly`, so the
answer is invalid and the carrier tears the dialog down ~60 ms later.

On a bridged call (two SIP legs in one room, e.g. after a cold transfer) the BYE collapses the
whole call: dropping the held leg drops the other party too. The user-visible symptom is
**pressing HOLD hangs up the call**.

Reproduced on **v1.8.0** and **v1.9.0** against two different carrier SBCs.

## Reproduction

1. Establish a call with two SIP participants bridged in one room (inbound caller + outbound leg).
2. From the outbound party's phone, press **hold**.
3. The carrier sends a re-INVITE with `a=sendonly`; livekit-sip answers `a=sendrecv`; the carrier
   sends BYE and the call ends.

## Packet capture

`tcpdump -ni any udp port 5060` on the SIP pod. Phone numbers, IPs and call IDs are redacted;
SDP bodies and timings are verbatim.

```
06:19:05.464  IN   INVITE sip:203.0.113.10:5060 SIP/2.0     <- carrier: HOLD
06:19:05.465  OUT  SIP/2.0 200 OK                            <- our answer
06:19:05.521  IN   ACK
06:19:05.527  IN   BYE                                       <- carrier gives up (63 ms)
06:19:05.527  OUT  SIP/2.0 200 OK
06:19:05.552  OUT  BYE  -> other leg                          <- bridged call collapses
```

**Offer (carrier → us)** — note `a=sendonly`:

```
v=0
o=genband 1642851412 1894065156 IN IP4 203.0.113.20
s=-
c=IN IP4 203.0.113.20
t=0 0
m=audio 29076 RTP/AVP 0 8 18 9 13 101
c=IN IP4 203.0.113.20
a=rtpmap:0 PCMU/8000
a=rtpmap:8 PCMA/8000
a=rtpmap:101 telephone-event/8000
a=fmtp:101 0-15
a=sendonly
a=ptime:20
```

**Answer (us → carrier)** — `a=sendrecv`, which is what triggers the BYE:

```
v=0
o=- 9157931411869268290 9157931411869268294 IN IP4 203.0.113.10
s=LiveKit
c=IN IP4 203.0.113.10
t=0 0
m=audio 17830 RTP/AVP 0 101
a=rtpmap:0 PCMU/8000
a=rtpmap:101 telephone-event/8000
a=fmtp:101 0-16
a=ptime:20
a=sendrecv
```

Server-side log for the same moment:

```
06:19:05.464  processing invite
06:19:05.464  accepting reinvite      participant=<outbound-leg>
06:19:05.527  BYE from remote         participant=<outbound-leg>
06:19:05.528  Closing outbound call   participant=<outbound-leg>  reason=bye
```

## Cause

Both re-INVITE paths pass the cached local SDP straight through
(line numbers from `main` @ `4259552`):

- `pkg/sip/inbound.go:422` — inbound path: `cc.AcceptAsKeepAlive(existing.cc.OwnSDP())`
- `pkg/sip/inbound.go:434` — outbound path: `cc.AcceptAsKeepAlive(localSDP)`

`AcceptAsKeepAlive` (`inbound.go:2000`) responds `200 OK` with that body unmodified. The cached
SDP is the previously negotiated `sendrecv` answer, so the direction attribute is never adjusted
to the new offer.

Consistent with that, the repository contains **no occurrences** of `sendonly`, `recvonly`,
`sendrecv` or `inactive` in non-test Go source — RFC 3264 direction handling appears simply not to
be implemented for re-INVITEs.

## Expected behaviour (RFC 3264 §6.1)

| Offer | Correct answer |
|---|---|
| `a=sendonly` | `a=recvonly` |
| `a=recvonly` | `a=sendonly` |
| `a=inactive` | `a=inactive` |
| `a=sendrecv` / absent | `a=sendrecv` (implied, RFC 4566 §6) |

## Patch

We have a working fix running in production, plus tests. It adds two helpers and uses them in both
re-INVITE paths:

- `answerDirectionFor(offer []byte) string` — maps the offer's direction to the RFC-correct answer,
  returning `""` when the offer is `sendrecv` or states no direction.
- `withSDPDirection(local []byte, dir string) []byte` — replaces the single `a=` direction line in
  the cached SDP (appending one if absent), preserving CRLF.

Two deliberate choices, both open to discussion:

1. **Text-level rewrite rather than parse/re-serialize.** The cached SDP is an already-negotiated,
   working body; reserializing risks perturbing codec/`ptime`/connection lines the carrier has
   accepted. Only the direction attribute changes. If you'd rather this lived in the SDP layer
   (`media-sdk`) where parsing already happens, that's very likely the better long-term home and
   we're happy to redo it there.
2. **No-op when the offer has no direction**, so every non-hold re-INVITE (codec renegotiation,
   port change per #728, session-timer refresh) behaves exactly as today. This is the common case
   and we didn't want to change it.

Test coverage added (`pkg/sip/sdp_hold_direction_test.go`): the captured carrier SDP above, the
full direction mapping, the untouched-when-no-direction case, unhold, appending a missing direction
line, and collapsing duplicate direction lines. The existing `pkg/sip` suite passes unchanged.

**Not verified:** we've confirmed the SDP answer is now correct and that hold no longer ends the
call, but we have not characterised media behaviour across repeated hold/unhold cycles (whether
RTP flow is torn down and re-established as an implementation might expect). Worth a maintainer's
eye.

Happy to open a PR — wanted to check the preferred layer first.

## Environment

- livekit-sip **v1.8.0** and **v1.9.0** (both reproduce; digests `ad8dafcb…` and `a906ae85…`)
- Self-hosted on Kubernetes, `hostNetwork: true`, single replica
- Two carriers reproduce it; the offering SBC in the capture above identifies as `genband`

prasanth-33460:
Fixes #779.

## Problem

A hold re-INVITE offers `a=sendonly`. Both re-INVITE paths answer by replaying the **cached** local SDP verbatim, which still carries `a=sendrecv`. RFC 3264 §6.1 requires a `sendonly` offer to be answered `recvonly`, so the answer is invalid and the carrier tears the dialog down ~60 ms later.

On a bridged call (two SIP legs in one room) that BYE collapses the whole call — so pressing **hold** hangs up.

Captured on v1.8.0 and v1.9.0, against two carrier SBCs:

```
IN   INVITE  a=sendonly     <- carrier: hold
OUT  200 OK  a=sendrecv     <- invalid answer
IN   ACK
IN   BYE                    <- 63 ms later
```

The cached SDP is the previously negotiated `sendrecv` answer, so the direction attribute was never adjusted to the new offer:

- `pkg/sip/inbound.go:422` — inbound: `cc.AcceptAsKeepAlive(existing.cc.OwnSDP())`
- `pkg/sip/inbound.go:434` — outbound: `cc.AcceptAsKeepAlive(localSDP)`

Consistent with that, there were no occurrences of `sendonly` / `recvonly` / `sendrecv` / `inactive` in non-test source — direction handling wasn't implemented for re-INVITEs.

## Fix

Two helpers, used in both paths:

- **`answerDirectionFor(offer)`** — maps the offer's direction to the correct answer (`sendonly`→`recvonly`, `recvonly`→`sendonly`, `inactive`→`inactive`), returning `""` when the offer is `sendrecv` or states no direction.
- **`withSDPDirection(local, dir)`** — rewrites the single `a=` direction line in the cached SDP, appending one if absent, preserving CRLF.

### Two deliberate choices, both open to discussion

**1. Text-level rewrite rather than parse/re-serialize.** The cached SDP is an already-negotiated, working body; reserializing risks perturbing codec / `ptime` / connection lines the far end has accepted, so only the direction attribute changes.

If you'd rather this lived in the SDP layer (`media-sdk`) where parsing already happens, that's very likely the better long-term home — happy to redo it there.

**2. No-op when the offer states no direction** (`sendrecv` is implied, RFC 4566 §6), so every non-hold re-INVITE — codec renegotiation, port change (#728), session-timer refresh — behaves exactly as before. That's the common case and I didn't want to change it.

## Tests

`pkg/sip/sdp_hold_direction_test.go` covers the captured carrier SDP, the full direction mapping, the no-direction passthrough, unhold, appending a missing direction line, and collapsing duplicates. The existing `pkg/sip` suite passes unchanged.

## Not verified

The SDP answer is now correct and hold no longer ends the call (confirmed on a live trunk). I have **not** characterised media behaviour across repeated hold/unhold cycles — whether RTP flow is torn down and re-established as the implementation might expect. Worth a maintainer's eye.

### DIFF
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -349,6 +349,98 @@ func updateRemoteFromSDP(media *MediaPort, log logger.Logger, codecs *msdk.Codec
 	media.UpdateRemote(desc.Addr)
 }
 
+// ── RFC 3264 §6.1: ANSWER THE OFFER'S DIRECTION, DON'T ECHO OUR CACHED SDP ──────────
+//
+// Reproduced on v1.8.0 and v1.9.0. When a carrier puts a bridged call on hold it sends a
+// re-INVITE offering `a=sendonly`. Both the inbound and outbound re-INVITE paths answer by
+// replaying the CACHED local SDP verbatim, which carries `a=sendrecv`. RFC 3264 §6.1
+// requires a `sendonly` offer to be answered `recvonly` (and `inactive` -> `inactive`), so
+// the answer is invalid and the carrier tears the dialog down ~60ms after our 200 OK:
+//
+//	06:19:05.464  IN   INVITE  a=sendonly    <- carrier: hold
+//	06:19:05.465  OUT  200 OK  a=sendrecv    <- WRONG, must be a=recvonly
+//	06:19:05.521  IN   ACK
+//	06:19:05.527  IN   BYE                   <- carrier gives up
+//
+// The whole call then collapses: dropping the held leg drops the other party too, so a
+// supervisor pressing HOLD ends the customer's call. No application-layer workaround is
+// possible — the dialog is dead before the agent sees anything.
+//
+// answerDirectionFor maps the offer's direction to the RFC-correct answer, and
+// withSDPDirection rewrites the single a= line in our cached SDP. Deliberately a
+// text-level rewrite rather than a parse/re-serialize: the cached SDP is already a
+// negotiated, working body and reserializing it risks perturbing codecs/ptime/ICE lines
+// that the carrier has accepted. Only the direction attribute changes.
+//
+// If the offer states no direction, sendrecv is implied (RFC 4566 §6) and the cached SDP is
+// returned untouched — preserving today's behaviour for every non-hold re-INVITE (codec
+// renegotiation, port change, session-timer refresh).
+func answerDirectionFor(offer []byte) string {
+	// Scan media-level then session-level attributes; last one wins, matching how the
+	// direction applies to the (single) audio stream we negotiate.
+	dir := ""
+	for _, line := range strings.Split(string(offer), "\n") {
+		switch strings.TrimSpace(line) {
+		case "a=sendonly":
+			dir = "sendonly"
+		case "a=recvonly":
+			dir = "recvonly"
+		case "a=inactive":
+			dir = "inactive"
+		case "a=sendrecv":
+			dir = "sendrecv"
+		}
+	}
+	switch dir {
+	case "sendonly":
+		return "recvonly" // remote holds us: it sends, we only receive
+	case "recvonly":
+		return "sendonly"
+	case "inactive":
+		return "inactive"
+	default:
+		return "" // absent or sendrecv -> no rewrite needed
+	}
+}
+
+// withSDPDirection returns local with its direction attribute replaced by dir. If local has
+// no direction line, dir is appended to the audio media section (falling back to the end of
+// the body when no m= line is present).
+func withSDPDirection(local []byte, dir string) []byte {
+	if dir == "" || len(local) == 0 {
+		return local
+	}
+	// Preserve CRLF vs LF: SDP is CRLF per RFC 4566 and some SBCs are strict.
+	eol := "\n"
+	if strings.Contains(string(local), "\r\n") {
+		eol = "\r\n"
+	}
+	lines := strings.Split(strings.ReplaceAll(string(local), "\r\n", "\n"), "\n")
+	out := make([]string, 0, len(lines)+1)
+	replaced := false
+	for _, line := range lines {
+		switch strings.TrimSpace(line) {
+		case "a=sendrecv", "a=sendonly", "a=recvonly", "a=inactive":
+			if !replaced {
+				out = append(out, "a="+dir)
+				replaced = true
+			}
+			// drop any further duplicate direction lines
+		default:
+			out = append(out, line)
+		}
+	}
+	if !replaced {
+		// No direction line: insert after the audio m= section's attributes, i.e. at the
+		// end of the body. sendrecv was implied before, so stating it explicitly is safe.
+		for len(out) > 0 && strings.TrimSpace(out[len(out)-1]) == "" {
+			out = out[:len(out)-1]
+		}
+		out = append(out, "a="+dir, "")
+	}
+	return []byte(strings.Join(out, eol))
+}
+
 func (s *Server) onInvite(log *slog.Logger, req *sip.Request, tx sip.ServerTransaction) {
 	// Error processed in defer
 	_ = s.processInvite(req, tx)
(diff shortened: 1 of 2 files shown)


prasanth-33460:
The `test` failure here is a **pre-existing flake on `main`**, unrelated to this PR.

Failing test: `TestService_Interceptors` — added in #777 (`4259552`), which is also the base commit of this branch. This PR touches only `pkg/sip/inbound.go` (re-INVITE SDP direction) and adds `pkg/sip/sdp_hold_direction_test.go`; neither is reachable from that test.

## Reproduced on clean `main`, without this patch

```
$ git clone --depth 1 https://github.com/livekit/sip.git && cd sip
$ git log -1 --format='%h %s'
4259552 Allow SIPServer to be configured with interceptors. (#777)

$ go test -race ./pkg/sip/ -run TestService_Interceptors -count=5
--- FAIL: TestService_Interceptors (0.24s)
    testing.go:1712: race detected during execution of test
FAIL
```

## The race

`interceptorRecorder.calls` is written from the sipgo request-handling goroutine while the test goroutine reads it, with no synchronisation:

```
WARNING: DATA RACE
Write at 0x00c0002170a8 by goroutine 67:
  sip.(*interceptorRecorder).log()
      pkg/sip/service_test.go:454
  sip.TestService_Interceptors.(*interceptorRecorder).loggingInterceptor.func3.1()
      pkg/sip/service_test.go:444
  sipgo.(*Server).handleRequest()
      sipgo@v0.13.2/server.go:235

Previous read at 0x00c0002170a8 by goroutine 23:
  sip.TestService_Interceptors()
      pkg/sip/service_test.go:487
```

The duplicated `enter a` / `enter b` in the CI assertion diff (`listB` has 6 entries where `listA` expects 4) looks like the same unsynchronised slice being appended to concurrently rather than a behavioural difference.

A mutex around `interceptorRecorder.log`/read, or collecting via a channel, should settle it. Happy to include that fix here if you'd like, though it seems cleaner as its own change since it's independent of this one.

`Build` and `integration` both pass on this branch, and the 6 tests added here pass along with the rest of `pkg/sip` when `TestService_Interceptors` is excluded.

`license/cla` is still PENDING — let me know if there's a CLA to sign.

PR #786, opened 2026-08-11
TEL-886: Initial internal rework

alexlivekit:
Does not yet compile, working on it, but this should be the general shape.

### DIFF
--- a/pkg/sip/media_pipeline.go
+++ b/pkg/sip/media_pipeline.go
@@ -1,3 +1,17 @@
+// Copyright 2026 LiveKit, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package sip
 
 import (
--- a/pkg/sip/media_pipeline.go
+++ b/pkg/sip/media_pipeline.go
@@ -1,153 +1,167 @@
-// Copyright 2024 LiveKit, Inc.
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-//
-// 	http://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
-
 package sip
 
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
+	"math"
 	"os"
 	"strings"
+	"sync"
 	"sync/atomic"
-	"time"
 
 	msdk "github.com/livekit/media-sdk"
 	"github.com/livekit/media-sdk/dtmf"
 	"github.com/livekit/media-sdk/jitter"
 	"github.com/livekit/media-sdk/mixer"
 	"github.com/livekit/media-sdk/rtp"
+	"github.com/livekit/media-sdk/sdp"
 	"github.com/livekit/media-sdk/srtp"
+	"github.com/livekit/protocol/livekit"
 	"github.com/livekit/protocol/logger"
-	"github.com/livekit/protocol/utils/traceid"
+	"github.com/livekit/sip/pkg/stats"
 )
 
-func (p *mediaPort) SetConfig(c *MediaConf) error {
-	if p.closed.IsBroken() {
-		return errors.New("media is already closed")
-	}
+// A datastrucxture owning the implementation of everything between a udpConn and two output Switches
+// Has to two directions, with both audio and optionally DTMF for each
+// constructed once per negotiation, possibly N times in the lifetime of udpConn/Switch anchors.
+type mediaPortPipeline struct {
+	// Expected to be set by caller, not managed or owned
+	localSDP  []byte
+	log       logger.Logger
+	opts      *MediaOptions
+	mon       *stats.CallMonitor
+	stats     *PortStats
+	onNewSSRC func() bool
+	onPacket  func()
+
+	// Owned by pipeline
+	ctx               context.Context
+	cancel            context.CancelFunc
+	sess              rtp.Session
+	rtpLoopWG         sync.WaitGroup
+	muxToRoom         atomic.Pointer[rtp.HandlerCloser]
+	dtmfMixer         *mixer.Mixer
+	audioToRoom       rtp.HandlerCloser
+	dtmfToRoom        rtp.HandlerCloser
+	dtmfHandler       msdk.WriteCloser[*livekit.SipDTMF] // Merely a reference, we don't close this
+	audioToPort       msdk.PCM16Writer
+	dtmfToPort        msdk.WriteCloser[*livekit.SipDTMF]
+	lastDTMFTimestamp atomic.Uint32 // rtp timestamp of last DTMF packet seen
+}
+
+func (p *mediaPortPipeline) Configure(
+	mc *sdp.MediaConfig,
+	port *udpConn,
+	audioToRoom msdk.PCM16Writer,
+	dtmfToRoom msdk.WriteCloser[*livekit.SipDTMF],
+) (msdk.PCM16Writer, msdk.WriteCloser[*livekit.SipDTMF], error) {
+	p.ctx, p.cancel = context.WithCancel(context.Background())
+
 	var crypto string
-	if c.Crypto != nil {
-		crypto = c.Crypto.Profile.String()
+	if mc.Crypto != nil {
+		crypto = mc.Crypto.Profile.String()
 	}
 	p.log.Infow("using codecs",
-		"audio-codec", c.Audio.Codec.Info().SDPName, "audio-rtp", c.Audio.Type,
-		"dtmf-rtp", c.Audio.DTMFType,
+		"audio-codec", mc.Audio.Codec.Info().SDPName, "audio-rtp", mc.Audio.Type,
+		"dtmf-rtp", mc.Audio.DTMFType,
 		"srtp", crypto,
 	)
 
-	symmetric := p.opts.SymmetricRTP || (p.opts.IgnoreLocalAddrInSDP && c.Remote.Addr().IsPrivate())
-	p.port.SetDst(c.Remote)
-	if symmetric {
-		p.port.SetSymmetric(true)
+	port.SetDst(mc.Remote)
+	if p.opts.IgnoreLocalAddrInSDP && mc.Remote.Addr().IsPrivate() {
+		port.SetSymmetric(true) // Already initialized with opts, turn on for edge case
 	}
-	if p.opts.IgnorePreanswerData {
-		// this needs to happen before the SRTP session is created, otherwise the read deadline will be
-		// overwritten and we may get stuck in the discard loop
-		p.port.stopDiscarding()
-	}
-	var (
-		sess rtp.Session
-		err  error
-	)
-	if c.Crypto != nil {
-		sess, err = srtp.NewSession(p.log, p.port, c.Crypto)
+
+	// stopDiscarding() must be done ahead of SRTP session creation, or risk dead read without read deadlines
+	port.stopDiscarding()
+	p.lastDTMFTimestamp.Store(math.MaxUint32)
+
+	var err error
+	if mc.Crypto != nil {
+		p.sess, err = srtp.NewSession(p.log, port, mc.Crypto)
 	} else {
-		sess = rtp.NewSession(p.log, p.port)
+		p.sess = rtp.NewSession(p.log, port)
 	}
 	if err != nil {
-		return err
+		return nil, nil, fmt.Errorf("failed to setup pipeline: %w", err)
 	}
 
-	p.mu.Lock()
-	defer p.mu.Unlock()
-	p.conf = c
-	p.sess = sess
-
-	if err = p.setupOutput(p.tid); err != nil {
-		return err
+	err = p.setupInput(mc, audioToRoom, dtmfToRoom)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to setup pipeline: %w", err)
 	}
-	p.setupInput()
-	return nil
+	audioToPort, dtmfToPort, err := p.setupOutput(mc)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to setup pipeline: %w", err)
+	}
+
+	return audioToPort, dtmfToPort, nil
 }
 
-func (p *mediaPort) setupInput() {
-	// Decoding pipeline (SIP RTP -> LK PCM)
-	codec := p.conf.Audio.Codec
-	codecInfo := codec.Info()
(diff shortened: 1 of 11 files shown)


pkg/sip/media_pipeline.go
```diff
@@ -1,153 +1,167 @@
...
 	"github.com/livekit/protocol/logger"
-	"github.com/livekit/protocol/utils/traceid"
+	"github.com/livekit/sip/pkg/stats"
 )
 
-func (p *mediaPort) SetConfig(c *MediaConf) error {
-	if p.closed.IsBroken() {
-		return errors.New("media is already closed")
-	}
+// A datastrucxture owning the implementation of everything between a udpConn and two output Switches
```
alexfish8: datastrucxture -> data structure

pkg/sip/media_pipeline.go
```diff
@@ -1,153 +1,167 @@
...
+// Has to two directions, with both audio and optionally DTMF for each
```
alexfish8: to two -> two

pkg/sip/media_pipeline.go
```diff
@@ -17,137 +17,165 @@ package sip
...
+	dtmfMixer         *mixer.Mixer
+	audioToRoom       rtp.HandlerCloser
+	dtmfToRoom        rtp.HandlerCloser
+	dtmfHandler       msdk.WriteCloser[*livekit.SipDTMF] // Merely a reference, we don't close this
+	audioToPort       msdk.PCM16Writer
+	dtmfToPort        msdk.WriteCloser[*livekit.SipDTMF]
+	lastDTMFTimestamp atomic.Uint32 // rtp timestamp of last DTMF packet seen
+}
+
+func (p *mediaPortPipeline) Configure(
```
alexfish8: nit: Even though this is a method on a package-private struct, a comment describing what this function returns would be nice.

pkg/sip/media_pipeline.go
```diff
@@ -17,137 +17,165 @@ package sip
...
-		mux.Register(
-			p.conf.Audio.DTMFType, newRTPHandlerCount(
-				newRTPStatsHandler(p.mon, dtmf.SDPNameAndRate, rtp.HandlerFunc(p.dtmfHandler)),
-				&p.stats.DTMFPackets, &p.stats.DTMFBytes,
-			),
+	p.audioToRoom = audioType
+	mux.Register(mc.Audio.Type, audioType)
+
+	if mc.Audio.DTMFType != 0 {
+		p.dtmfHandler = dtmfToRoom // Close doesn't propagate through rtp.HandlerFunc
```
alexfish8, line 197: At first pass, I'm not sure if I understand the meaning behind the comment "Close doesn't propagate through rtp.HandlerFunc". Is this saying something like, "We need to maintain a separate handle to the dtmf handler so that we can close it later, as it won't be closed automatically"?

pkg/sip/media_pipeline.go
```diff
@@ -17,137 +17,165 @@ package sip
...
-func (p *mediaPort) dtmfHandler(h *rtp.Header, payload []byte) error {
-	ptr := p.dtmfIn.Load()
-	if ptr == nil {
-		return nil
-	}
-	fnc := *ptr
-	if fnc == nil {
-		return nil
-	}
+func (p *mediaPortPipeline) dtmfHandleFunc(h *rtp.Header, payload []byte) error {
```
alexfish8: naming nit: Thoughts on calling this something like `handleDTMF` or `handleDTMFEvent`? The Func suffix feels redundant (since it's not like this function returns something of type function).

pkg/sip/media_pipeline.go
```diff
@@ -158,92 +186,104 @@ func (p *mediaPort) dtmfHandler(h *rtp.Header, payload []byte) error {
...
+		Digit: string([]byte{ev.Digit}),
+	})
 }
 
-// Must be called holding the lock
-func (p *mediaPort) setupOutput(tid traceid.ID) error {
-	if p.closed.IsBroken() {
-		return errors.New("media is already closed")
-	}
+func (p *mediaPortPipeline) setupOutput(mc *sdp.MediaConfig) (msdk.PCM16Writer, msdk.WriteCloser[*livekit.SipDTMF], error) {
```
alexfish8: ditto here: Add a function comment that describes what this does and what this returns.

pkg/sip/media_pipeline.go
```diff
@@ -158,92 +186,104 @@ func (p *mediaPort) dtmfHandler(h *rtp.Header, payload []byte) error {
...
 		}
 		p.stats.Streams.Add(1)
-		p.mediaReceived.Break()
+		if p.onNewSSRC != nil {
+			p.onNewSSRC()
+		}
 		log := p.log.WithValues("ssrc", ssrc)
 		log.Debugw("accepting RTP stream")
-		go p.rtpReadLoop(tid, log, r)
+		p.rtpLoopWG.Add(1)
```
alexfish8: nit: ditto re preferring `WaitGroup.Go`.

pkg/sip/media_pipeline.go
```diff
@@ -158,92 +186,104 @@ func (p *mediaPort) dtmfHandler(h *rtp.Header, payload []byte) error {
...
 }
 
-// Must be called holding the lock
-func (p *mediaPort) setupOutput(tid traceid.ID) error {
-	if p.closed.IsBroken() {
-		return errors.New("media is already closed")
-	}
+func (p *mediaPortPipeline) setupOutput(mc *sdp.MediaConfig) (msdk.PCM16Writer, msdk.WriteCloser[*livekit.SipDTMF], error) {
+	// Encoding pipeline (LK PCM -> SIP RTP)
 	p.rtpLoopWG.Add(1)
```
alexfish8: nit: There's a [newish `WaitGroup.Go` method](https://pkg.go.dev/sync#WaitGroup.Go) that is a bit tidier to use than manually incrementing and decrementing wait group counters.

pkg/sip/media_pipeline.go
```diff
@@ -299,30 +337,83 @@ func (p *mediaPort) rtpReadLoop(tid traceid.ID, log logger.Logger, r rtp.ReadStr
...
-		pipeline = ""
 	}
 }
 
-func (p *mediaPort) WriteDTMF(ctx context.Context, digits string) error {
-	if len(digits) == 0 {
-		return nil
+func (p *mediaPortPipeline) Close() error {
+	if p.cancel != nil {
+		p.cancel() // unblocks a DTMF digit train before we tear the streams down
```
alexfish8: This comment puzzled me for a second until I checked the code and saw that the DTMF out writer does actually depend on p.ctx.

Might be worth being explicit about the DTMF output path depending on p.ctx?

pkg/sip/media_pipeline.go
```diff
@@ -299,30 +337,83 @@ func (p *mediaPort) rtpReadLoop(tid traceid.ID, log logger.Logger, r rtp.ReadStr
...
+	}
+	var errs []error
```
alexfish8, line 385: I like that you're collecting all of the errors and returning a collated error (instead of just the first error)  :).

pkg/sip/media_pipeline.go
```diff
@@ -299,30 +337,83 @@ func (p *mediaPort) rtpReadLoop(tid traceid.ID, log logger.Logger, r rtp.ReadStr
...
+type dtmfOutWriter struct {
+	mu           sync.Mutex
+	ctx          context.Context // canceled by pipeline Close, aborts an in-flight digit train
+	dtmfEvents   *rtp.Stream
+	dtmfAudio    msdk.PCM16Writer
+	getTimestamp func() uint32
+}
+
+func (w *dtmfOutWriter) String() string {
+	return fmt.Sprintf("dtmfOutWriter(dtmfAudio: %v)", w.dtmfAudio != nil)
```
alexfish8, line 421: Is `w.dtmfAudio != nil` intentional (rather than just passing in `w.dtmfAudio` for the format specifier argument)?

pkg/sip/media_port_negotiation_test.go
```diff
@@ -0,0 +1,290 @@
...
+		require.NoError(t, err)
+
+		var names []string
+		for _, c := range offer.Codecs {
+			if c.Codec == nil || (offer.DTMFType != 0 && c.Type == offer.DTMFType) {
+				continue
+			}
+			names = append(names, c.Codec.Info().SDPName)
+		}
+		require.Equal(t, []string{g711.ALawSDPNameAndRate}, names)
```
alexfish8: nit: (here and elsewhere) It'd be nicer to use assert.Equal and assert.NotZero here and on the line below so that both errors (if present) could be printed in the same test invocation.

pkg/sip/media_port_negotiation_test.go
```diff
@@ -0,0 +1,290 @@
...
+			base, err := m2.GenerateOffer()
+			require.NoError(t, err)
+			_, err = m1.GenerateAnswer([]byte(tc.hold(t, string(base))))
+			require.NoError(t, err)
+
+			// m1 no longer sends: no destination to write to, and the room-facing
+			// writer is detached. Room audio is dropped rather than erroring.
+			dst := m1.port.dst.Load()
+			require.NotNil(t, dst)
+			require.True(t, dst.Addr().IsUnspecified(), "held port kept a destination: %v", dst)
```
alexfish8: Prefer assert assert over require for these checks

pkg/sip/media_port_negotiation_test.go
```diff
@@ -0,0 +1,290 @@
...
+			// ...while m2's media still reaches us.
+			requireAudioFlows(t, m2, recv1)
+
+			// Resume with the original offer.
+			_, err = m1.GenerateAnswer(base)
+			require.NoError(t, err)
+
+			dst = m1.port.dst.Load()
+			require.NotNil(t, dst)
+			require.False(t, dst.Addr().IsUnspecified(), "destination not restored")
```
alexfish8: ditto (re preferring assert over require)

pkg/sip/media_port.go
```diff
@@ -386,23 +403,15 @@ type MediaPort interface {
...
-	return NewmediaPortWith(tid, log, mon, nil, opts, sampleRate)
+func NewMediaPort(log logger.Logger, mon *stats.CallMonitor, opts *MediaOptions, targetSampleRate int) (MediaPort, error) {
+	return NewMediaPortWith(log, mon, nil, opts, targetSampleRate)
 }
 
-func NewmediaPortWith(tid traceid.ID, log logger.Logger, mon *stats.CallMonitor, conn UDPConn, opts *MediaOptions, sampleRate int) (*mediaPort, error) {
+func NewMediaPortWith(log logger.Logger, mon *stats.CallMonitor, conn UDPConn, opts *MediaOptions, targetSampleRate int) (MediaPort, error) {
 	if opts == nil {
 		opts = &MediaOptions{}
 	}
```
alexfish8, line 413: From what I can tell, `opts` is passed into `mediaPort` directly during construction, meaning a caller that `mutates` `opts` could potentially affect any `mediaPort` constructed with those `opts`. Are we intentionally allowing the callers to do this sort of after-the-fact updating of configuration? If not can we create a copy of these opts before passing them in (and, ideally, `mediaPort.opts` would be a non-pointer type to reflect the fact that it can never be nil.

This PR is already really large, so feel free to leave a TODO for now if you'd like.

pkg/sip/media_port.go
```diff
@@ -638,45 +625,46 @@ func (p *mediaPort) timeoutLoop(timeoutCallback func()) {
...
 }
 
+func (p *mediaPort) closePipeline() {
```
alexfish8: nit: `closePipelineLocked` (to reinforce that this should only be called when the lock is called)?

pkg/sip/media_pipeline.go
alexfish8: Also, WDYT about renaming this to `Init()`? 

It looks like this method is only invoked a single time the struct is created, and, from my understanding, we'll never re-configure an existing mediaPortPipline, we'll just close the old one and create a new one? It's valid for a `mediaPort` to be re-configured, but, since mediaPortPipelines are just thrown away and re-created, having a different method name for `mediaPortPipeline` would further emphasize this difference in behavior.

pkg/sip/media_port.go
```diff
@@ -689,103 +677,184 @@ func (p *mediaPort) RemoteAddr() netip.AddrPort {
...
+		opts:      p.opts,
+		mon:       p.mon,
+		stats:     p.stats,
+		onNewSSRC: p.mediaReceived.Break,
+		onPacket:  p.onNewMediaPacket,
+	}
+	audioToPort, dtmfToPort, err := newPipeline.Configure(c, p.port, p.audioIn, &p.dtmfIn)
+	if err != nil {
+		// Close orphan pipeline, then error leads to mediaPort teardown
+		newPipeline.Close()
```
alexfish8: If a `mediaPortPipeline` fails to be configured, are we always just going to `Close` it? If so, why don't we just update the implementation of `Configure` to automatically close its resources so that the caller doesn't need to remember to do this? This would tighten the `mediaPortPipeline` interface and eliminate a potential footgun.

Also, assuming that `Configure` is only called a single time, it'd be nice if there were a constructor a la `newMediaPortPipeline` that created the `*mediaPortPipeline` struct and then invoked `Configure` internally, so that this calling code wouldn't have to touch the internals of the struct at all.

pkg/sip/media_port.go
```diff
@@ -689,103 +677,184 @@ func (p *mediaPort) RemoteAddr() netip.AddrPort {
...
+		return err
+	}
+	// Note: audioToPort & dtmfToPort are not propagating Close()
+	p.pipeline = newPipeline
+
+	if c.PeerDirection == psdp.DirectionSendOnly || c.Remote.Addr().IsUnspecified() {
+		// Older hold semantics: c=0.0.0.0; Newer hold semantics: a=sendonly
+		// TODO: Support a=recvonly/inactive; requires toggling media timeout;
+		//		maybe gate these on timers being actibe on the session to prevend dud calls
```
alexfish8: "being active"; "prevent"

pkg/sip/media_pipeline.go
alexlivekit: `rtp.HandlerFunc`, the type, doesn't get initialized with a close, and when it is closed, the signal gets swallowed up. So we need to close it manually.

pkg/sip/media_pipeline.go
alexlivekit: ~~?~~ Ah.

pkg/sip/media_pipeline.go
alexlivekit: Hmm...
```go
// Either
p.rtpLoopWG.Go(func() { p.rtpReadLoop(log, r) })
// Or
p.rtpLoopWG.Add(1)
go p.rtpReadLoop(log, r)
```
Can't say I see the extra tidiness, but it is a single line and has no chance of being separated, so probably the right way to go.

pkg/sip/media_pipeline.go
alexlivekit: yes. I don't need the serialization/stringification of w.dtmfAudio, we just want to know whether it's enabled or not.

pkg/sip/media_port.go
alexlivekit: AFAICT The idea of these opts is that they're read only, and the reason it's a pointer is to avoid allocation and copying. I'd say we save that convo for later.

pkg/sip/media_pipeline.go
alexlivekit: Sure.
I have had plans to call configure again (e.g. and if it's not dynamically adjustible without rebuilding it'd error out and cause a full rebuild), but we can rename it now and see about then when it happens (we might choose a dedicated method for that).

pkg/sip/media_port.go
alexlivekit: Done, but with a change.
NewMediaPortPipeline indeed ensures the close mechanic.
MediaPortPipelineConfig now exists an is used as-is.

alexfish8 (approved):

PR #788, opened 2026-08-13
Update inbound.go outbound.go to use new MediaPort interface.

alexfish8:
- Update the `MediaPort` interface to accommodate a few things within inbound.go
  - Add a `NegotiatedAudio` method to return the audio settings (if configured). We need this to update the `SIPCallInfo` in inbound.go and outbound.go
  - Add a `SetTimeoutForDelayedAck` method: see code comment for full details.
  -Added a few TODOs in places where we might need to send an updated SDP on re-INVITE.

### DIFF
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -404,26 +404,29 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 	s.cmu.RLock()
 	existing := s.byLocalTag[cc.ID()]
 	s.cmu.RUnlock()
-	if existing != nil && existing.cc.InviteCSeq() < cc.InviteCSeq() {
+	if existing != nil && existing.cc.InviteCSeq() < cc.InviteCSeq() && existing.media != nil {
 		existing.log().Infow("reinvite", "content-length", req.ContentLength(), "cseq", cc.InviteCSeq())
 		if _, err = existing.media.GenerateAnswer(sdpBodyFromRequest(req)); err != nil {
 			log.Errorw("failed to update outbound call SDP", err)
+			return err
 		}
 
+		// TODO(rework): Reply with the new SDP.
 		cc.AcceptAsKeepAlive(existing.cc.OwnSDP())
 		return nil
 	}
 	if s.cli != nil { // Process reinvite for existing outbound calls
 		oc := s.cli.getActiveCall(cc.ID())
 		newCSeq := cc.InviteCSeq()
-		if oc != nil && oc.cc != nil && oc.cc.InviteCSeq() < newCSeq {
+		if oc != nil && oc.cc != nil && oc.cc.InviteCSeq() < newCSeq && oc.media != nil {
 			if localSDP, err := oc.media.GetLocalSDP(); err == nil && len(localSDP) > 0 {
 				oc.log.Infow("accepting reinvite", "content-length", req.ContentLength(), "cseq", cc.InviteCSeq())
 				if _, err = oc.media.GenerateAnswer(sdpBodyFromRequest(req)); err != nil {
 					log.Errorw("failed to update outbound call SDP", err)
-					return nil
+					return err
 				}
 				oc.cc.RecordInvite(newCSeq)
+				// TODO(rework): Reply with the new SDP.
 				cc.AcceptAsKeepAlive(localSDP)
 				return nil
 			}
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -73,6 +73,7 @@ const (
 var allowHeader = sip.NewHeader("Allow", "INVITE, ACK, CANCEL, BYE, NOTIFY, REFER, MESSAGE, OPTIONS, INFO, SUBSCRIBE")
 
 var errNoACK = errors.New("no ACK received for 200 OK")
+var errInternal = errors.New("internal error")
 
 // hashPassword creates a SHA256 hash of the password for logging purposes
 func hashPassword(password string) string {
@@ -417,19 +418,25 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 	if s.cli != nil { // Process reinvite for existing outbound calls
 		oc := s.cli.getActiveCall(cc.ID())
 		newCSeq := cc.InviteCSeq()
-		if oc != nil && oc.cc != nil && oc.cc.InviteCSeq() < newCSeq && oc.media != nil {
-			localSDP, err := oc.media.GetLocalSDP() // eror
-			if err == nil && len(localSDP) > 0 {
-				oc.log.Infow("accepting reinvite", "content-length", req.ContentLength(), "cseq", cc.InviteCSeq())
-				if err := oc.updateRemoteFromSDP(sdpBodyFromRequest(req)); err != nil {
-					log.Errorw("failed to update outbound call SDP", err)
-					return err
-				}
-				oc.cc.RecordInvite(newCSeq)
-				// TODO(alexfish): Reply with the new SDP.
-				cc.AcceptAsKeepAlive(localSDP)
-				return nil
+		if oc != nil && oc.cc != nil && oc.cc.InviteCSeq() < newCSeq {
+			if oc.media == nil {
+				oc.log.Errorw("outbound call media has not been negotiated", nil)
+				return errInternal
 			}
+			localSDP, err := oc.media.GetLocalSDP()
+			if err != nil || len(localSDP) == 0 {
+				oc.log.Errorw("outbound call does not have an SDP", nil)
+				return errInternal
+			}
+			oc.log.Infow("accepting reinvite", "content-length", req.ContentLength(), "cseq", cc.InviteCSeq())
+			if err := oc.updateRemoteFromSDP(sdpBodyFromRequest(req)); err != nil {
+				log.Errorw("failed to update outbound call SDP", err)
+				return errInternal
+			}
+			oc.cc.RecordInvite(newCSeq)
+			// TODO(alexfish): Reply with the new SDP.
+			cc.AcceptAsKeepAlive(localSDP)
+			return nil
 		}
 
 	}
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -417,9 +417,9 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 	if s.cli != nil { // Process reinvite for existing outbound calls
 		oc := s.cli.getActiveCall(cc.ID())
 		newCSeq := cc.InviteCSeq()
-		if oc != nil && oc.cc != nil && oc.cc.InviteCSeq() < newCSeq {
-			localSDP := oc.GetLocalSDP()
-			if len(localSDP) > 0 {
+		if oc != nil && oc.cc != nil && oc.cc.InviteCSeq() < newCSeq && oc.media != nil {
+			localSDP, err := oc.media.GetLocalSDP() // eror
+			if err == nil && len(localSDP) > 0 {
 				oc.log.Infow("accepting reinvite", "content-length", req.ContentLength(), "cseq", cc.InviteCSeq())
 				if err := oc.updateRemoteFromSDP(sdpBodyFromRequest(req)); err != nil {
 					log.Errorw("failed to update outbound call SDP", err)
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -404,33 +404,34 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 	s.cmu.RLock()
 	existing := s.byLocalTag[cc.ID()]
 	s.cmu.RUnlock()
-	if existing != nil && existing.cc.InviteCSeq() < cc.InviteCSeq() && existing.media != nil {
+	if existing != nil && existing.cc.InviteCSeq() < cc.InviteCSeq() {
 		existing.log().Infow("reinvite", "content-length", req.ContentLength(), "cseq", cc.InviteCSeq())
-		if _, err = existing.media.GenerateAnswer(sdpBodyFromRequest(req)); err != nil {
-			log.Errorw("failed to update outbound call SDP", err)
+		if err := existing.updateRemoteFromSDP(sdpBodyFromRequest(req)); err != nil {
+			log.Errorw("failed to update inbound call SDP", err)
 			return err
 		}
-
-		// TODO(rework): Reply with the new SDP.
+		// TODO(alexfish): Reply with the new SDP.
 		cc.AcceptAsKeepAlive(existing.cc.OwnSDP())
 		return nil
 	}
 	if s.cli != nil { // Process reinvite for existing outbound calls
 		oc := s.cli.getActiveCall(cc.ID())
 		newCSeq := cc.InviteCSeq()
-		if oc != nil && oc.cc != nil && oc.cc.InviteCSeq() < newCSeq && oc.media != nil {
-			if localSDP, err := oc.media.GetLocalSDP(); err == nil && len(localSDP) > 0 {
+		if oc != nil && oc.cc != nil && oc.cc.InviteCSeq() < newCSeq {
+			localSDP := oc.GetLocalSDP()
+			if len(localSDP) > 0 {
 				oc.log.Infow("accepting reinvite", "content-length", req.ContentLength(), "cseq", cc.InviteCSeq())
-				if _, err = oc.media.GenerateAnswer(sdpBodyFromRequest(req)); err != nil {
+				if err := oc.updateRemoteFromSDP(sdpBodyFromRequest(req)); err != nil {
 					log.Errorw("failed to update outbound call SDP", err)
 					return err
 				}
 				oc.cc.RecordInvite(newCSeq)
-				// TODO(rework): Reply with the new SDP.
+				// TODO(alexfish): Reply with the new SDP.
 				cc.AcceptAsKeepAlive(localSDP)
 				return nil
 			}
 		}
+
 	}
 
 	from, to := cc.From(), cc.To()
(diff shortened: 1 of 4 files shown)


pkg/sip/inbound.go
```diff
@@ -406,18 +406,23 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
...
+			if localSDP, err := oc.media.GetLocalSDP(); err == nil && len(localSDP) > 0 {
 				oc.log.Infow("accepting reinvite", "content-length", req.ContentLength(), "cseq", cc.InviteCSeq())
-				oc.updateRemoteFromSDP(sdpBodyFromRequest(req))
+				if _, err = oc.media.GenerateAnswer(sdpBodyFromRequest(req)); err != nil {
+					log.Errorw("failed to update outbound call SDP", err)
+					return nil
+				}
 				oc.cc.RecordInvite(newCSeq)
 				cc.AcceptAsKeepAlive(localSDP)
 				return nil
```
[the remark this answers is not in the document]
alexfish8: Added a todo.

pkg/sip/media_port.go
```diff
@@ -858,6 +887,8 @@ func (p *mediaPort) configure(c *sdp.MediaConfig, localSDP []byte) error {
 	}
 	p.offer = nil // Pipeline build done, can now proceed to offer anew
 	p.localSDP = localSDP
+	p.audioConf = &c.Audio
```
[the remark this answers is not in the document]
alexfish8: @alexlivekit : This (the timeout being armed when the media is set up rather than when the call is excepted) was something that I thought originally motivated having a new method on `MediaPort` (something like `Start` or just `StartTimeout`), which would give the client control as to when to start the timeout.

Is this something you think we need to worry about? Ostensibly, after media is set up, it seems like the call should be accepted relatively quickly (i.e., it doesn't seem like it would take a long time for a the LiveKit participant and track to be set up, which would motivate us not adding another interface method).

pkg/sip/outbound.go
```diff
@@ -510,16 +508,37 @@ func (c *outboundCall) dialSIP(ctx context.Context, tid traceid.ID) error {
...
 
+	var errs []error
+
 	if digits := c.sipConf.dtmf; digits != "" {
 		c.setStatus(CallAutomation)
 		// Write initial DTMF to SIP
-		if err := c.media.WriteDTMF(ctx, digits); err != nil {
-			return err
+		dtmfWriter := c.media.GetDTMFWriter()
+		for i := range len(digits) {
```
alexfish8: Instead of doing this loop, it's probably better to just update the media pipeline dtmf out writer to accept digit strings of arbitrary lengths.

pkg/sip/media_port.go
```diff
@@ -394,11 +394,43 @@ type MediaPort interface {
 	GetDTMFWriter() msdk.WriteCloser[*livekit.SipDTMF]
 	WriteDTMFTo(w msdk.WriteCloser[*livekit.SipDTMF]) msdk.WriteCloser[*livekit.SipDTMF]
 
+	// GenerateOffer returns an encoded SDP offer. Returns an error if an offer
+	// has been already been generated or media has already been negotiated.
```
alexlivekit: If there is no offer, this generates an offer.
If there is an offer, this simply returns the SDP of that offer.
An offer is cleared once a negotiation is successful.

pkg/sip/media_port.go
```diff
@@ -394,11 +394,43 @@ type MediaPort interface {
...
 	GenerateOffer() ([]byte, error)
+
+	// GenerateAnswer returns an encoded SDP answer for the given offer.
+	//
+	// SIDE EFFECT: Rebuilds the media pipeline.
```
alexlivekit:
```suggestion
	// SIDE EFFECT: May cause a rebuild of the pipeline.
```
nit: While it's worth calling this out, this is not a necessary part of the interface.

pkg/sip/media_port.go
```diff
@@ -394,11 +394,43 @@ type MediaPort interface {
...
 	GenerateAnswer(offer []byte) ([]byte, error)
+
+	// ProcessAnswer processes an encoded SDP answer from the remote client. Returns an
+	// error if the answer is invalid, the offer has not yet been generated, or
+	// if media has already been negotiated.
```
alexlivekit, line 409:
```
// Note: May cause a rebuild of the pipeline.
```
nit: Suggesting this here as well.

pkg/sip/media_port.go
```diff
@@ -394,11 +394,43 @@ type MediaPort interface {
...
+
+	// SetTimeout resets the media timeout with the given values.
+	//
+	// NOTE: This method will be removed at some point in the future and is only
+	// used by one caller. Do not call this method: specify media timeout
+	// settings when constructing a MediaPort.
+	//
+	// We would like to enforce the fact that all invites must be ACKed.
+	// Sometimes, though, for whatever reason, we do not see ACKs for invites
+	// (presumably, this is due to an issue within sipfe). In such cases where
```
alexlivekit:
```suggestion
	// (e.g due to possible issues with load balancing). In such cases where
```

pkg/sip/media_port.go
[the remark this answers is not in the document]
alexlivekit: Two separate things:
1. sip-wise, it is possible to negotiate media well ahead of the call being answered (via INVITE-183). Implementation-wise, we do not currently support that.
2. Regardless of the above, and even if/when we choose to support early media, starting the media timeout as soon as media is fully negotiated is the right thing imo. When we finish setting up a pipeline, that basically means the local endpoint has both offer and answer, and the remote endpoint either already has both of them, or can be reasonably expected to within a short amount of time. If we do not see media for `media_timeout` time (typically > sip rtt), it should fire.

pkg/sip/media_port.go
```diff
@@ -858,6 +887,8 @@ func (p *mediaPort) configure(c *sdp.MediaConfig, localSDP []byte) error {
 	}
 	p.offer = nil // Pipeline build done, can now proceed to offer anew
 	p.localSDP = localSDP
+	p.audioConf = &c.Audio
+	p.enableTimeoutWithDefaults()
```
alexlivekit, line 893: good catch!

pkg/sip/outbound.go
alexlivekit: Yeah, that's the right idea. Lets add a TODO in media pipeline to support multi-digit payloads and drop these loops in favor of a single call.

pkg/sip/outbound.go
```diff
@@ -510,16 +508,37 @@ func (c *outboundCall) dialSIP(ctx context.Context, tid traceid.ID) error {
...
+			if err := dtmfWriter.WriteSample(&livekit.SipDTMF{
+				Digit: string(digits[i]),
+			}); err != nil {
+				errs = append(errs, fmt.Errorf("error writing digit no. %d (%s): %w", i, string(digits[i]), err))
```
alexlivekit: In this case I think we should break early. If you have an error it means one of two things - 
1. Something went wrong with the pipeline (e.g. closed), which would mean repeated errors.
2. Something is wrong with a single digit (idk if that is even possible), but it still renders the point of this feature moot - if I need to get to a meeting with a pin, sending 1245 instead of the intended 12345 is a broken call, and imo we should error out immediately without completing the rest.

pkg/sip/outbound.go
```diff
@@ -737,8 +756,6 @@ func (c *outboundCall) sipSignal(ctx context.Context, tid traceid.ID) error {
 		return err
 	}
 
-	mc.Processor = c.c.handler.GetMediaProcessor(c.sipConf.enabledFeatures, c.sipConf.featureFlags, string(c.cc.ID()), MediaProcessorOpts{InputSampleRate: c.media.InputSampleRate()})
```
alexlivekit, line 740: Lets either not remove this, or add a TODO to reinstall this. This is important.

pkg/sip/outbound.go
```diff
@@ -756,22 +773,38 @@ func (c *outboundCall) sipSignal(ctx context.Context, tid traceid.ID) error {
 	}
 
 	c.setExtraAttrs(c.sipConf.headersToAttrs, c.sipConf.includeHeaders, c.cc, nil)
-	c.state.DeferUpdate(func(info *livekit.SIPCallInfo) {
-		info.AudioCodec = mc.Audio.Codec.Info().SDPName
-		if r := c.lkRoom.Room(); r != nil {
-			info.ParticipantAttributes = r.LocalParticipant.Attributes() // clones
-		}
-	})
+	if audio := c.media.NegotiatedAudio(); audio != nil {
```
alexlivekit:
```suggestion
	if audio := c.media.NegotiatedAudio(); audio == nil {
	  return ... // error
	}
	... // insides of current `if` block.
```
We fully expect the media to be negotiated at this point.

pkg/sip/outbound.go
```diff
@@ -756,22 +773,38 @@ func (c *outboundCall) sipSignal(ctx context.Context, tid traceid.ID) error {
...
+				info.ParticipantAttributes = r.LocalParticipant.Attributes() // clones
+			}
+		})
+
+	}
 	return nil
 }
 
-func (c *outboundCall) handleDTMF(ev dtmf.Event) {
+func (c *outboundCall) handleDTMF(msg *livekit.SipDTMF) {
```
alexlivekit, line 803: Note we already receive `msg *livekit.SipDTMF`, we should just forward it to room. Or, better yet, completely skip the call itself (by doing `port.WriteDTMFTo(room.GetDTMFWriter())`).

pkg/sip/inbound.go
```diff
@@ -406,23 +406,32 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
...
 		cc.AcceptAsKeepAlive(existing.cc.OwnSDP())
 		return nil
 	}
 	if s.cli != nil { // Process reinvite for existing outbound calls
 		oc := s.cli.getActiveCall(cc.ID())
 		newCSeq := cc.InviteCSeq()
-		if oc != nil && oc.cc != nil && oc.cc.InviteCSeq() < newCSeq {
-			localSDP := oc.cc.LocalSDP()
-			if len(localSDP) != 0 {
+		if oc != nil && oc.cc != nil && oc.cc.InviteCSeq() < newCSeq && oc.media != nil {
```
alexlivekit: This is adding a skip state in `oc.media == nil`. This should not be possible. Error out if that happens, and proceed without the skip condition.

pkg/sip/inbound.go
```diff
@@ -406,23 +406,32 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
...
+			localSDP, err := oc.media.GetLocalSDP() // eror
+			if err == nil && len(localSDP) > 0 {
```
alexlivekit: ditto

pkg/sip/inbound.go
```diff
@@ -1122,35 +1153,30 @@ func (c *inboundCall) runMediaConn(tid traceid.ID, offerData []byte, mconf *sipM
...
 
 	answerData, err := mp.GenerateAnswer(offerData)
 	if err != nil {
 		return nil, err
 	}
+
 	c.mon.SDPSize(len(answerData), false)
 	c.log().Debugw("SDP answer", "sdp", string(answerData))
 
-	mc.Processor = c.s.handler.GetMediaProcessor(features, featureFlags, string(c.cc.ID()), MediaProcessorOpts{InputSampleRate: c.media.InputSampleRate()})
```
alexlivekit, line 1135: Same as with the outbound case - should not be removed.

pkg/sip/inbound.go
```diff
@@ -1122,35 +1153,30 @@ func (c *inboundCall) runMediaConn(tid traceid.ID, offerData []byte, mconf *sipM
...
-
-	if mc.Audio.DTMFType != 0 {
-		mp.HandleDTMF(c.handleDTMF)
-	}
+	mp.WriteDTMFTo(&dtmfEventWriter{handler: c.handleDTMF})
```
alexlivekit, line 1178: Same as with the outbound case: ideally we simply connect the room to media, skipping this awkward step.

pkg/sip/inbound.go
```diff
@@ -1122,35 +1153,30 @@ func (c *inboundCall) runMediaConn(tid traceid.ID, offerData []byte, mconf *sipM
...
 
 	// Must be set earlier to send the pin prompts.
-	if w := c.lkRoom.SwapOutput(mp.GetAudioWriter()); w != nil {
+	if w := c.lkRoom.SwapOutput(c.audioOut); w != nil {
 		_ = w.Close()
 	}
-	if mc.Audio.DTMFType != 0 {
-		c.lkRoom.SetDTMFOutput(mp.GetDTMFWriter())
+	c.lkRoom.SetDTMFOutput(c.media.GetDTMFWriter())
```
alexlivekit, line 1184: nit: I'm not sure I'm fully processing Go's typed nil pointer, but we need to make sure this doesn't die. Sounds like a TODO.

pkg/sip/inbound.go
```diff
@@ -1122,35 +1153,30 @@ func (c *inboundCall) runMediaConn(tid traceid.ID, offerData []byte, mconf *sipM
...
+
+	if audio := mp.NegotiatedAudio(); audio != nil {
```
alexlivekit: ditto on `if audio != nil `

pkg/sip/inbound.go
```diff
@@ -1484,6 +1510,20 @@ func (c *inboundCall) Shutdown(ctx context.Context) {
...
+	var mp MediaPort
+
+	c.mmu.Lock()
+	mp = c.media
+	c.mmu.Unlock()
+
+	if mp == nil {
+		return nil
+	}
+	_, err := mp.GenerateAnswer(body)
```
alexlivekit, line 1538: Note, this is something we need to fix on the branch before landing - `mp.GenerateAnswer(body)` will completely overwrite the local sdp, deviating from the way it works today.
This is exactly how we want to operate long term, but for now we'd need to align.

pkg/sip/inbound.go
```diff
@@ -1599,17 +1639,33 @@ func (c *inboundCall) playAudio(ctx context.Context, frames []msdk.PCM16Sample)
 	_ = msdk.PlayAudio[msdk.PCM16Sample](ctx, t, rtp.DefFrameDur, frames)
 }
 
-func (c *inboundCall) handleDTMF(tone dtmf.Event) {
+func (c *inboundCall) handleDTMF(msg *livekit.SipDTMF) {
```
alexlivekit, line 1661: ditto, no real need for this.

pkg/sip/outbound.go
```diff
@@ -150,8 +150,6 @@ func (c *Client) newCall(ctx context.Context, tid traceid.ID, conf *config.Confi
 		})
 		return nil, err
 	}
-	call.media.EnableTimeout(false)
-	call.media.DisableOut() // disabled until we get 200
```
alexlivekit, line 154: Might need to turn into a TODO, or done. Can't just drop that.

pkg/sip/inbound.go
```diff
@@ -1122,35 +1153,30 @@ func (c *inboundCall) runMediaConn(tid traceid.ID, offerData []byte, mconf *sipM
 	}
 	c.media = mp
 	c.mediaCodecs = mconf.Codecs
-	mp.EnableTimeout(false) // enabled once we accept the call
-	mp.DisableOut()         // disabled until we send 200
```
alexlivekit, line 1126: Dropping `mp.EnableTimeout(false)` is good here, but lets make sure `mp.DisableOut()` functionality is still there.

pkg/sip/inbound.go
alexfish8: Good catch, fixed!

pkg/sip/inbound.go
alexfish8: Fixed.

pkg/sip/outbound.go
alexfish8: Good call, fixed.

pkg/sip/inbound.go
alexfish8: In `inboundCall`, we maintain `audioOut` (which we wire up to the room and only enable wire up to the media output when the call is accepted).

pkg/sip/inbound.go
alexfish8: Good call, done.

pkg/sip/inbound.go
alexfish8: Added back the functionality.

pkg/sip/outbound.go
alexfish8: Added back the functionality.

pkg/sip/media_port.go
alexfish8: Good call, done.

pkg/sip/media_port.go
alexfish8: Done.

pkg/sip/media_port.go
alexfish8: Done.

pkg/sip/media_port.go
alexfish8: Done.

pkg/sip/outbound.go
alexfish8: Done.

pkg/sip/outbound.go
alexfish8: Fixed.

pkg/sip/inbound.go
alexfish8: Good call. Leaving this in, as I believe your PR #791 will handle this.

pkg/sip/inbound.go
alexfish8: Double checked: `media.GetDTMFWriter()` always returns non-nil (specifically, it returns the address of some non-pointer struct field).

alexfish8 (commented):
Thanks for the review!

alexfish8:
I haven't yet addressed the DTMF handler cleanups. Do you mind if I do this in a subsequent PR?

pkg/sip/inbound.go
alexfish8: Adding a TODO for now.

alexlivekit (approved):

PR #793, opened 2026-08-14
Clean up writer interfaces.

alexfish8:
- Rename `MediaSession` methods to be more explicit about the direction of a particular writer
  - [My brain kept getting confused about directions, especially when I was tweaking `Room` to satisfy the `MediaSession` interface, and so, for my own sanity, I renamed these things. Let me know what you think, I can always swap the names back or we can pick some thing else if you'd prefer]
 - Update `Room` method names to be similar to those in `MediaPort`
 - Move some DTMF event conversion logic out of inbound/outbound and into room.
   - Note: I couldn't entirely remove the `dtmfEventWriter` adapter out of inbound.go, as, even though Room's methods have been updated to the new pattern, the adapter is still responsible for populating some inbound-internal DTFM channel that gets read when validating pin prompts.

### DIFF
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -45,7 +45,6 @@ import (
 	lksip "github.com/livekit/protocol/sip"
 	"github.com/livekit/protocol/utils/traceid"
 	"github.com/livekit/psrpc"
-	lksdk "github.com/livekit/server-sdk-go/v2"
 	"github.com/livekit/sipgo/sip"
 
 	"github.com/livekit/sip/pkg/config"
@@ -1113,9 +1112,8 @@ func (c *inboundCall) waitForCallEnd(ctx context.Context, ackReceived <-chan str
 	}
 }
 
-// TODO(alexfish): Update Room so that we don't need this adapater.
 type dtmfEventWriter struct {
-	handler func(msg *livekit.SipDTMF)
+	c *inboundCall
 }
 
 func (w *dtmfEventWriter) String() string {
@@ -1130,9 +1128,29 @@ func (w *dtmfEventWriter) Close() error {
 	return nil
 }
 
-func (w *dtmfEventWriter) WriteSample(sample *livekit.SipDTMF) error {
-	w.handler(sample)
+func (w *dtmfEventWriter) WriteSample(msg *livekit.SipDTMF) error {
+	if msg == nil {
+		return nil
+	}
+
+	if w.c.forwardDTMF.Load() {
+		if err := w.c.lkRoom.GetInboundDTMFWriter().WriteSample(msg); err != nil {
+			w.c.log().Errorw("failed to send dtmf to room", err)
+		}
+	}
+
+	// TODO(alexfish): Just have the channel accept the message instead?
+	event := dtmfEventFromSipDTMF(msg)
+	// We should have enough buffer here.
+	select {
+	case w.c.dtmf <- event:
+	default:
+	}
 	return nil
+}
+
+func (c *inboundCall) getNewInboundDTMFWriter() msdk.WriteCloser[*livekit.SipDTMF] {
+	return &dtmfEventWriter{c}
 }
 
 func (c *inboundCall) runMediaConn(tid traceid.ID, offerData []byte, mconf *sipMediaConfig, conf *config.Config, featureFlags map[string]string) ([]byte, error) {
@@ -1173,13 +1191,11 @@ func (c *inboundCall) runMediaConn(tid traceid.ID, offerData []byte, mconf *sipM
 	c.mon.SDPSize(len(answerData), false)
 	c.log().Debugw("SDP answer", "sdp", string(answerData))
 
-	mp.WriteInboundDTMFTo(&dtmfEventWriter{handler: c.handleDTMF})
+	mp.WriteInboundDTMFTo(c.getNewInboundDTMFWriter())
 
 	// Must be set earlier to send the pin prompts.
-	if w := c.lkRoom.SwapOutput(c.audioOut); w != nil {
-		_ = w.Close()
-	}
-	c.lkRoom.SetDTMFOutput(c.media.GetOutboundDTMFWriter())
+	c.lkRoom.WriteOutboundAudioTo(c.audioOut)
+	c.lkRoom.WriteOutboundDTMFTo(c.media.GetOutboundDTMFWriter())
 
 	audio := mp.NegotiatedAudio()
 	if audio == nil {
@@ -1607,16 +1623,16 @@ func (c *inboundCall) createLiveKitParticipant(ctx context.Context, rconf RoomCo
 
 func (c *inboundCall) publishTrack(features []livekit.SIPFeature, featureFlags map[string]string) error {
 	defer c.mon.StageDurTimer("track-publish")()
-	local, err := c.lkRoom.NewParticipantTrack(RoomSampleRate)
+	inboundAudio, err := c.lkRoom.GetInboundAudioWriter()
 	if err != nil {
 		_ = c.lkRoom.Close()
 		return err
 	}
 
 	if audioInProcessor := c.s.handler.GetMediaProcessor(features, featureFlags, string(c.cc.ID()), MediaProcessorOpts{InputSampleRate: RoomSampleRate}); audioInProcessor != nil {
-		local = audioInProcessor(local)
+		inboundAudio = audioInProcessor(inboundAudio)
 	}
-	c.media.WriteInboundAudioTo(local)
+	c.media.WriteInboundAudioTo(inboundAudio)
 	return nil
 }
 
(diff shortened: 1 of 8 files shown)


pkg/sip/room.go
```diff
@@ -797,6 +779,45 @@ func (r *Room) NewTrack() *mixer.Input {
 	return r.mix.NewInput()
 }
 
+func (r *Room) WriteOutboundAudioTo(w msdk.PCM16Writer) msdk.PCM16Writer {
+	return r.outboundAudio.Swap(w)
```
genseric-ghiro, line 783: We don't need `r.mix.SampleRate()` anymore?

pkg/sip/inbound.go
```diff
@@ -1174,13 +1191,11 @@ func (c *inboundCall) runMediaConn(tid traceid.ID, offerData []byte, mconf *sipM
...
 
-	mp.WriteDTMFTo(&dtmfEventWriter{handler: c.handleDTMF})
+	mp.WriteInboundDTMFTo(c.getNewInboundDTMFWriter())
 
 	// Must be set earlier to send the pin prompts.
-	if w := c.lkRoom.SwapOutput(c.audioOut); w != nil {
-		_ = w.Close()
-	}
-	c.lkRoom.SetDTMFOutput(c.media.GetDTMFWriter())
+	c.lkRoom.WriteOutboundAudioTo(c.audioOut)
```
genseric-ghiro: Do we not need to close the previous audio anymore?

pkg/sip/media_port.go
```diff
@@ -391,10 +391,15 @@ func (o *MediaOptions) ApplyDefaults() {
...
+	// GetOutboundAudioWriter returns the LK room -> SIP writer.
+	GetOutboundAudioWriter() msdk.PCM16Writer
+	// GetOutboundDTMFWriter returns the LK room -> SIP DTMF writer.
+	GetOutboundDTMFWriter() msdk.WriteCloser[*livekit.SipDTMF]
+
+	// WriteInboundAudioTo tells the MediaSegment where to write inbound SIP audio.
+	WriteInboundAudioTo(w msdk.PCM16Writer) msdk.PCM16Writer
+	// WriteInboundDTMFTo tells the MediaSegment where to write inbound SIP DTMF.
+	WriteInboundDTMFTo(w msdk.WriteCloser[*livekit.SipDTMF]) msdk.WriteCloser[*livekit.SipDTMF]
 }
```
alexlivekit: This doesn't make much sense then as a separate interface.

The current convention w/r/t inbound and outbound directions in SIP is with respect to the telephony side - inbound == from sip/tel to room, and outbound is the other way.
The original was abstracted without the directions deliberately to allow it to apply to both, but yes it is confusing to follow and reason about.
However, now with the updated names this can only apply to the media port.

pkg/sip/inbound.go
```diff
@@ -1131,11 +1128,31 @@ func (w *dtmfEventWriter) Close() error {
...
+func (w *dtmfEventWriter) WriteSample(msg *livekit.SipDTMF) error {
+	if msg == nil {
+		return nil
+	}
+
+	if w.c.forwardDTMF.Load() {
+		if err := w.c.lkRoom.GetInboundDTMFWriter().WriteSample(msg); err != nil {
+			w.c.log().Errorw("failed to send dtmf to room", err)
+		}
+	}
```
alexlivekit:
```go
 } else {
  event := dtmfEventFromSipDTMF(msg)
  ...
}
```
Presumably, when we require a pin, there is no room yet, and when there is a room we're not waiting for a pin, right?

pkg/sip/inbound.go
```diff
@@ -1131,11 +1128,31 @@ func (w *dtmfEventWriter) Close() error {
...
 
-func (w *dtmfEventWriter) WriteSample(sample *livekit.SipDTMF) error {
-	w.handler(sample)
+func (w *dtmfEventWriter) WriteSample(msg *livekit.SipDTMF) error {
+	if msg == nil {
+		return nil
+	}
+
+	if w.c.forwardDTMF.Load() {
+		if err := w.c.lkRoom.GetInboundDTMFWriter().WriteSample(msg); err != nil {
```
alexlivekit: nit: The writer here should be fixed for the lifetime of dtmfEventWriter. I'd do something like this: 
```go
func (c *inboundCall) getNewInboundDTMFWriter() msdk.WriteCloser[*livekit.SipDTMF] {
	return &dtmfEventWriter{c, c.lkRoom.GetInboundDTMFWriter()}
}

type dtmfEventWriter struct {
	c *inboundCall
	w msdk.WriteCloser[*livekit.SipDTMF]
}

// and here:
	if w.c.forwardDTMF.Load() {
		if err := w.w.WriteSample(msg); err != nil {
		...
```

pkg/sip/inbound.go
```diff
@@ -1174,13 +1191,11 @@ func (c *inboundCall) runMediaConn(tid traceid.ID, offerData []byte, mconf *sipM
 	c.mon.SDPSize(len(answerData), false)
 	c.log().Debugw("SDP answer", "sdp", string(answerData))
 
-	mp.WriteDTMFTo(&dtmfEventWriter{handler: c.handleDTMF})
+	mp.WriteInboundDTMFTo(c.getNewInboundDTMFWriter())
```
alexlivekit: +1 To Genseric's comment - Lets either close the return code, or warn if it's not nil.

pkg/sip/inbound.go
```diff
@@ -1658,30 +1673,32 @@ func (c *inboundCall) playAudio(ctx context.Context, frames []msdk.PCM16Sample)
...
 		digit = dtmf.CodeToChar(code)
 	}
-	event := dtmf.Event{
+	return dtmf.Event{
 		Code:  code,
 		Digit: digit,
 	}
+}
+
+func (c *inboundCall) handleDTMF(msg *livekit.SipDTMF) {
```
alexlivekit: sorry, what is calling this now? I thought you want inbound dtmf to go from media_port to inboundDTMFWriter, which hash a virtual carbon copy of this code.

pkg/sip/outbound.go
```diff
@@ -545,16 +542,14 @@ func (c *outboundCall) updateRemoteFromSDP(body []byte) error {
 }
 
 func (c *outboundCall) connectMedia() {
-	if w := c.lkRoom.SwapOutput(c.audioOut); w != nil {
-		_ = w.Close()
-	}
-	c.lkRoom.SetDTMFOutput(c.media.GetDTMFWriter())
+	c.lkRoom.WriteOutboundAudioTo(c.audioOut)
+	c.lkRoom.WriteOutboundDTMFTo(c.media.GetOutboundDTMFWriter())
```
alexlivekit: ditto. (comment or close)

pkg/sip/room.go
```diff
@@ -797,6 +779,45 @@ func (r *Room) NewTrack() *mixer.Input {
...
+}
+
+func (r *Room) WriteOutboundDTMFTo(w msdk.WriteCloser[*livekit.SipDTMF]) msdk.WriteCloser[*livekit.SipDTMF] {
+	return r.outboundDTMF.Swap(w)
+}
+
+func (r *Room) GetInboundAudioWriter() (msdk.PCM16Writer, error) {
+	return r.NewParticipantTrack(RoomSampleRate)
```
alexlivekit, line 791: Not sure I like this one. Would we not create a new track every call? (where we simply return the fixed writer in other places, including GetInboundDTMFWriter below)

pkg/sip/inbound.go
alexlivekit: Actually, the even better thing imo would be to do this:
- Media port instantiate, `mp.WriteInboundDTMFTo(pinDTMFWriter{c.dtmf})`. This will always forward to the internal channel.
- When we're good with the prompt and ready to connect to room, `mp.WriteInboundDTMFTo(r.GetInboundDTMFWriter())`

pkg/sip/media_port.go
alexfish8: Removed the interface.

pkg/sip/inbound.go
alexfish8: Forgot to remove, deleting.

pkg/sip/inbound.go
alexfish8: I like this a lot better, updated.

pkg/sip/inbound.go
alexfish8: Using two distinct inbound DTMF writers for the different phases as discussed.

pkg/sip/inbound.go
alexfish8: There shouldn't be anything there, as this is the first time we're connecting the room's audio to the MediaPort. Adding a warning log to make this self-evident.

pkg/sip/inbound.go
alexfish8: Adding a warning.

pkg/sip/outbound.go
alexfish8: Added some warning logs.

pkg/sip/room.go
alexfish8: This is mirroring the existing behavior (where, previously, inbound and outbound would get new inbound writers by invoking `NewParticipantTrack`. We could have a fixed writer stored on Room, but we'd probably want to lazily initialize it the first time `GetInboundAudioWriter` is called. IMO, that doesn't really seem that much better.

alexfish8 (commented):
Thanks for the reviews!

pkg/sip/room.go
alexfish8: We're now assuming that this audio writer's sample rate will just be the default room rate.

Also, w.r.t. these writer types, the convention has changed a bit, s.t. all of these NewWriteCloserSwitches have fixed sample rates, and, if, internally, they get a new writer that has a different sampling rate than the old one, a panic occurs. The motivation here was just to eliminate any unnecessary resampling.

alexlivekit (approved):
LGTM!

PR #796, opened 2026-08-16
Fix DTMF deduplication to use (timestamp, event code) composite key

cuihang:
## Problem

The DTMF handler in `dtmfHandler` (media_port.go) deduplicated incoming DTMF packets using **only** the RTP timestamp:

```go
if h.Timestamp == p.lastDTMFTimestamp.Load() {
    return nil
}
```

RFC 4733 requires all packets of a given digit to share identical timestamps, so this correctly filters redundant packets within one digit. However, **some SIP devices or carriers reuse the timestamp of the previous digit when sending the next digit**. When this happens, the next digit is incorrectly filtered out and never reported.

## Fix

Replace the timestamp-only dedup with a **(timestamp, event code)** composite key:

```go
ev, err := dtmf.Decode(payload)
if err != nil {
    return nil
}
eventID := uint64(h.Timestamp)<<8 | uint64(ev.Code)
if eventID == p.lastDTMFEvent.Load() {
    return nil
}
p.lastDTMFEvent.Store(eventID)
fnc(ev)
```

### Behavior

| Scenario | Before | After |
|---|---|---|
| Same digit, redundant packets (same ts + same code) | Deduped ✓ | Deduped ✓ |
| Different digits sharing timestamp (same ts + different code, e.g. `*`, `0`, `1`) | **Incorrectly dropped** ✗ | Each reported ✓ |
| No dependency on RTP marker bit | ✓ | ✓ |

## Known limitation

If an upstream sends two identical digits (e.g. two `0`s) with the exact same timestamp **and** event code, the `(timestamp, event code)` key alone cannot distinguish them. Resolving that case requires full DTMF lifecycle state tracking (marker bit, End bit, duration reset, sequence number gaps). That is a more complex enhancement beyond this fix; in the worst case (all fields identical) the receiver cannot distinguish the digits at all.

## Changes

- `pkg/sip/media_port.go`: Replace `lastDTMFTimestamp atomic.Uint32` with `lastDTMFEvent atomic.Uint64`; update `dtmfHandler` to decode first, then dedup on composite key.
- `pkg/sip/media_port_test.go`: Update field reference; add `TestMediaPortDTMFSameTimestamp` verifying two different digits (`0` and `1`) sharing the same RTP timestamp are both reported.

## Test plan

- [x] `TestMediaPortDTMF` — all 12 existing subtests pass (digits 1/12/123 × loss none/first/last/middle)
- [x] `TestMediaPortDTMFSameTimestamp` — new test passes: digits `0` and `1` with same timestamp → both reported as `01`

### DIFF
--- a/pkg/sip/media_pipeline.go
+++ b/pkg/sip/media_pipeline.go
@@ -148,16 +148,19 @@ func (p *MediaPort) dtmfHandler(h *rtp.Header, payload []byte) error {
 	if fnc == nil {
 		return nil
 	}
-	// RFC 4733 requires all packets of a given digit to share identical timestamps.
-	// The marker bit could be used instead, but it is prone to occasional loss.
-	if h.Timestamp == p.lastDTMFTimestamp.Load() {
-		return nil
-	}
 	ev, err := dtmf.Decode(payload)
 	if err != nil {
 		return nil
 	}
-	p.lastDTMFTimestamp.Store(h.Timestamp)
+	// RFC 4733 requires all packets of a given digit to share identical timestamps.
+	// Some SIP devices or carriers may reuse the timestamp of the previous digit
+	// for the next one, so we combine timestamp and event code for deduplication.
+	// The marker bit could be used instead, but it is prone to occasional loss.
+	eventID := uint64(h.Timestamp)<<8 | uint64(ev.Code)
+	if eventID == p.lastDTMFEvent.Load() {
+		return nil
+	}
+	p.lastDTMFEvent.Store(eventID)
 	fnc(ev)
 	return nil
 }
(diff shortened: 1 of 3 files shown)


cuihang:
Hi @dennwc , just wanted to check if someone could take a look at this PR. We've hit this DTMF loss issue in production with our SIP carrier. All CI checks pass and the fix is minimal. Happy to make any adjustments if needed.

dennwc (approved):
Looks good to me! Thank you for the fix!

cuihang:
Hi @dennwc, just a quick update — I've rebased onto the latest main to resolve the merge conflicts (the dtmf handler had moved to `media_pipeline.go`). All CI checks are now passing. The core fix logic remains the same as what you reviewed. Could you take another look when you get a chance?

dennwc (approved):

PR #799, opened 2026-08-19
TEL-886: Prep for reconfigure, re-introduce timeout control

alexlivekit:
- Prep for reconfigure: Introduced `changeSetSummary`, to allow skipping reconfig on changes that do not require it. Code paths still disabled.
- Re-introduce timeout control: One major deviation from main was found to be in the wait for room connection path: In new code, we would process invite, generate answer, and arm timeout immediately. However, we would not send the 200 back until room is entered. This would run the risk of timing out on media before remote could possibly even know where to send it. As a temp measure, we have renamed `SetTimeoutForDelayedAck` to `SetTimeout`, allowed to process offer without arming timeout, and added a call site to `SetTimeout` right after sending an INVITE-200 for inbound calls.

### DIFF
--- a/pkg/sip/media_port.go
+++ b/pkg/sip/media_port.go
@@ -860,11 +860,6 @@ func (p *mediaPort) configure(c *sdp.MediaConfig, localSDP []byte) error {
 
 	changeSetSummary := NewChangeSetSummary(p.negotiated, c)
 
-	// Explicitly disable renegotiation for now
-	if changeSetSummary.includes(changeSetAudioCodec | changeSetDTMF | changeSetCrypto) {
-		return SDPError{Err: ErrRenegotiationDisabled} // Results in a 400 response
-	}
-
 	if changeSetSummary.includes(changeSetLocalAddr) {
 		return errors.New("unexpected local address change")
 	}
@@ -874,6 +869,10 @@ func (p *mediaPort) configure(c *sdp.MediaConfig, localSDP []byte) error {
 	hold := false
 
 	if changeSetSummary.shouldReconfigure() {
+		if changeSetSummary != changeSetNew { // Explicitly disable renegotiation for now
+			return SDPError{Err: ErrRenegotiationDisabled} // Results in a 400 response
+		}
+
 		p.closePipelineLocked()
 		p.port.stopDiscarding() // Needs readDeadline. Must be ahead of Reopen() and NewMediaPortPipeline()
 		p.port.Reopen()         // Allow reads from socket again
--- a/pkg/sip/media_port.go
+++ b/pkg/sip/media_port.go
@@ -32,6 +32,7 @@ import (
 	msdk "github.com/livekit/media-sdk"
 	"github.com/livekit/media-sdk/rtp"
 	"github.com/livekit/media-sdk/sdp"
+	"github.com/livekit/media-sdk/srtp"
 	"github.com/livekit/mediatransportutil/pkg/rtcconfig"
 	"github.com/livekit/protocol/livekit"
 	"github.com/livekit/protocol/logger"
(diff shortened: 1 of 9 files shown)


pkg/sip/media_port.go
```diff
@@ -920,26 +930,69 @@ func (p *mediaPort) configure(c *sdp.MediaConfig, localSDP []byte) error {
 		p.log.Infow("peer requested hold", "direction", c.PeerDirection.String(), "remote", c.Remote.String())
 	}
```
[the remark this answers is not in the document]
alexlivekit: Same as main, deliberately done to maintain parity.

pkg/sip/media_port.go
```diff
@@ -920,26 +930,69 @@ func (p *mediaPort) configure(c *sdp.MediaConfig, localSDP []byte) error {
...
+		}
+	} else { // Prodile exists on both
+		if a.Profile != b.Profile ||
+			!bytes.Equal(a.Keys.LocalMasterKey, b.Keys.LocalMasterKey) ||
+			!bytes.Equal(a.Keys.LocalMasterSalt, b.Keys.LocalMasterSalt) ||
+			!bytes.Equal(a.Keys.RemoteMasterKey, b.Keys.RemoteMasterKey) ||
+			!bytes.Equal(a.Keys.RemoteMasterSalt, b.Keys.RemoteMasterSalt) {
+			changeSetSummary |= changeSetCrypto
+		}
+	}
```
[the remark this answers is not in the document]
alexlivekit: Same as main today.

pkg/sip/media_port.go
```diff
@@ -920,26 +944,69 @@ func (p *mediaPort) configure(c *sdp.MediaConfig, localSDP []byte) error {
...
+func NewChangeSetSummary(current, new *sdp.MediaConfig) changeSetSummary {
+	if current == nil {
+		return changeSetNew
+	}
+	var changeSetSummary changeSetSummary
+	if current.Audio.Codec.Info().SDPName != new.Audio.Codec.Info().SDPName {
+		changeSetSummary |= changeSetAudioCodec
+	}
+	if current.Audio.DTMFType != new.Audio.DTMFType || current.Audio.Type != new.Audio.Type {
+		changeSetSummary |= changeSetDTMF
```
alexfish8, line 978: Should we call this `changeSetAudio` instead?

pkg/sip/service_test.go
```diff
@@ -49,11 +48,28 @@ func getResponseOrFail(t *testing.T, tx sip.ClientTransaction) *sip.Response {
 
 	return nil
 }
+func getResponseOrFailTimeout(t *testing.T, ctx context.Context, tx sip.ClientTransaction) *sip.Response {
+	var ctxDone <-chan struct{} = nil
```
alexfish8, line 53: t.Helper()?

pkg/sip/media_port.go
```diff
@@ -868,48 +872,68 @@ func (p *mediaPort) configure(c *sdp.MediaConfig, localSDP []byte) error {
...
-	audioToPort, dtmfToPort := newPipeline.GetConnectors() // These are not propagating Close()
-	p.pipeline = newPipeline
+		audioToPort, dtmfToPort = newPipeline.GetConnectors() // These are not propagating Close()
+		p.pipeline = newPipeline
 
-	if c.PeerDirection == psdp.DirectionSendOnly || c.Remote.Addr().IsUnspecified() {
-		// Older hold semantics: c=0.0.0.0; Newer hold semantics: a=sendonly
+		p.localSDP = localSDP // TODO: Move to end of function when reconfiguring is supported
+		p.reportPeerCodecs(p.offer.MediaDesc)
+	} else if changeSetSummary.includes(changeSetRemoteAddr) {
```
alexfish8: Should this be `if` and not `else if`? i.e., if the pipeline is reconfigured and the remote address has changed, will rebuilding the pipeline update `p.port`?

alexfish8 (approved):
Looks good! Just a few small comments

pkg/sip/media_port.go
alexlivekit: imo `changeSetAudioCodec` is accurate, but I will move `|| current.Audio.Type != new.Audio.Type` to the proper spot now that you mention it!

pkg/sip/media_port.go
alexlivekit: > will rebuilding the pipeline update p.port?

Yes, it will.

However, it is such a minor thing that we should probably drop this little tie-in to `media_pipeline`'s implementation.

PR #800, opened 2026-08-20
TEL-886: Done reviving tests

alexlivekit:


pkg/sip/media_port.go
```diff
@@ -778,6 +778,7 @@ func (p *mediaPort) GenerateAnswer(offerData []byte, activateTimeout bool) ([]by
 	if err != nil {
 		return nil, SDPError{Err: err}
 	}
+	p.reportPeerCodecs(offer.MediaDesc)
```
alexlivekit, line 781: Only on offer, regardless of whether it went though. As originally designed.

pkg/sip/media_pipeline_test.go
```diff
@@ -522,3 +526,72 @@ func TestMediaPipelineReuseUDPConn(t *testing.T) {
...
+				for _, digitPackets := range packets {
+					sendPackets := dropPackets(t, lossPackets, digitPackets)
+					t.Logf("sending %d/%d packets", len(sendPackets), len(digitPackets))
+					for _, pkt := range sendPackets {
+						h := pkt.Header
+						t.Logf("sending packet: seq=%d, ts=%d, marker=%t", h.SequenceNumber, h.Timestamp, h.Marker)
+						require.NoError(t, p.handleEventRTP(&h, pkt.Payload))
+					}
+				}
+				t.Logf("sent: %s", digits)
```
alexfish8, line 591: nit: remove these lines (since, we'd probably only want to log in case of test failure, in which case they'd be printed out anyways from the assertion) ?

pkg/sip/media_port_negotiation_test.go
```diff
@@ -318,3 +343,91 @@ func TestMediaPortHold(t *testing.T) {
...
+						}
+					})
+				}
+			})
+		}
+	}
+
+	t.Run("inbound", func(t *testing.T) { // Receive offer
+		negotiate := func(t *testing.T, mp *mediaPort, policy sdp.Encryption) (*sdp.MediaConfig, error) {
+			t.Helper()
```
alexfish8: nit: here and elsewhere, I think we should omit this, as otherwise, the line number of the assertion failure will be considered to be 376 (in foreach) instead of each of these functions.

pkg/sip/media_port_test.go
```diff
@@ -348,38 +379,227 @@ func newMediaPairWithAddr(t testing.TB, ip1, ip2 netip.Addr, opt1, opt2 *MediaOp
...
+							for i := range packetSize {
+								to2[i] = int16(amp1 * math.Sin(freq*2*math.Pi*float64(i)/float64(packetSize)))
+								to1[i] = int16(amp2 * math.Sin(freq*2*math.Pi*float64(i)/float64(packetSize)))
+							}
+
+							codecConfig := codecConfigMap[info.SDPName] // defaults to 0,0
+
+							// Ramp-up time for the codec.
+							// Some codecs have "inertia" and cannot immediately represent the sound exactly.
+							// This is shy we write signal multiple times to give it some time to adapt.
```
alexfish8: "shy" -> "why"

alexfish8 (approved):
lgtm! Small nits

pkg/sip/media_pipeline_test.go
alexlivekit: In the OG tests, and these are not visible if the tests didn't fail.

PR #802, opened 2026-08-20
TEL-886: Media Port rework

alexlivekit:
This change is intended to:
- Have little functional differences between it an main.
  - the one known thing: Plug-in processor moved back to ingesting 48khz audio and returning 48khz audio.
- Add interface for clarity.
- Add machinery to support renegotiation, early media, hold, etc. All these are still disabled for now, but there.
- Add tests.

### DIFF
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -953,8 +953,6 @@ func (c *inboundCall) handleInvite(ctx context.Context, tid traceid.ID, req *sip
 		ackTimeout  <-chan time.Time
 	)
 
-	roomAudioOut := msdk.NewWriteCloserSwitch[msdk.PCM16Sample](RoomSampleRate)
-
 	acceptCall := func(answerData []byte) (bool, error) {
 		defer c.mon.StageDurTimer("call-accept")()
 		headers := disp.Headers
@@ -985,8 +983,13 @@ func (c *inboundCall) handleInvite(ctx context.Context, tid traceid.ID, req *sip
 			// Start this timer right after the Accept.
 			ackTimeout = time.After(inviteOkAckLateTimeout)
 		}
-		if old := roomAudioOut.Swap(c.media.GetOutboundAudioWriter()); old != nil {
-			c.log().Warnw("unexpected audio out writer", nil)
+		// Attach room outputs
+		if old := c.lkRoom.WriteOutboundAudioTo(c.media.GetOutboundAudioWriter()); old != nil {
+			c.log().Warnw("room has unexpected outbound audio writer", nil)
+			old.Close()
+		}
+		if old := c.lkRoom.WriteOutboundDTMFTo(c.media.GetOutboundDTMFWriter()); old != nil {
+			c.log().Warnw("room has unexpected outbound audio DTMF writer", nil)
 			old.Close()
 		}
 		if ok, err := c.waitMedia(ctx); !ok {
@@ -996,7 +999,7 @@ func (c *inboundCall) handleInvite(ctx context.Context, tid traceid.ID, req *sip
 		return true, nil
 	}
 
-	if err := c.createMediaPort(mconf, conf, roomAudioOut, disp.FeatureFlags); err != nil {
+	if err := c.createMediaPort(mconf, conf, disp.FeatureFlags); err != nil {
 		return rejectMedia(err)
 	}
 
@@ -1160,7 +1163,7 @@ func (w *pinDTMFWriter) WriteSample(msg *livekit.SipDTMF) error {
 	return nil
 }
 
-func (c *inboundCall) createMediaPort(mconf *sipMediaConfig, conf *config.Config, roomAudioOut *msdk.WriteCloserSwitch[msdk.PCM16Sample], featureFlags map[string]string) error {
+func (c *inboundCall) createMediaPort(mconf *sipMediaConfig, conf *config.Config, featureFlags map[string]string) error {
 	c.mmu.Lock()
 	defer c.mmu.Unlock()
 	if c.media != nil {
@@ -1190,15 +1193,8 @@ func (c *inboundCall) createMediaPort(mconf *sipMediaConfig, conf *config.Config
 	c.media = mp
 	c.mediaCodecs = mconf.Codecs
 
-	// Mixer is created with the room; attach it now so pin prompts can play.
-	if old := c.lkRoom.WriteOutboundAudioTo(roomAudioOut); old != nil {
-		c.log().Warnw("room has unexpected outbound audio writer", nil)
-		old.Close()
-	}
-	if old := c.lkRoom.WriteOutboundDTMFTo(c.media.GetOutboundDTMFWriter()); old != nil {
-		c.log().Warnw("room has unexpected outbound audio DTMF writer", nil)
-		old.Close()
-	}
+	// Do not attach room outputs yet, we dont necessarily want it plumbed yet
+
 	return nil
 }
 
@@ -1737,22 +1733,23 @@ func (c *inboundCall) transferCall(ctx context.Context, transferTo string, heade
 	if dialtone && c.started.IsBroken() && !c.done.Load() {
 		const ringVolume = math.MaxInt16 / 2
 
-		// Mute the room audio to the SIP participant.
-		oldRoomAudioOut := c.lkRoom.WriteOutboundAudioTo(nil)
+		c.lkRoom.WriteOutboundAudioTo(nil) // Mute room
 
+		c.mmu.Lock()
+		mp := c.media
+		c.mmu.Unlock()
+		if mp == nil {
+			return transferID, fmt.Errorf("media port not found")
+		}
+		oldRoomAudioOut := mp.GetOutboundAudioWriter()
 		defer func() {
 			if retErr != nil && !c.done.Load() {
 				c.lkRoom.WriteOutboundAudioTo(oldRoomAudioOut)
-			} else if oldRoomAudioOut != nil {
-				if err := oldRoomAudioOut.Close(); err != nil {
-					c.log().Warnw("failed to close old audio output", err)
-				}
 			}
 		}()
 
 		rctx, rcancel := context.WithCancel(ctx)
 		defer rcancel()
-
 		go func() {
 			err := tones.Play(rctx, oldRoomAudioOut, ringVolume, tones.ETSIRinging)
 			if err != nil && !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -422,8 +422,7 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 		return nil
 	}
 	if s.cli != nil { // Process reinvite for existing outbound calls
-		// TODO(alexfish): Move this logic into outbound.go after the refactor
-		// has landed.
+		// TODO(alexfish): Consider moving this to outbound
 		oc := s.cli.getActiveCall(cc.ID())
 		newCSeq := cc.InviteCSeq()
 
@@ -1733,17 +1732,14 @@ func (c *inboundCall) transferCall(ctx context.Context, transferTo string, heade
 	if dialtone && c.started.IsBroken() && !c.done.Load() {
 		const ringVolume = math.MaxInt16 / 2
 
-		// Mute the room audio to the SIP participant.
-		if old := c.lkRoom.WriteOutboundAudioTo(nil); old != nil {
-			old.Close()
-		}
-
 		c.mmu.Lock()
 		mp := c.media
 		c.mmu.Unlock()
 		if mp == nil {
 			return transferID, fmt.Errorf("media port not found")
 		}
+		// Mute the room audio to the SIP participant.
+		_ = c.lkRoom.WriteOutboundAudioTo(nil) // Not closing mp anchor
 		defer func() {
 			if retErr != nil && !c.done.Load() {
 				c.lkRoom.WriteOutboundAudioTo(mp.GetOutboundAudioWriter())
@@ -1753,8 +1749,7 @@ func (c *inboundCall) transferCall(ctx context.Context, transferTo string, heade
 		rctx, rcancel := context.WithCancel(ctx)
 		defer rcancel()
 		go func() {
-			ctx := rctx
-			err := tones.Play(ctx, mp.GetOutboundAudioWriter(), ringVolume, tones.ETSIRinging)
+			err := tones.Play(rctx, mp.GetOutboundAudioWriter(), ringVolume, tones.ETSIRinging)
 			if err != nil && !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
 				c.log().Infow("cannot play dial tone", "error", err)
 			}
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -737,7 +737,6 @@ type inboundCall struct {
 	sigTs       SignalingTimestamps
 	jitterBuf   bool
 	projectID   string
-	audioOut    *msdk.WriteCloserSwitch[msdk.PCM16Sample] // inner writer owned by MediaPort
 }
 
 func (s *Server) newInboundCall(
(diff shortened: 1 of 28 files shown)


pkg/sip/outbound.go
```diff
@@ -385,9 +383,14 @@ func (c *outboundCall) close(ctx context.Context, end EndCall) bool {
...
 		if r := c.lkRoom; r != nil {
-			_ = r.CloseOutput()
 			_ = r.CloseWithReason(end.Status.DisconnectReason())
 		}
+
+		if c.lkRoomIn != nil {
+			if err := c.lkRoomIn.Close(); err != nil {
+				log.Warnw("error closing livekit room audio input", err)
+			}
+		}
```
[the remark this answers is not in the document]
alexfish8: Confirmed that this is actually a thing: I'm seeing the log warning a lot in staging. Will address and fix.

alexlivekit:
Huh, the race in CI is real, and is apparently a problem in media-sdk (yay zerocopy) that previous tests simply did not expose.
So... no regression compared to main.

alexfish8:
> Huh, the race in CI is real, and is apparently a problem in media-sdk (yay zerocopy) that previous tests simply did not expose. So... no regression compared to main.

Will fix this in media-sdk.

pkg/sip/inbound.go
```diff
@@ -418,23 +406,40 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
...
-				oc.updateRemoteFromSDP(sdpBodyFromRequest(req))
-				oc.cc.RecordInvite(newCSeq)
-				cc.AcceptAsKeepAlive(localSDP)
+			if oc.media == nil {
+				oc.log.Errorw("outbound call media has not been negotiated", nil)
+				return errInternal
+			}
+			localSDP, err := oc.media.GetLocalSDP()
+			if err != nil || len(localSDP) == 0 {
+				oc.log.Errorw("outbound call does not have an SDP", nil)
```
dennwc:
```suggestion
				oc.log.Errorw("outbound call does not have an SDP", err)
```

pkg/sip/inbound.go
```diff
@@ -1123,47 +1166,64 @@ func (c *inboundCall) runMediaConn(tid traceid.ID, offerData []byte, mconf *sipM
...
-	c.mon.SDPSize(len(answerData), false)
-	c.log().Debugw("SDP answer", "sdp", string(answerData))
+	return nil
+}
 
-	if err = mp.SetConfig(mc); err != nil {
-		return nil, err
+func (c *inboundCall) connectPinDTMF() {
+	if old := c.media.WriteInboundDTMFTo(&pinDTMFWriter{c.dtmf}); old != nil {
+		c.log().Warnw("media port has unexpected inbound DTMF writer", nil)
```
dennwc, line 1202: Missing `old.Close`?

pkg/sip/inbound.go
```diff
@@ -1650,21 +1727,18 @@ func (c *inboundCall) transferCall(ctx context.Context, transferTo string, heade
 		rctx, rcancel := context.WithCancel(ctx)
 		defer rcancel()
 
-		// mute the room audio to the SIP participant
-		w := c.lkRoom.SwapOutput(nil)
+		// Mute the room audio to the SIP participant.
+		// Skip closing the existing writer, which is c.audioOut.
+		_ = c.lkRoom.WriteOutboundAudioTo(nil)
```
dennwc: Why not keep `w` as it was before? It makes sure we swap back to the same pipeline that was used before the transfer.

For closing, if I read it correctly, it's already protected by `NopCloser` in the media pipeline, right? So we should close here to avoid leaks, and media pipeline will decide if it wants to expose close or not.

pkg/sip/media_pipeline.go
```diff
@@ -17,137 +17,204 @@ package sip
...
-	p.mu.Lock()
-	defer p.mu.Unlock()
-	p.conf = c
-	p.sess = sess
-
-	if err = p.setupOutput(p.tid); err != nil {
-		return err
+	err = p.setupInput(mc, audioToRoom, dtmfToRoom)
+	if err != nil {
+		return fmt.Errorf("failed to setup pipeline: %w", err)
```
dennwc:
```suggestion
		return fmt.Errorf("failed to setup pipeline input: %w", err)
```

pkg/sip/media_pipeline.go
```diff
@@ -17,137 +17,204 @@ package sip
...
+	}
+	err = p.setupOutput(mc, incomingSampleRate)
+	if err != nil {
+		return fmt.Errorf("failed to setup pipeline: %w", err)
```
dennwc:
```suggestion
		return fmt.Errorf("failed to setup pipeline output: %w", err)
```

pkg/sip/media_pipeline.go
```diff
@@ -158,92 +225,101 @@ func (p *MediaPort) dtmfHandler(h *rtp.Header, payload []byte) error {
...
 }
 
-// Must be called holding the lock
-func (p *MediaPort) setupOutput(tid traceid.ID) error {
-	if p.closed.IsBroken() {
-		return errors.New("media is already closed")
-	}
-	p.rtpLoopWG.Add(1)
-	go p.rtpLoop(tid, p.sess)
+// Construct the Audio and optionally DTM pipleine from LK PCM to SIP RTP
```
dennwc:
```suggestion
// Construct the Audio and optionally DTMF pipeline from LK PCM to SIP RTP
```

pkg/sip/media_port.go
```diff
@@ -390,106 +464,84 @@ func NewMediaPortWith(tid traceid.ID, log logger.Logger, mon *stats.CallMonitor,
...
+	localCrypto      []srtp.Profile // our SRTP material, generated once per port
+
+	mu         sync.RWMutex
+	pipeline   *mediaPortPipeline
+	localSDP   []byte
+	offer      *sdp.Offer
+	negotiated *sdp.MediaConfig
+
+	audioIn  *msdk.WriteCloserSwitch[msdk.PCM16Sample] // SIP RTP -> LK PCM
+	audioOut *msdk.WriteCloserSwitch[msdk.PCM16Sample] // LK PCM -> SIP RTP
```
dennwc, line 536: Why these two are pointers, but the other two are not?

pkg/sip/media_port.go
```diff
@@ -617,154 +660,357 @@ func (p *MediaPort) timeoutLoop(timeoutCallback func()) {
...
+		return errors.New("media is already closed")
+	}
+
+	changeSetSummary := NewChangeSetSummary(p.negotiated, c)
+
+	if changeSetSummary.includes(changeSetLocalAddr) {
+		return errors.New("unexpected local address change")
+	}
+
+	audioToPort := p.audioOut.Swap(nil) // either nil or no-op closer
```
dennwc, line 890: Maybe assert that it's the case during swap?

pkg/sip/outbound.go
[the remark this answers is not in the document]
dennwc: In general every `Close` must be safe to call twice. If it's not the case, it's worth doing so.

pkg/sip/room.go
```diff
@@ -185,17 +185,27 @@ type RoomInterface interface {
...
 	NewTrack() *mixer.Input
 	lksdk.RoomRPCInterface
+
+	// WriteOutboundAudioTo tells the room where to send audio to.
+	// Returns the previously-set writer (if one exists).
+	WriteOutboundAudioTo(w msdk.PCM16Writer) msdk.PCM16Writer
+
+	// WriteOutboundDTMFTo tells the room where to send DTMF to.
+	// Returns the previously-set writer (if one exists).
+	WriteOutboundDTMFTo(w msdk.WriteCloser[*livekit.SipDTMF]) msdk.WriteCloser[*livekit.SipDTMF]
```
dennwc, line 201: Any specific reason to use pointers in `livekit.SipDTMF`? It's a tiny payload, might as well just copy it around.

pkg/sip/outbound.go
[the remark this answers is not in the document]
alexlivekit: I understand some of the reasons native go went there (panic on double-close), but this is such a convenience matter that I'm happy to uphold the exact opposite :)

pkg/sip/media_port.go
alexlivekit: ha, nice catch.
We want to initialize the audio ones with msdk.NewWriteCloserSwitch, so that's why they're pointers.
Why the other ones are not? No reason. Fixed.

pkg/sip/media_port.go
alexlivekit: Not sure I follow. Are you talking about something like:
```
audioToPort := p.audioOut.Swap(nil)
if audioToPort != nil {
  if _, ok := audioToPort.(msdk.writeCloser[x]); !ok {
    log.Errorw(...)
  }
}
```
?

The only thin that swaps these is the configure function itself, so I'm not sure what the benefit would be there. 
The comment is there to explain why we don't close the popped value.

pkg/sip/room.go
alexlivekit: I guess the thread safety is a good benefit of copying. Fair enough.

alexlivekit: `WriteSample passes lock by value: github.com/livekit/protocol/livekit.SipDTMF contains google.golang.org/protobuf/runtime/protoimpl.MessageState contains sync.Mutex` I guess this is one,

pkg/sip/outbound.go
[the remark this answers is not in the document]
alexfish8: Yeah, it's safe to call twice twice! In any case, merged the fix in anyways.

pkg/sip/inbound.go
alexfish8: Good call, reverted to the previous flow in #808 (but renamed w).

The thing we pass to `lkRoom` is a switch writer that wraps the media port's writer, and so closing it will prevent audio from reaching the media port.

alexfish8:
> > Huh, the race in CI is real, and is apparently a problem in media-sdk (yay zerocopy) that previous tests simply did not expose. So... no regression compared to main.
> 
> Will fix this in media-sdk.

Fixed in [media-sdk#81](https://github.com/livekit/media-sdk/pull/81).

pkg/sip/inbound.go
```diff
@@ -1622,49 +1712,49 @@ func (c *inboundCall) playAudio(ctx context.Context, frames []msdk.PCM16Sample)
...
+		// Mute the room audio to the SIP participant.
+		oldRoomAudioOut := c.lkRoom.WriteOutboundAudioTo(nil)
 
 		defer func() {
 			if retErr != nil && !c.done.Load() {
-				c.lkRoom.SwapOutput(w)
-			} else if w != nil {
-				w.Close()
+				c.lkRoom.WriteOutboundAudioTo(oldRoomAudioOut)
+			} else {
```
alexfish8: We should check for nil here.

pkg/sip/inbound.go
```diff
@@ -418,23 +408,50 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
...
 		oc := s.cli.getActiveCall(cc.ID())
 		newCSeq := cc.InviteCSeq()
 		if oc != nil && oc.cc != nil && oc.cc.InviteCSeq() < newCSeq {
-			localSDP := oc.cc.LocalSDP()
-			if len(localSDP) != 0 {
-				oc.log.Infow("accepting reinvite", "content-length", req.ContentLength(), "cseq", cc.InviteCSeq())
-				oc.updateRemoteFromSDP(sdpBodyFromRequest(req))
-				oc.cc.RecordInvite(newCSeq)
-				cc.AcceptAsKeepAlive(localSDP)
+			if oc.media == nil {
```
dennwc: Should it hold any mutexes while accessing `oc.media`?

pkg/sip/inbound.go
```diff
@@ -1647,24 +1732,26 @@ func (c *inboundCall) transferCall(ctx context.Context, transferTo string, heade
...
-		w := c.lkRoom.SwapOutput(nil)
+		c.lkRoom.WriteOutboundAudioTo(nil) // Mute room
 
+		c.mmu.Lock()
+		mp := c.media
+		c.mmu.Unlock()
+		if mp == nil {
+			return transferID, fmt.Errorf("media port not found")
+		}
+		oldRoomAudioOut := mp.GetOutboundAudioWriter()
```
dennwc: Why it's called old here? It's not swapped here at least.

pkg/sip/inbound.go
```diff
@@ -1647,24 +1732,26 @@ func (c *inboundCall) transferCall(ctx context.Context, transferTo string, heade
...
 			}
 		}()
 
+		rctx, rcancel := context.WithCancel(ctx)
+		defer rcancel()
 		go func() {
-			aw := c.media.GetAudioWriter()
-
-			err := tones.Play(rctx, aw, ringVolume, tones.ETSIRinging)
+			err := tones.Play(rctx, oldRoomAudioOut, ringVolume, tones.ETSIRinging)
```
dennwc:
```suggestion
			ctx := rctx
			err := tones.Play(ctx, oldRoomAudioOut, ringVolume, tones.ETSIRinging)
```

pkg/sip/outbound.go
```diff
@@ -804,21 +828,18 @@ func (c *outboundCall) transferCall(ctx context.Context, transferTo string, head
 		rctx, rcancel := context.WithCancel(ctx)
 		defer rcancel()
 
-		// mute the room audio to the SIP participant
-		w := c.lkRoom.SwapOutput(nil)
+		// Mute the room audio to the SIP participant.
+		// Skip closing the existing writer, which is c.audioOut.
+		_ = c.lkRoom.WriteOutboundAudioTo(nil)
```
dennwc: I'd really prefer if the returned writer was wrapped into `NopCloser`, so we can avoid workarounds like these.

If you really think it's necessary, maybe check that returned value is `c.audioOut`, it will at least detect that something is off earlier, if someone changes it to something else down the line.

pkg/sip/outbound.go
```diff
@@ -804,21 +828,18 @@ func (c *outboundCall) transferCall(ctx context.Context, transferTo string, head
...
-				w.Close()
+				c.lkRoom.WriteOutboundAudioTo(c.audioOut)
 			}
 		}()
 
 		go func() {
-			aw := c.media.GetAudioWriter()
-
-			err := tones.Play(rctx, aw, ringVolume, tones.ETSIRinging)
+			err := tones.Play(rctx, c.audioOut, ringVolume, tones.ETSIRinging)
```
```suggestion
			ctx := rctx
			err := tones.Play(rctx, c.audioOut, ringVolume, tones.ETSIRinging)
```
A bit safer this way, helps avoid typos.

pkg/sip/room.go
dennwc: Oh, right, it's a protobuf. Can we swap it for a non-proto struct? I remember we had races on proto structs before.

pkg/sip/media_port.go
dennwc: Aren't they safe to use as zero values?

pkg/sip/inbound.go
alexfish8: Good call out. We're good, `oc.media` is only assigned once when the `outboundCall` is created (as is outboundCall.cc). Removing the nil checks here.

pkg/sip/inbound.go
```diff
@@ -418,23 +408,50 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
...
+			return nil
+		}
+		// TODO(alexfish): Reply with the new SDP.
 		cc.AcceptAsKeepAlive(existing.cc.OwnSDP())
 		return nil
 	}
 	if s.cli != nil { // Process reinvite for existing outbound calls
 		oc := s.cli.getActiveCall(cc.ID())
 		newCSeq := cc.InviteCSeq()
 		if oc != nil && oc.cc != nil && oc.cc.InviteCSeq() < newCSeq {
```
alexfish8: This is pre-existing, behavior, but, in a follow-up PR, we should reply with some kind of error if the new sequence number is lower than the existing one (per spec, I think this should be internal server error). Adding a todo.

pkg/sip/inbound.go
alexfish8: Done.

pkg/sip/outbound.go
alexfish8: Good call, done.

pkg/sip/inbound.go
alexfish8: Updated.

pkg/sip/outbound.go
alexfish8: Removed c.audioOut, and so now we can safely close the old audio writer, as it is now the no-op closer owned by `MediaPort`.

pkg/sip/media_port.go
alexlivekit: They are, but I wanted to fix the sample rate at allocation rather than whenever we actually negotiate media.

dennwc (approved):

dennwc:
But lets land https://github.com/livekit/sip/pull/819 first, as it's safer and easier to validate.

PR #804, opened 2026-08-21
Fix double close of inbound audio writer.

alexfish8:
This should fix the issue raised by the [Devin comment](https://github.com/livekit/sip/pull/802#discussion_r3824750040) in the main PR.

### DIFF
--- a/pkg/sip/media_port.go
+++ b/pkg/sip/media_port.go
@@ -405,8 +405,14 @@ type MediaPort interface {
 	GetOutboundDTMFWriter() msdk.WriteCloser[*livekit.SipDTMF]
 
 	// WriteInboundAudioTo tells port where to write inbound SIP audio.
+	//
+	// MediaPort.Close() will propagate to the argument writer. The caller is
+	// responsible for closing the returned media writer.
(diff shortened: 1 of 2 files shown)


pkg/sip/media_port.go
```diff
@@ -405,9 +405,20 @@ type MediaPort interface {
...
 	GetOutboundDTMFWriter() msdk.WriteCloser[*livekit.SipDTMF]
 
-	// WriteInboundAudioTo tells the MediaSegment where to write inbound SIP audio.
+	// WriteInboundAudioTo tells the MediaSegment where to write inbound SIP
+	// audio.
+	//
+	// Takes ownership of the argument writer and relinquishes ownership of the
+	// returned writer. The caller must not close the argument writer (unless
+	// later relinquishes), and the caller is responsible for closing the
+	// returned writer.
```
alexlivekit: nit: Instead of ownership comment, jusat mention something like `propagates Close() when media port closes.`

alexlivekit (approved):

pkg/sip/media_port.go
alexfish8: Done.

PR #807, opened 2026-08-21
TEL-912 Response on transfer requests

genseric-ghiro:
## Summary
- Most notable change is a new TransferResponse with status and reason details
- Adding a 1s wait in `waitReferResult` such that the final `NOTIFY` has a chance to get in before the `BYE`
- Failing fast when the `NOTIFY` ends the subscription (same behavior as before, but no longer waiting for the timeout)


pkg/sip/protocol.go
```diff
@@ -360,60 +375,93 @@ func parseNotifyBody(body string) (int, string, error) {
...
+		// whose notifier gives up, ends the same way.
+		if !info.Sub.Terminated() {
+			// still trying
+			return
+		}
+		reason := info.Sub.Reason
+		if reason == "" {
+			reason = "unspecified"
+		}
+		result = psrpc.NewErrorf(psrpc.UpstreamServerError, "call transfer failed: %w (reason %q, last status %d)",
```
genseric-ghiro, line 458: Still the same behavior as before (returning an error), but returning it faster, before waiting for the timeout, and with more details

pkg/sip/protocol.go
```diff
@@ -425,6 +473,34 @@ func handleReferNotify(cseq uint32, status int, reason string, referCseq uint32,
...
+		return err
+	case <-callDone:
+		select {
+		case err := <-referDone:
+			log.Infow("refer result raced call end", "error", err)
+			return err
+		case <-time.After(referResultGrace):
+		}
+		log.Infow("refer failed: call ended before transfer completed")
+		return psrpc.NewError(psrpc.Aborted, errTransferCallEnded)
```
genseric-ghiro, line 507: Returning an error here now, but catching it in service.go and nulling it out (otherwise it's not possible to set the status and reason for the response, which happens at the service level)

dennwc (approved):

PR #816, opened 2026-08-27
Add recovery handlers and metrics for SDP parsing.

alexfish8:
More defensiveness around invalid SDP.

### DIFF
--- a/pkg/sip/media_port.go
+++ b/pkg/sip/media_port.go
@@ -785,7 +785,7 @@ func (p *mediaPort) GenerateAnswer(offerData []byte) ([]byte, error) {
 		return p.GetLocalSDP()
 	}
 
-	offer, err := parseOfferWith(p.mon, p.codecs, offerData)
+	offer, err := parseOfferWith(p.log, p.mon, p.codecs, offerData)
 	if err != nil {
 		return nil, SDPError{Err: err}
 	}
@@ -822,7 +822,7 @@ func (p *mediaPort) ProcessAnswer(answerData []byte) error {
 		return errors.New("no offer generated")
 	}
 
-	answer, err := parseAnswerWith(p.mon, p.codecs, answerData)
+	answer, err := parseAnswerWith(p.log, p.mon, p.codecs, answerData)
 	if err != nil {
 		return SDPError{Err: err}
 	}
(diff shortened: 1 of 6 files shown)


genseric-ghiro:
How frequently does that happen today, i.e. panics during SDP parsing? And when it does happen, it brings down the entire pod?

pkg/sip/media_port.go
```diff
@@ -709,6 +711,35 @@ func (p *MediaPort) GetAudioWriter() msdk.PCM16Writer {
...
 }
 
+func parseSDPWithRecovery[T any](mon *stats.CallMonitor, fn func(*msdk.CodecSet, []byte) (*T, error), codecs *msdk.CodecSet, data []byte) (res *T, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			panicErr, ok := r.(error)
+			if !ok {
+				panicErr = fmt.Errorf("%v", r)
+			}
+			logger.GetLogger().Errorw("panic while parsing SDP", panicErr, "stacktrace", string(debug.Stack()))
```
dennwc: We do have a log at the call site, so maybe pass it as well?

pkg/stats/monitor.go
```diff
@@ -252,6 +253,14 @@ func (m *Monitor) Start(conf *config.Config) error {
 		ConstLabels: prometheus.Labels{"node_id": conf.NodeID},
 	}, []string{"dir", "provider"}))
 
+	m.sdpParsePanics = mustRegister(m, prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace:   "livekit",
+		Subsystem:   "sip",
+		Name:        "sdp_parse_panics_total",
```
dennwc: Maybe add `sdp_parse_errors_total` and add `reason` as a label?

dennwc (approved):

pkg/sip/media_port.go
alexfish8: Ah, good idea, done.

pkg/stats/monitor.go
alexfish8: Done.

alexfish8:
> How frequently does that happen today, i.e. panics during SDP parsing? And when it does happen, it brings down the entire pod?

I think this last happened about a month ago? Previously, this would take down the pod, but, I added some recovery handlers at the SIP-handler level, so now, a pod wouldn't crash in that case. In that sense, the recovery handlers in this PR are sort of redundant, but, I think they're still worthwhile, as they ensure that these SDP-parse panics will be properly classified as SDP errors and go through the normal SDP-error handling flow rather than bubbling up to the server recovery handler and being classified as generic errors.

alexfish8 (commented):
thanks for the review!

PR #818, opened 2026-08-28
TELCU-25 Fallback to next address if current one fails

genseric-ghiro:
## Summary
- Add a feature flag at the project level to allow choosing between starting with DNS A records or SRV records when doing hostname resolution
- TODO: Bump sipgo once this is merged https://github.com/livekit/sipgo/pull/11

### DIFF
--- a/pkg/sip/client.go
+++ b/pkg/sip/client.go
@@ -38,6 +38,7 @@ import (
 	"github.com/livekit/psrpc"
 	"github.com/livekit/sipgo"
 	"github.com/livekit/sipgo/sip"
+	"github.com/livekit/sipgo/transport"
 
 	"github.com/livekit/sip/pkg/config"
 	siperrors "github.com/livekit/sip/pkg/errors"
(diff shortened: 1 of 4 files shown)


pkg/sip/client.go
```diff
@@ -50,6 +51,9 @@ import (
 type SIPClient interface {
 	TransactionRequest(req *sip.Request, options ...sipgo.ClientRequestOption) (sip.ClientTransaction, error)
 	WriteRequest(req *sip.Request, options ...sipgo.ClientRequestOption) error
+	// TransportLayer exposes sipgo's resolution and connection handling. It may
+	// be nil for clients that do not own one.
+	TransportLayer() *transport.Layer
```
dennwc: I'm not sure yet if it's a good idea, but what if we expose a single method which is similar to the `net.Resolver`? Basically with a fake SIP client, regular resolver would do. And without one, sipgo will use it's logic for SRV, etc.

Or as an alternative, maybe the resolver in sipgo can be split, so that it doesn't rely on transport layer there. Just let it be a new interface and pass to sipgo. Or let sipgo create this special implementation internally in transport layer, and sip will create a second one here, if needed.

genseric-ghiro: Changed it to `ResolveTargets()` specifically

PR #820, opened 2026-08-29
Add late SDP answer support for inbound calls.

alexfish8:
- Whenever we don't receive an offer with the initial invite, we respond with a 200 and an SDP offer, and we expect an ACK to eventually arrive with an answer
  - In the event that we are waiting for a late answer, do not proceed with the call until we receive an ACK (previously, this logic was solely gated on some experimental flag being set)
  - Note: this is is gated by a feature-flag currently
- Small refactor of `acceptCall` function variable (defined in `handleInvite`)
  - Move logic out of function variable into three top-level functions: `acceptCallAndWaitForMedia`, `acceptCall`, and `waitForMedia`
- Reinvite handling for late offers
    - Reply with the negotiated SDP (rather than the offer we sent in response to the offerless INVITE)
- Tests

### DIFF
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -409,6 +409,13 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 	existing := s.byLocalTag[cc.ID()]
 	s.cmu.RUnlock()
 	if existing != nil && existing.cc.InviteCSeq() < cc.InviteCSeq() {
+		if existing.LateAnswerPending() {
+			// Our offer from the 200 OK has not been answered yet. Negotiating a new offer now would
+			// discard the pending one and break the late answer when the ACK arrives.
+			existing.log().Infow("rejecting reinvite, late answer pending", "cseq", cc.InviteCSeq())
+			cc.RejectAsKeepAlive(statusRequestPending, "Request Pending")
+			return nil
+		}
 		existing.log().Infow("reinvite", "content-length", req.ContentLength(), "cseq", cc.InviteCSeq())
 		if err := existing.updateRemoteFromSDP(sdpBodyFromRequest(req)); err != nil {
 			log.Errorw("failed to update inbound call SDP", err)
@@ -734,10 +741,13 @@ type inboundCall struct {
 	joinDur     func() time.Duration
 	done        atomic.Bool
 	started     core.Fuse
-	stats       Stats
-	sigTs       SignalingTimestamps
-	jitterBuf   bool
-	projectID   string
+	// lateAnswerPending is set while we have sent an SDP offer in the 200 OK and have not yet
+	// processed the answer from the ACK.
+	lateAnswerPending atomic.Bool
+	stats             Stats
+	sigTs             SignalingTimestamps
+	jitterBuf         bool
+	projectID         string
 }
 
 func (s *Server) newInboundCall(
@@ -961,6 +971,7 @@ func (c *inboundCall) handleInvite(ctx context.Context, tid traceid.ID, req *sip
 	var sdpBody []byte
 	expectingLateAnswer := len(rawSDP) == 0
 	if expectingLateAnswer {
+		c.lateAnswerPending.Store(true)
 		sdpBody, err = c.media.GenerateOffer()
 		if err != nil {
 			return rejectMedia(err)
@@ -1309,6 +1320,8 @@ func (c *inboundCall) negotiateMedia(sdpData []byte) ([]byte, error) {
 func (c *inboundCall) negotiateMediaForLateAnswer(answerData []byte) error {
 	c.mmu.Lock()
 	defer c.mmu.Unlock()
+	defer c.lateAnswerPending.Store(false)
+
 	if c.media == nil {
 		return errors.New("media port not created")
 	}
@@ -1326,6 +1339,13 @@ func (c *inboundCall) negotiateMediaForLateAnswer(answerData []byte) error {
 	}
 
 	return c.updateCallStateAudioLocked()
+}
+
+// LateAnswerPending reports whether the call sent an SDP offer in its 200 OK and is still
+// waiting for the answer in the ACK. While this is true, the offer/answer exchange is open
+// and a re-INVITE cannot be negotiated (RFC 3264 §5, RFC 3261 §14.2).
+func (c *inboundCall) LateAnswerPending() bool {
+	return c.lateAnswerPending.Load()
 }
 
 func (c *inboundCall) waitMedia(ctx context.Context) (bool, error) {
@@ -2257,6 +2277,12 @@ retries:
 }
 
 func (c *sipInbound) AcceptAck(req *sip.Request, tx sip.ServerTransaction) {
+	cseq := req.CSeq()
+	if cseq == nil || cseq.SeqNo != c.inviteCSeq {
+		c.log.Debugw("ignoring ACK for another INVITE", "inviteCSeq", c.inviteCSeq, "ackCSeq", cseq)
+		return
+	}
+	// Only store the first ACK seen.
 	c.ack.CompareAndSwap(nil, req)
 	c.acked.Break()
 }
--- a/pkg/sip/inbound.go
+++ b/pkg/sip/inbound.go
@@ -1074,7 +1074,7 @@ func (c *inboundCall) handleInvite(ctx context.Context, tid traceid.ID, req *sip
 	return c.waitForCallEnd(ctx, ackReceived, ackTimeout, mconf.MediaTimeout)
 }
 
-func (c *inboundCall) acceptCall(ctx context.Context, disp CallDispatch, sdpData []byte, expectingLateOffer bool) error {
+func (c *inboundCall) acceptCall(ctx context.Context, disp CallDispatch, sdpData []byte, waitForAck bool) error {
 	headers := disp.Headers
 	c.attrsToHdr = disp.AttributesToHeaders
 	if r := c.lkRoom.Room(); r != nil {
@@ -1082,7 +1082,7 @@ func (c *inboundCall) acceptCall(ctx context.Context, disp CallDispatch, sdpData
 	}
 	c.log().Infow("Accepting the call", "headers", headers)
 	taccept := c.mon.StageDurTimer("sip-accept")
-	err := c.cc.Accept(ctx, sdpData, headers, expectingLateOffer)
+	err := c.cc.Accept(ctx, sdpData, headers, waitForAck)
 	taccept()
 	c.sigTs.AcceptTime = time.Now()
 	if errors.Is(err, errNoACK) {
@@ -1307,13 +1307,10 @@ func (c *inboundCall) negotiateMedia(sdpData []byte) ([]byte, error) {
 	}
 	defer c.mon.StageDurTimer("start-media")()
 
-	var answerData []byte
-	var err error
-
 	c.mon.SDPSize(len(sdpData), true)
 	c.log().Debugw("SDP offer", "sdp", string(sdpData))
 
-	answerData, err = c.media.GenerateAnswer(sdpData)
+	answerData, err := c.media.GenerateAnswer(sdpData)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


pkg/sip/inbound.go
```diff
@@ -1219,29 +1287,48 @@ func (c *inboundCall) negotiateMedia(offerData []byte) ([]byte, error) {
...
 
+func (c *inboundCall) negotiateMediaForLateAnswer(answerData []byte) error {
+	c.mmu.Lock()
+	defer c.mmu.Unlock()
+	if c.media == nil {
+		return errors.New("media port not created")
+	}
+	if c.media.NegotiatedAudio() != nil {
+		return nil
+	}
```
alexlivekit, line 1337: Highly unexpected, no? 
Can this happen on a re-invite?

pkg/sip/inbound.go
```diff
@@ -1219,29 +1287,48 @@ func (c *inboundCall) negotiateMedia(offerData []byte) ([]byte, error) {
...
+	defer c.mon.StageDurTimer("process-late-answer")()
+
+	// TODO(alexfish): Should we create a new stats histogram for late
+	// answers?
+	c.mon.SDPSize(len(answerData) /*isOffer=*/, false)
```
alexlivekit: Odd way to phrase this. I'd sooner do:
```go
isOffer := false
c.mon.SDPSize(len(answerData), isOffer)
```
Ditto in other places. At the very least move the comma `c.mon.SDPSize(len(answerData), /*isOffer=*/ false)`.

pkg/sip/inbound.go
```diff
@@ -1219,29 +1287,48 @@ func (c *inboundCall) negotiateMedia(offerData []byte) ([]byte, error) {
...
 
-	answerData, err := c.media.GenerateAnswer(offerData)
-	if err != nil {
+	var answerData []byte
+	var err error
+
+	c.mon.SDPSize(len(sdpData) /*isOffer=*/, true)
+	c.log().Debugw("SDP offer", "sdp", string(sdpData))
+
+	if answerData, err = c.media.GenerateAnswer(sdpData); err != nil {
```
alexlivekit: NIT: imo the original code is more "go" in style:
```go
answerData, err := c.media.GenerateAnswer(offerData)
if err != nil { ..}
```

pkg/sip/inbound.go
```diff
@@ -952,67 +954,32 @@ func (c *inboundCall) handleInvite(ctx context.Context, tid traceid.ID, req *sip
...
 		return rejectMedia(err)
 	}
 
-	ok := false
-	var answerData []byte
-	if pinPrompt {
-		// Negotiate before Accept so pin prompts and DTMF have a live pipeline.
-		// Encryption is picked here, before the room is selected.
-		answerData, err = c.negotiateMedia(rawSDP)
+	var sdpResponseBody []byte
```
alexlivekit: NIT: call it `sdpBody`. `sdpResponse` is pretty much implying `answer` which isn't what you're going for here.

pkg/sip/inbound.go
```diff
@@ -1037,10 +1004,15 @@ func (c *inboundCall) handleInvite(ctx context.Context, tid traceid.ID, req *sip
 	if pinPrompt {
 		status = CallActive
 	}
-	answerData, err = c.negotiateMedia(rawSDP)
-	if err != nil {
-		return rejectMedia(err)
+
+	// We can't negotiate media until we've received an answer.
+	if !expectingLateAnswer && !pinPrompt {
```
alexlivekit: NIT: `} else if if !expectingLateAnswer {`

alexlivekit (commented):
Nice. Mostly a bunch of nits, but one concern to map out.

pkg/sip/inbound.go
alexfish8: Reverted to the previous style.

pkg/sip/inbound.go
alexfish8: Done.

pkg/sip/inbound.go
alexfish8: Done.

pkg/sip/inbound.go
alexfish8: Removed (for some reason, my IDE auto-formatter insists upon putting the comment before the comma, and I'd rather not fight it). I might send out a follow-up PR that adds different helper methods like `SDPOfferSize` and `SDPAnswerSize` to get the marginal readability improvement.

pkg/sip/inbound.go
alexfish8: Yeah, this seems pretty unlikely and would involve a re-invite. We'd need to see a sequence like:

1. INVITE comes in without an offer. `handleInvite` sends an offer and waits for an ACK
2. A second INVITE comes with both an offer and a higher cseq than the first invite, and so `updateRemoteFromSDP` is invoked, which sets up media.
3. ACK comes in for the first invite, `negotiateMediaForLateAnswer` is eventually invoked, and this branch is hit.

Per spec, I think it's illegal for a client to send a second, new invite while the first invite's transaction hasn't closed, but, this is still probably something we should guard against. This would also affect initial INVITEs with offers, too. Given that this is a pre-existing thing, maybe we add guards against this in a follow-up PR?

alexfish8 (commented):
thanks for the review!

alexfish8:
Still need to write tests and handle the ACK issue.

alexfish8:
PR updates:

1. Add tests
2. Gate late offer support on a feature flag (there's additional codec changes that will be deployed to cloud-sip, and so as to minimize the risk of breaking anything by deploying yet more changes, we'll feature gate this for now)
3. Make code more robust to reinvites (ignore reinvites requests if we're waiting for a late answer in an ACK for a previous invite)
4. Ensure that, if a reinvite request comes in, we reply with the negotiated SDP (rather than the offer we sent in response to the offerless invite).

pkg/sip/inbound.go
```diff
@@ -2170,9 +2258,15 @@ retries:
 }
 
 func (c *sipInbound) AcceptAck(req *sip.Request, tx sip.ServerTransaction) {
+	c.ack.CompareAndSwap(nil, req)
```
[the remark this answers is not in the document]
alexfish8: Not supporting reinvites yet.

pkg/sip/inbound.go
```diff
@@ -1058,9 +1049,93 @@ func (c *inboundCall) handleInvite(ctx context.Context, tid traceid.ID, req *sip
...
+	if !waitForAck {
+		// Start this timer right after the Accept.
+		ackTimeout = time.After(inviteOkAckLateTimeout)
+	}
+
+	if expectingLateAnswer {
+		ack := c.cc.Ack()
+		if ack == nil {
+			c.log().Errorw("ack not found", nil)
+			return false, nil, fmt.Errorf("ack not found")
```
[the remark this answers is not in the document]
alexfish8: If we were expecting a late answer, we would have already waited for the ack.

pkg/sip/inbound.go
```diff
@@ -732,10 +741,13 @@ type inboundCall struct {
 	joinDur     func() time.Duration
 	done        atomic.Bool
 	started     core.Fuse
-	stats       Stats
-	sigTs       SignalingTimestamps
-	jitterBuf   bool
-	projectID   string
+	// lateAnswerPending is set while we have sent an SDP offer in the 200 OK and have not yet
+	// processed the answer from the ACK.
+	lateAnswerPending atomic.Bool
```
alexlivekit: NIT
```suggestion
	lateAnswerPending atomic.Bool // An invite was revieved without an offer, pending late offer/answer
```

pkg/sip/inbound.go
```diff
@@ -952,67 +964,43 @@ func (c *inboundCall) handleInvite(ctx context.Context, tid traceid.ID, req *sip
...
+	if err := c.createMediaPort(mconf, conf, disp.FeatureFlags); err != nil {
+		return rejectMedia(err)
+	}
+
+	var expectingLateAnswer bool
+	if len(rawSDP) == 0 {
+		expectingLateAnswer = featureFlagEnabled(disp.FeatureFlags, lateOfferFeatureFlag)
+		if !expectingLateAnswer {
+			log.Infow("Offerless INVITE, but late offer is not enabled for this project")
+		} else {
```
alexlivekit: This needs to abort. If we have an offerless INVITE and the flag is not enabled, we need to kill the call attempt.

pkg/sip/inbound.go
```diff
@@ -952,67 +964,43 @@ func (c *inboundCall) handleInvite(ctx context.Context, tid traceid.ID, req *sip
...
-	if err := c.createMediaPort(mconf, conf, disp.FeatureFlags); err != nil {
-		return rejectMedia(err)
+	var sdpBody []byte
+	if expectingLateAnswer {
+		c.lateAnswerPending.Store(true)
+		sdpBody, err = c.media.GenerateOffer()
+		if err != nil {
+			return rejectMedia(err)
+		}
+		c.mon.SDPSize(len(sdpBody), true)
```
alexlivekit: I'm not sure a single boolean captures the spirit of it here. Alongside `isOffer`, I'd also add `isLateOffer`, so we would know which offers/answers we're getting from a remote, and which are generated by us.

pkg/sip/inbound.go
```diff
@@ -407,6 +409,13 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 	existing := s.byLocalTag[cc.ID()]
 	s.cmu.RUnlock()
 	if existing != nil && existing.cc.InviteCSeq() < cc.InviteCSeq() {
+		if existing.LateAnswerPending() {
+			// Our offer from the 200 OK has not been answered yet. Negotiating a new offer now would
+			// discard the pending one and break the late answer when the ACK arrives.
```
alexlivekit: super-nit: This comment is now repeated on the variable, function, and in here. Should we consolidate this?

pkg/sip/inbound.go
```diff
@@ -1219,29 +1306,60 @@ func (c *inboundCall) negotiateMedia(offerData []byte) ([]byte, error) {
...
+		return err
+	}
+	c.cc.SetOwnSDP(localSDP)
+
+	return c.updateCallStateAudioLocked()
+}
+
+// LateAnswerPending reports whether the call sent an SDP offer in its 200 OK and is still
+// waiting for the answer in the ACK. While this is true, the offer/answer exchange is open
+// and a re-INVITE cannot be negotiated (RFC 3264 §5, RFC 3261 §14.2).
```
alexlivekit: Slightly misleading: This is true from the moment we have generated the offer. It does not mean we have sent it out (may be delayed by room join, for example).

pkg/sip/inbound.go
```diff
@@ -2101,7 +2220,13 @@ func (c *sipInbound) OwnSDP() []byte {
 	return c.lastSDP
 }
 
-func (c *sipInbound) Accept(ctx context.Context, sdpData []byte, headers map[string]string) error {
+func (c *sipInbound) SetOwnSDP(sdpData []byte) {
```
alexlivekit, line 2199: I'm not I'm a big fan of this. iiuc, `c.lastSDP` is supposed to represent the last local sdp.
Two things about this:
1. `c.lastSDP` is curently set in `sipInbound.accept()`. For the normal path, that works. For late offer, the value is going to be the generated offer, which will then at some point be overwritten by calling  `SetOwnSDP()` at the end of the SDP exchange.
2. We already have the local SDP in the form of ic/oc`.media.GetLocalSDP()`. Using this fact is going to be somewhat of a rewrite, so we can defer this bit.

alexlivekit (approved):
Ended up writing a follow-up PR in https://github.com/livekit/sip/pull/835. No real outstanding issues remain. LGTM

pkg/sip/inbound.go
alexfish8: Good call, thanks for fixing.

pkg/sip/inbound.go
alexfish8: That makes sense. Added a TODO for now.

PR #823, opened 2026-08-31
TEL-886: Change internal DTMF type away from protobuf

alexlivekit:


pkg/sip/room.go
```diff
@@ -543,9 +544,10 @@ func (r *Room) newRoomCallback(conf *config.Config, rconf RoomConfig) *lksdk.Roo
 				switch data := data.(type) {
 				case *livekit.SipDTMF:
 					r.stats.dataPackets.Add(1)
-					// TODO: Only generate audio DTMF if the message was a broadcast from another SIP participant.
-					//       DTMF audio tone will be automatically mixed in any case.
-					r.sendDTMF(context.Background(), data)
+					err := r.outboundDTMF.WriteSample(data.Digit) // Note: This ignores code
```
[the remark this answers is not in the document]
alexlivekit: This is consistent with [current main](https://github.com/livekit/sip/blob/55e76daca321960fb29dad153ed8f8ec41a403b4/pkg/sip/room.go#L724) .

pkg/sip/inbound.go
```diff
@@ -1148,15 +1148,21 @@ func (w *pinDTMFWriter) Close() error {
 	return nil
 }
 
-func (w *pinDTMFWriter) WriteSample(msg *livekit.SipDTMF) error {
-	if msg == nil {
-		return nil
+func (w *pinDTMFWriter) WriteSample(sample string) error {
+	if len(sample) != 1 {
```
dennwc, line 1152: Maybe just have `(sample byte)` then?

dennwc (approved):

pkg/sip/inbound.go
alexlivekit: For now we're using the inconsistency. Main supports receiving multiple digits from LK room, but the other direction it's only 1 digit at a time. String allows both, without needing to manually manage encoding pauses.

PR #828, opened 2026-09-02
Give each incoming SSRC its own jitter buffer so a transferred stream is not muted

ahsang:
## Summary

Inbound audio on a SIP call went permanently silent after a PBX transfer (auto-attendant -> extension) while signalling, outbound audio and the media timeout all stayed healthy. Reproduced 5/5 on transfer calls.

- After the transfer the far end keeps the same 5-tuple but relays a different source, so the service sees a **new SSRC with an unrelated sequence-number space**. On the captured call the old stream was near seq 21294 and the new one started at 12676.
- All SSRCs on a port shared **one** `jitter.Buffer` (`rtp.HandleJitter` wrapped once in `setupInput`). That buffer drops anything up to 32768 behind its last popped sequence number and never advances on a drop, so every packet of the new stream was discarded: 8618 packets, ~172 s at 50 pps, longer than the rest of the call.
- The media timeout never fires because packets are counted before the handler chain, and the mid-dialog re-INVITE is answered with the cached SDP without touching the pipeline, so nothing reset the buffer.

## Changes

- `mediaPortPipeline` keeps the inbound chain jitter-free (`inputChain`), and each RTP read stream (one per SSRC) wraps it in its **own** jitter buffer via `newStreamHandler()`. A no-op-close wrapper stops a finishing stream from closing the shared mux, and each read loop closes only its own buffer on exit, before `Close()` tears down the chain.
- Per-buffer loss/drop counters are folded into `PortStats` as **deltas**. `jitter_buffer_packets_lost` / `_dropped` on the `call statistics` line are now the monotonic sum across all SSRCs and pipeline rebuilds of a call; previously they were one buffer's cumulative count and could regress on renegotiation.
- With `enable_jitter_buffer: false` the chain is unchanged.

## Tests

- `TestMediaPipelineJitterNewSSRCBackwardSeq`: two SSRCs through a real UDP socket with the captured sequence numbers. Fails on main with 15 dropped packets and no room audio; passes here.
- `TestMediaPipelineJitterStatsAccumulateAcrossSSRC`: loss stats sum across buffers.
- `TestMediaPipelineTeardownMultiSSRC` and `TestMediaPipelineConcurrentSSRCPump` now run with jitter on and off under `-race`.
- Full `go test -race ./pkg/...` and `./test/integration/...` pass locally.

## Related

- Companion media-sdk change so a far end that keeps its SSRC but restarts sequence numbers also resyncs: https://github.com/livekit/media-sdk/pull/86. A guard test for that case is ready to land with the dependency bump once it merges.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MND95Kk3MD7828eZBw8ipa

### DIFF
--- a/pkg/sip/media_pipeline.go
+++ b/pkg/sip/media_pipeline.go
@@ -71,19 +71,21 @@ type mediaPortPipeline struct {
 	conf *MediaPortPipelineConfig // Expected to be owned by caller, not managed
 
 	// Owned by pipeline
-	ctx           context.Context
-	cancel        context.CancelFunc
-	sess          rtp.Session
-	rtpLoopWG     sync.WaitGroup
-	muxToRoom     atomic.Pointer[rtp.HandlerCloser]
-	dtmfMixer     *mixer.Mixer
-	audioToRoom   rtp.HandlerCloser
-	dtmfToRoom    rtp.HandlerCloser
-	dtmfHandler   msdk.WriteCloser[string] // Reference, not closed
-	audioToPort   msdk.PCM16Writer         // post-mixer chain towards port
-	mixerToPort   msdk.PCM16Writer         // Reference, not closed
-	dtmfToPort    msdk.WriteCloser[string]
-	lastDTMFEvent atomic.Uint64 // composite (timestamp, event code) of last DTMF packet seen
+	ctx                 context.Context
+	cancel              context.CancelFunc
+	sess                rtp.Session
+	rtpLoopWG           sync.WaitGroup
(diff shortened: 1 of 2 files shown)


ahsang:
Closing: livekit/media-sdk#93 resyncs the jitter buffer on an SSRC change, which covers the transferred-stream case this PR addressed. The same-SSRC restart case is handled by media-sdk#95, plus livekit/media-sdk#86 for jumps beyond the reorder window. Thanks!

PR #842, opened 2026-09-10
Add Opus and DTLS-SRTP/ICE-lite support for Meta WhatsApp calling

mab3321:
## Summary

Adds opt-in Opus and inbound WebRTC-style DTLS-SRTP support, including ICE-lite interoperability required by Meta WhatsApp Business Calling.

## Changes

- add an opt-in Opus codec with configurable bitrate, complexity, FEC, and expected packet loss
- parse DTLS fingerprints, setup roles, RTCP mux, and ICE credentials/candidates from inbound SDP
- generate DTLS-SRTP/ICE answer attributes while keeping legacy RTP and SDES-SRTP paths unchanged
- establish the controlling ICE connection for ICE-lite peers before DTLS
- derive SRTP/SRTCP keys from the negotiated DTLS profile
- close ICE, mux, SRTP, SRTCP, and DTLS transports without blocking call teardown
- add unit coverage for Opus negotiation and Meta-style DTLS/ICE SDP
- document the new opt-in configuration

## Configuration

```yaml
enable_opus: true
dtls_srtp:
  enabled: true
  handshake_timeout: 10s
```

Both features default to disabled.

## Validation

- Docker image builds against the current upstream `main`
- `go test ./pkg/sip -count=1` passes
- verified real inbound Meta WhatsApp calls with bidirectional Opus audio over ICE/DTLS-SRTP
- verified remote hangup completes promptly after transport teardown
- verified an ordinary Twilio PSTN call continues through the existing RTP path with zero packet loss and sub-millisecond SIP media processing


mab3321:
Implemented the four requested review fixes in `fc373f0`:

- **DTLS/ICE re-INVITE:** renegotiation now includes fingerprint, setup role, ICE credentials, and all candidates. Unchanged session refreshes reuse the active transport SDP parameters; changed transport state rebuilds the transport.
- **Cancelable teardown:** the session-scoped context is created synchronously and canceled first by `Close()`. ICE gathering/checks, DTLS handshake, publication, readiness, and startup waits derive from it. Resource publication and cleanup are race-safe, including Close racing ICE publication, DTLS completion, SRTP creation, or readiness signaling.
- **Multiple ICE candidates:** all supported component-1 UDP/IPv4 candidates are retained with metadata, deduplicated, and added before connectivity checks.
- **Opus settings:** the pinned media-sdk has no encoder configuration API, so unsupported non-default bitrate, complexity, FEC, and packet-loss settings now fail explicitly instead of being silently ignored.

Compatibility is covered by tests proving ordinary RTP/AVP and SDES-SRTP do not enter DTLS/ICE state. No transcoding was introduced.

Server-side validation completed: focused DTLS/ICE/Opus regressions, existing re-INVITE and late-offer tests, race-sensitive DTLS teardown tests, legacy RTP/SDES isolation under the race detector, and the full `./pkg/sip` package. The final full package run passed.

Operational diagnosis: the affected call had bidirectional DTLS-SRTP media until inbound RTP stopped, followed by a media timeout and a five-minute teardown delay. The re-INVITE transport reuse and cancel-safe teardown changes directly cover those failure modes.

PR #844, opened 2026-09-11
Update code to be tolerant of unknown authentication failure reasons.

alexfish8:
Treat unrecognized `rpc.SIPTrunkAuthenticationError` values as an authentication failure instead of accepting the call. Also, map certain authentication error codes to `ClientError` rather than `ServerError`.

Today the adapter in pkg/service/psrpc.go handles only `QUOTA_EXCEEDED` and `NO_TRUNK_FOUND`. Any other non-zero code falls through the drop and password checks and is returned as `AuthAccept` with an empty trunk ID. This PR adds an AuthFailureUnknown result for that case. The INVITE is answered with 401 and recorded as `ClientError("auth-unknown")`.

Also, update the switch statements to handle a new kind of authentication failure: `ROUTE_NOT_ALLOWED`.

Motivation: The auth service will start returning new enum values for rejections it currently reports as errors. Those errors make SIP end the call as `ServerError("auth-error")` with a 503, which should be reserved for availability failures. This change has to roll out before the auth service does, so a newer auth service talking to an older SIP rejects rather than admits.

NOTE: This depends on [livekit/protocol#1798](https://github.com/livekit/protocol/pull/1798).

### DIFF
--- a/pkg/service/psrpc.go
+++ b/pkg/service/psrpc.go
@@ -28,19 +28,19 @@ func GetAuthCredentials(ctx context.Context, psrpcClient rpc.IOInfoSIPClient, ca
 		return sip.AuthInfo{}, err
 	}
 
-	// Handle specific authentication error codes
-	switch resp.ErrorCode {
-	case rpc.SIPTrunkAuthenticationError_SIP_TRUNK_AUTH_ERROR_QUOTA_EXCEEDED:
+	if resp.ErrorCode != rpc.SIPTrunkAuthenticationError_SIP_TRUNK_AUTH_ERROR_NONE {
+		var authResult sip.AuthResult
+		switch resp.ErrorCode {
+		case rpc.SIPTrunkAuthenticationError_SIP_TRUNK_AUTH_ERROR_QUOTA_EXCEEDED:
+			authResult = sip.AuthQuotaExceeded
+		case rpc.SIPTrunkAuthenticationError_SIP_TRUNK_AUTH_ERROR_NO_TRUNK_FOUND:
+			authResult = sip.AuthNoTrunkFound
+		default:
+			authResult = sip.AuthFailureUnknown
+		}
 		return sip.AuthInfo{
 			ProjectID:     resp.ProjectId,
-			Result:        sip.AuthQuotaExceeded,
-			ProviderInfo:  resp.ProviderInfo,
-			Observability: resp.Observability,
-		}, nil
-	case rpc.SIPTrunkAuthenticationError_SIP_TRUNK_AUTH_ERROR_NO_TRUNK_FOUND:
-		return sip.AuthInfo{
-			ProjectID:     resp.ProjectId,
-			Result:        sip.AuthNoTrunkFound,
+			Result:        authResult,
 			ProviderInfo:  resp.ProviderInfo,
 			Observability: resp.Observability,
 		}, nil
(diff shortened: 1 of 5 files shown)


pkg/service/psrpc.go
```diff
@@ -25,22 +26,27 @@ func GetAuthCredentials(ctx context.Context, psrpcClient rpc.IOInfoSIPClient, ca
...
-			ProviderInfo:  resp.ProviderInfo,
-			Observability: resp.Observability,
-		}, nil
-	case rpc.SIPTrunkAuthenticationError_SIP_TRUNK_AUTH_ERROR_NO_TRUNK_FOUND:
+	if resp.ErrorCode != rpc.SIPTrunkAuthenticationError_SIP_TRUNK_AUTH_ERROR_NONE {
+		var authResult sip.AuthResult
+		switch resp.ErrorCode {
+		case rpc.SIPTrunkAuthenticationError_SIP_TRUNK_AUTH_ERROR_QUOTA_EXCEEDED:
+			authResult = sip.AuthQuotaExceeded
+		case rpc.SIPTrunkAuthenticationError_SIP_TRUNK_AUTH_ERROR_NO_TRUNK_FOUND:
```
hechen-eng, line 42: what's the difference between `resp.ErrorCode` with `rpc.SIPTrunkAuthenticationError_SIP_TRUNK_AUTH_ERROR_NO_TRUNK_FOUND` and `psrpc.GetErrorCode(err)` with `psrpc.NotFound`? both are not found 🤔

pkg/sip/inbound.go
```diff
@@ -556,6 +556,16 @@ func (s *Server) processInvite(req *sip.Request, tx sip.ServerTransaction) (retE
 		log.Warnw("Rejecting inbound, no trunk found", nil)
 		cc.RespondAndDrop(sip.StatusNotFound, "No trunk found")
 		return psrpc.NewErrorf(psrpc.NotFound, "no trunk found for call")
+	case AuthFailureOther:
+		cmon.InviteErrorShort(stats.ClientError("auth-other"))
+		cc.RespondAndDrop(sip.StatusForbidden, "Auth failure")
+		return psrpc.NewErrorf(psrpc.PermissionDenied, "auth failure")
+	case AuthRejectedAsError:
+		// Own metric reason so auth-service contract violations show up as their own series.
+		cmon.InviteErrorShort(stats.ClientError("auth-rejected-as-error"))
```
hechen-eng, line 569: do we know what kind of other errors could there be? are they all client errors?

hechen-eng (approved):

pkg/service/psrpc.go
alexfish8: The auth service might return errors in two different channels: (a) the RPC error itself and (b) the `ErrorCode` field in the response itself. The channels are mutually exclusive, and so if we receive an RPC error then we won't have a response and an ErrorCode, and if we have a response and an ErrorCode then we won't have an RPC error.

pkg/sip/inbound.go
alexfish8: Since you've last viewed the PR, I've added handling for a new kind of auth failure (`ROUTE_NOT_ALLOWED`, which is returned when anycast isn't enabled).

The other sorts of auth failure are related to trunk configuration issues (i.e. client errors).

PR #837, opened 2026-09-07
Update media-sdk. Support codec parameters.

dennwc:

PR #845, opened 2026-09-12
feat: enable Opus codec on SIP trunks

### LINKED ISSUE #281: Support Opus on SIP Trunks
When trying to enable OPUS with `"OPUS/48000/2": true` in the config file, SIP invites still don't offer the Opus codec over the trunk.

According to the contents of [sip/pkg/sip](<https://github.com/livekit/sip/blob/418097f21eff8b87531004b428a25933ccbd4a0f/pkg/sip/media_codecs.go#L18-L22>)<br>[/media_codecs.go](<https://github.com/livekit/sip/blob/418097f21eff8b87531004b428a25933ccbd4a0f/pkg/sip/media_codecs.go#L18-L22>)<br>the only codecs supported on the SIP side of things are G711 and G722.

I looked at the `pkg/media/opus/opus.go` file and compared it to `pkg/media/g722/g722.go` and I see that some functions are similar, but the opus.go file is missing the `init()` function and doesn't register itself.

Many SIP trunks (particularly when connecting to a PBX of some sort) support the Opus codec, so it would be very useful to have it available for use on the SIP side.

lixuanqun:
## Summary

Adds Opus (`opus/48000/2`) to the default SIP media codec set, enabling Opus negotiation on both inbound and outbound SIP calls. Per RFC 7587 §6.1, the RTP clock rate is always 48000 Hz — different audio bandwidths are negotiated via `fmtp:maxplaybackrate`, not the rtpmap clock rate.

Closes #281 — the most-requested SIP feature with 40+ comments spanning WhatsApp Business Calling API, Twilio, Telnyx, and other SIP trunk providers.

### `pkg/sip/media_codecs.go`
- Added `opus` to the default codec set (priority 10, above G.711/G.722). The media-sdk registers the Opus media codec itself (via `media-sdk/all` on CGo builds) as of the #837 migration, so **no local codec registration is needed anymore** — the set entry is a literal name because `media-sdk/opus` is a CGo-only package and must not be imported unconditionally.
- Added `CheckCodecAvailability()` — runs at service startup, warns if Opus (or AMR-WB) is enabled in the set but its backing codec is not registered (typically because the binary was built with `CGO_ENABLED=0`)

### `pkg/sip/service.go`
- Calls `CheckCodecAvailability(s.log)` during `Service.Start()`

### `pkg/sip/media_port_test.go`
- Calibrates `TestMediaPortAudioRoundTrip` for Opus's encoder/decoder lookahead (~6.5 ms ≈ 311 samples at 48 kHz), same mechanism as the existing G.722/AMR-WB entries.

### `test/opus/`
- Unit tests for the codec set: default enablement, offer ordering (Opus first), set inheritance, name resolution, RFC 7587 clock-rate assertions.
- A docker-compose/sipp harness for manual trunk interop testing (not wired into CI).

## Relationship to #837 (media-sdk codec-params migration)

**Resolved.** #837 has landed on main and this PR is rebased onto it. The local `pkg/sip/opus_codec.go` registration was dropped in favor of the upstream one (`media-sdk/all` now registers Opus on CGo builds), and the codec-set/config handling was ported to the new name-based APIs (`SetEnabled`/`IsEnabledByName`/`CodecByNameWith` resolve by bare name, so the old rate-normalization special case for `opus` is no longer needed).

## Why this is safe

1. **The Opus codec implementation already exists** — `media-sdk/opus` provides a full Opus encoder/decoder with CGo (libopus); this PR only registers it as a media codec
2. **The RTP pipeline is codec-agnostic** — `media_port.go` uses `rtp.EncodePCM()`/`rtp.DecodePCM()` which work with any registered codec
3. **SDP negotiation already works** — `media-sdk/sdp` handles offer/answer; Opus is just another codec entry
4. **RTP clock matches LiveKit native rate** — `RoomSampleRate=48000` matches Opus RTP clock → zero resampling overhead (unlike G.711 which needs 8k↔48k conversion)
5. **DTMF payload types do not conflict** — Opus gets dynamic PT ≥101, telephone-event follows it, downstream handlers route by type

## RFC 7587 compliance note

Opus RTP clock rate is **always** 48000 Hz in `rtpmap`. Narrowband/wideband/fullband selection happens via `fmtp:maxplaybackrate`, not the clock rate. Our implementation does not emit `fmtp` lines (neither do G.711/G.722 in this codebase), meaning:
- As offerer: we accept any bandwidth the peer sends (libopus auto-detects per frame)
- As answerer: we match the peer's `rtpmap` (always 48000)

This is a valid minimal implementation. Adding `fmtp:useinbandfec=1` and `fmtp:minptime=10` would be a follow-up enhancement.

## Related

- [livekit/sip#281](https://github.com/livekit/sip/issues/281) — Support Opus on SIP Trunks (40 comments)
- [livekit/sip#837](https://github.com/livekit/sip/pull/837) — media-sdk codec-params migration (this PR will rebase onto it)
- [livekit/agents#3375](https://github.com/livekit/agents/issues/3375) — WhatsApp calling API support
- RFC 7587 — RTP Payload Format for the Opus Speech and Audio Codec


lixuanqun:
Heads-up for reviewers on the implementation approach:

The pinned media-sdk does not yet register Opus as a SIP media codec, and the API of the refactored media-sdk (which an earlier iteration of this PR was written against) doesn't exist there — that's what broke CI previously. To keep the change small and reviewable against current main, this PR now:

1. registers Opus locally in `pkg/sip/opus_codec.go`, mirroring what livekit/media-sdk#84 added upstream (including `Disabled: true` so it's opt-in via the codec set, and a `//go:build cgo` guard);
2. uses the canonical full SDP name `opus/48000/2` per RFC 7587 §6.1 and normalizes the bare `opus` config name in `codecSet()`;
3. adds a startup `CheckCodecAvailability` warning for CGo codecs that are enabled but not registered (e.g. `CGO_ENABLED=0` builds);
4. calibrates the media round-trip test for Opus' ~6.5 ms codec lookahead (311 samples at 48 kHz), same mechanism as the existing G.722/AMR-WB entries.

Once #837 (media-sdk codec-params migration) merges, we're happy to rebase and drop the local registration in favor of the upstream one — bare `opus` name resolution becomes native there.

Verified locally with golangci-lint v2.13.1 (same as CI) and the full `go test ./pkg/... ./test/opus/` suite; the round-trip test now exercises Opus end-to-end through the media pipeline.

PR #466, opened 2025-09-29
Only pass full frames to opus encoder. Increase buffer size.

dennwc:
Closes https://github.com/livekit/sip/issues/465

PR #847, opened 2026-09-14
Pad the final Opus frame before closing the encoder

inlanger:
Closing the Opus encoder with a partial PCM frame can return `opus: invalid argument`: `WriteSample` buffers complete frames, but `flush` passes the remaining samples directly to libopus. For example, writing one PCM sample at 48 kHz succeeds and then `Close` fails. Even a remainder that is a valid shorter Opus frame produces a duration different from the regular frames.

Pad a nonempty final frame with silence up to the existing `e.samples` size, clearing any stale samples in the reused buffer. Reset the buffer after encoding so a later close cannot emit the tail again. Empty buffers emit nothing, and the underlying writer is still closed when flushing fails. This adds no dependencies or configuration and leaves the regular encoding path unchanged.

Follow-up to #466. This addresses final Opus packetization; media draining and call teardown order are separate.

Validation (Go 1.26.3, libopus 1.5.2, Linux arm64):

- The new regression tests fail on the unmodified main implementation and pass with the fix: `go test -count=1 -race -v ./pkg/media/opus`.
- Tests cover empty, full and partial frames at 8, 16 and 48 kHz, reuse after a complete frame, repeated close, and writer errors. Output packets are compared with direct libopus encoding of the original PCM plus silence, then decoded to check their duration.
- `go test -count=1 -race ./pkg/...` reaches an unrelated data race in `TestMediaPipelinePermutations`: `media-sdk`'s `byteEncoder.Close` races with mixer writes. Reproduced on unmodified base `72da1b63` with `go test -count=1 -race -run '^TestMediaPipelinePermutations$' ./pkg/sip`. That code is unchanged here.
- End-to-end tests were not run locally.


inlanger:
Could a maintainer rerun the failed **Test** job?

The Opus tests and **Build / integration** passed. The [failed job](https://github.com/livekit/sip/actions/runs/34825649738/job/103917057409) reports `TestTransfer/outbound/terminated_subscription_success`: the REFER transaction fails with `transaction failed to complete (0 intermediate responses)`, followed by a timeout waiting for BYE.

This test uses [`testRoom.NewParticipantTrack`, which returns a `noOpWriter`](https://github.com/livekit/sip/blob/72da1b63e2bddb458a7e170f77d78f518c790c36/pkg/sip/utilities_test.go#L248-L251), so it does not exercise the Opus encoder changed here.

I also checked the unmodified base (`72da1b63`) with:

```sh
go test -count=100 -race -run '^TestTransfer$/^outbound$/^terminated_subscription_success$' ./pkg/sip
```

That run had 12 failures at `service_test.go:75` (`Transaction failed to complete`) before aborting with `panic: Write called after TestTransfer has completed` from a delayed callback at `outbound.go:849`. It did not complete all 100 iterations. The exact BYE timeout from CI did **not** reproduce locally, so this confirms instability on the base but does not establish an identical underlying cause.

No code changes have been made in response to this CI failure.

PR #853, opened 2026-09-17
TELCU-30 - Record inbound call status codes

ChelseaBradbury:
Currently we don't record the terminal SIP status code for inbound calls in a way that allows them to be persisted to analytics. So when a caller hangs up before the call is answered, or if there's a media failure, etc. we still display a `200` sip status for the call on the dashboard.

If you look here, you can see that we always set the SIP code for inbound calls to 200, since we never populate `SIPCallInfo.CallStatusCode`:
https://github.com/livekit/cloud-sip/blob/cb72d104bc70c8c0dafe74f7a30135c42bde4da2/pkg/service/analytics.go#L415-L418
https://github.com/livekit/cloud-sip/blob/cb72d104bc70c8c0dafe74f7a30135c42bde4da2/pkg/service/analytics.go#L485-L489


pkg/sip/inbound.go
```diff
@@ -1560,6 +1560,15 @@ func (c *inboundCall) close(ctx context.Context, end EndCall) {
 			Status: "Request Terminated",
 		}
 	}
+
+	// Record the call outcome
+	c.state.DeferUpdate(func(info *livekit.SIPCallInfo) {
+		info.CallStatusCode = &livekit.SIPStatus{
+			Code:   livekit.SIPStatusCode(result.Code),
+			Status: result.Status,
+		}
```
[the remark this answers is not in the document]
genseric-ghiro: @ChelseaBradbury I think Devin makes sense here, result is set to Busy by default above, so even on a normal flow we'd end up sending a 486 to analytics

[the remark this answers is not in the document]
ChelseaBradbury: Yeah, I'm gonna add some tests and fix it so every outcome gets the right code.

pkg/sip/inbound.go
```diff
@@ -308,12 +338,17 @@ func (s *Server) handleInviteAuth(tid traceid.ID, log logger.Logger, req *sip.Re
...
 			"receivedResponse", cred.Response,
 		)
-		_ = tx.Respond(sip.NewResponseFromRequest(req, 401, "Unauthorized", nil))
-		return false, false
+		res := sip.NewResponseFromRequest(req, 401, "Unauthorized", nil)
+		sentStatus := &Result{
+			Code:   res.StatusCode,
+			Status: res.Reason,
+		}
+		_ = tx.Respond(res)
```
genseric-ghiro: Seems like a similar issue that Devin pointed out earlier, i.e. we don't check if Respond succeeds, return sentStatus, and `processInvite` below record that status. Same thing for all other instances above

pkg/sip/inbound_test.go
```diff
@@ -16,12 +16,15 @@ package sip
```
genseric-ghiro, line 16: This is a lot of new tests for one status code. We can keep them, but we should isolate them better. I would move them to their own struct/class, and extract common logic (e.g. states := &recordingStateHandler{}, st := NewServiceTest(t, &serviceTestConfig{GetStateHandler: states.GetStateHandler()}), etc.`) into a beforeEach(). The rationale for this is that as tests are growing, we might start to see CI slow down. At some point, I expect we'll need to profile tests to see what takes so long, and organizing them by class will definitely help quickly answering where CI time is spent

pkg/sip/inbound.go
ChelseaBradbury: Ah, right; I missed these ones. I'll take care of it 👍

pkg/sip/inbound_test.go
ChelseaBradbury: I didn't really take the time to clean these up, but I'll do it now. Thanks!

pkg/sip/inbound.go
ChelseaBradbury: Okay, fixed

pkg/sip/inbound_test.go
ChelseaBradbury: Reorganized the tests and abstracted the setup logic... I think that covers your concerns. The tests take about 40 seconds total; I think it does make sense to keep them.

genseric-ghiro: > 40 seconds total

Oh, that's a lot. For comparison, this is what I see from the test runs in main vs your branch
```
 Elapsed │                 Package                  │ Cover
your branch -> 108.11s │ github.com/livekit/sip/pkg/sip           │ 68.8%
main -> 65.43s  │ github.com/livekit/sip/pkg/sip           │ 62.3%
```

+~43s is a 66% increase, definitely a lot, especially considering coverage jumps by only ~6%. What's taking so long to run these tests? Is it the setup? Or is it busy waiting? I saw a few `require.Eventually` up to 5s in your code, this is definitely high. Whatever it is, we should aim to have these run in 3-4s

ChelseaBradbury: Looks like it's mostly because of the `NoACK`, `PinTooLong`, and `WrongPin` tests. I'll see if I can sort out why those take so long.

I'm seeing other tests re-use the same `serviceTest` instance, so I'll give that a shot to see if I can cut down the setup time for the whole set.

Screenshot 2026-09-21 at 1 04 10 PM

ChelseaBradbury: Okay, managed to reduce the runtime for the tests.

`NoAck` is still slow because it's waiting on retries (hardcoded retry interval)
`PinTooLong` and `WrongPin` are still slow because they have to wait for the 'enter pin' audio to play.

I'd be okay with just removing those tests, if that's still too long.

Screenshot 2026-09-21 at 3 51 28 PM

genseric-ghiro: Yes, let's remove these 3 tests, I think there's already good coverage with the others

genseric-ghiro (approved):
Thanks, LGTM!

PR #855, opened 2026-09-17
feat: support early media (183/180 with SDP) on outbound calls

### LINKED ISSUE #813: Support early media (183/180 with SDP) on outbound calls
On outbound calls, livekit/sip doesn't relay any audio the far end sends before the final 200 OK. If the callee's network plays something during ringing via a 183 Session Progress (or 180 Ringing) with an SDP body — a custom ringback, an IVR announcement, or a carrier-side personalized ringback/caller-tune service — that audio never reaches the LiveKit room. The room just stays silent (or hears the local play_dialtone tone if that option is set) until the call is actually answered.

I traced this to pkg/sip/outbound.go:

NewMediaPort is created for outbound calls with IgnorePreanswerData: true (MediaOptions.IgnorePreanswerData), so any RTP that arrives before the RTP/SRTP session is set up gets explicitly discarded (media_port.go's startDiscarding/discard loop).
The RTP/SRTP session itself is only ever created once, in sipSignal, via c.media.SetAnswer(...) + c.media.SetConfig(...) — and that only runs after Invite() returns with the final 200 OK body.
The response-handling loop (sipResponse, used by attemptInvite/Invite) does read every provisional (1xx) response, but only extracts the status code and headers for call-state tracking (RingingTime, CallRinging, etc.) — the SDP body of a 1xx response is currently discarded, never passed to anything that could act on it.

So even though SetAnswer/SetConfig are perfectly capable of negotiating from any SDP answer, there's currently no path that calls them with a provisional response's body — only the final one.

Proposed behavior

When a provisional response carries an SDP body (most commonly a 183 Session Progress), negotiate and connect the RTP/SRTP pipeline from that SDP immediately, the same way it's done for the final 200 OK today. That lets real early-media audio reach the room during ringing. When the actual 200 OK arrives afterward, if it points to a different remote address, update the destination (there's already an UpdateRemote/updateRemoteFromSDP helper used for re-INVITEs that fits this) instead of tearing down and rebuilding the session.

Two things worth calling out for the design:

This should probably be opt-in (e.g. a feature flag alongside the existing sip.signal_logging / sip.outbound_route_headers flags), since some trunks may send inconsistent or incomplete SDP on provisional responses, and this is new codepath that not everyone will want by default.
It should only affect the SIP-audio-into-room direction. The room-to-SIP (agent/caller audio) direction is currently gated behind EnableOut() at final accept, and that should probably stay as-is — you don't want to start sending audio into a call the far end hasn't answered yet.

lixuanqun:
## Summary

On outbound calls, livekit/sip discarded any SDP body on a provisional 1xx response (183 Session Progress, 180 Ringing with SDP) and the RTP pipeline stayed dark until the final 200 OK. Carrier-side ringback, IVR announcements, and personalized caller-tune audio never reached the LiveKit room during ringing.

Closes #813.

### `pkg/sip/media_port.go`
- **New `MediaPort.ProcessEarlyAnswer(answer []byte) error`** — connects the RTP pipeline from a provisional response's SDP body. Unlike `ProcessAnswer`, it does **not** consume the offer: the offer is restored after `configure`, so the final 200 OK can still be processed via `ProcessAnswer`. If the final answer carries a different remote address or codec set, the pipeline is rebuilt by `ProcessAnswer`'s `configure` call (same path re-INVITEs use).

### `pkg/sip/outbound.go`
- **New `sip.early_media` feature flag** (opt-in, disabled by default) — some trunks send inconsistent or incomplete SDP on provisional responses, so this is gated behind a flag rather than on by default.
- **`sipResponse` takes an optional `onProvisional` callback** — invoked for 1xx responses that carry an SDP body. The callback checks `Content-Type: application/sdp` before calling `ProcessEarlyAnswer`.
- **`sipSignal` wires the callback** — when the flag is enabled, provisional SDP connects the RTP pipeline so early-media audio reaches the room before the call is answered.

### `pkg/sip/protocol.go`
- Updated the two existing `sipResponse` call sites (`sendBye`, `sendRefer`) to pass `nil` for the new `onProvisional` parameter — their behavior is unchanged.

## Design notes (from issue #813)

- **SIP-to-room direction only.** The room-to-SIP (agent/caller audio) direction is unchanged: outbound audio is still gated behind the final accept (`EnableOut`), so no audio is sent into a call the far end hasn't answered yet.
- **Offer preserved.** `ProcessEarlyAnswer` calls `configure` (which rebuilds the pipeline and sets `p.offer = nil`) then restores `p.offer`, so `ProcessAnswer` for the final 200 OK still works. If the final answer differs from the provisional one, `configure` is called again and the pipeline is rebuilt — same as a re-INVITE.

## Tests

Two new tests:

- **`TestMediaPortEarlyAnswer`** — generates an offer, processes the answer via `ProcessEarlyAnswer`, verifies `RemoteAddr` is set and the offer is still present, then calls `ProcessAnswer` for the final 200 OK and verifies `NegotiatedAudio` is available.
- **`TestSIPResponseEarlyMediaCallback`** — verifies `sipResponse` invokes `onProvisional` only for 1xx responses with an SDP body, not for bodyless provisionals or final responses.

Verified locally with golangci-lint v2.13.1 (same as CI, 0 issues) and the full `go test ./pkg/...` suite.


lixuanqun:
Self-review pass — Devin flagged 3 issues; I verified all three against the current code and they are real:

1. **Final-SDP codec change leaves stale pipeline** (`media_port.go:891`). `configure` returns early (`shouldReconfigure() == false` → `return nil`) when the pipeline is already established, so a final 200 OK that changes codec/PT/DTF/crypto relative to the provisional 183 does **not** rebuild. The call stays on the provisional codec. Fix: `ProcessEarlyAnswer` should either tag the early-media `configure` so the final `ProcessAnswer` forces a reconfigure (e.g. a `preliminary bool` on the changeSet / a dedicated reconfigure path), or `ProcessEarlyAnswer` should skip `configure` and only prime the RTP destination, deferring pipeline build to the final answer.

2. **Early media-timeout fuse is irreversible** (`media_port.go:897`). `SetTimeout` arms `mediaTimeout` (a `core.Fuse`); if the carrier sends 183 SDP but no early RTP, the 30s initial timeout breaks the fuse and `mediaTimeoutLoop` exits for the life of the call. The final 200 OK cannot re-arm it. Fix: do not arm the media timeout on `ProcessEarlyAnswer` (defer it to `ProcessAnswer`), or add a re-arm path that `ProcessAnswer` calls.

3. **Devin flagged "early media never reaches the room"** (`outbound.go:757`) — I traced this and the ordering is actually fine: `connectSIP` calls `connectMedia` (line 479) before `sipSignal` (line 549), so `WriteInboundAudioTo` is already wired when the 183 arrives. Noting it so a maintainer does not re-flag it.

Codecov patch coverage is 68% (13 missing lines in `outbound.go` and `media_port.go`), mostly the feature-gated early-media paths; worth unit tests around the `earlyMediaEnabled=false` and empty-body branches.

Happy to implement the fixes for 1 and 2 if you want them in this PR.

PR #858, opened 2026-09-17
TEL-1066: report ringing-timeout race as no-answer, not deadline-exceeded

### LINKED ISSUE #857: Outbound ringing timeout sometimes surfaces as failed_precondition / "context deadline exceeded" (412) instead of canceled / "sip request timed out" (408)
## Summary

`CreateSIPParticipant` with `wait_until_answered=true` and a `ringing_timeout` normally reports a no-answer as:

```
ServerError(code=canceled, message=twirp error unknown: sip request timed out, status=408)
```

Roughly 1 in ~1000 no-answers on our trunk instead returns:

```
ServerError(code=failed_precondition, message=twirp error unknown: context deadline exceeded, status=412)
```

Both fire at exactly `ringing_timeout` after the request (35 s in our case, measured to ±0.2 s across every occurrence). The user-facing outcome is identical (callee did not answer in time), but the error code and message are different, so any client branching on the documented `canceled` code to detect no-answer treats the second variant as a hard failure.

## Root cause (from reading `f75e9ce`)

Both errors originate from the same ringing-timeout context created in `sipSignal` (`pkg/sip/outbound.go:695`).

**Normal path (408):** `sipResponse` (`outbound.go:580`) selects on `ctx.Done()` while waiting for the final response and returns `psrpc.NewError(psrpc.Canceled, ErrSIPRequestTimeout)`. The comment there deliberately avoids `DeadlineExceeded` so psrpc doesn't auto-retry.

**Race path (412):** if the callee answers within the same instant the context expires, the `select` in `sipResponse` can take the 200 OK from `tx.Responses()` instead of `ctx.Done()`. `Invite` completes, `AckInviteOK` is sent, and then `sipSignal` hits the post-ACK abort check (`outbound.go:784-789`):

```go
if err := ctx.Err(); err != nil {
    // Aborted while the callee answered: the 200 is ACKed, so error out to
    // tear the dialog down with a BYE instead of leaking a zombie call.
    c.log.Infow("outbound call answered after abort; hanging up", "error", err)
    return err
}
```

This returns the raw `context.DeadlineExceeded`. `classifyInviteError` (`pkg/sip/errors.go:279`) maps it to `stats.ServerError("deadline-exceeded")` with `psrpc.DeadlineExceeded`, and because the INVITE was already sent, `inviteFailure.afterInvite()` (`errors.go:32`, called from `outbound.go:436`, introduced in #736) rewrites the server-error result to `psrpc.FailedPrecondition`. That is where `failed_precondition` / "context deadline exceeded" / HTTP 412 comes from.

Side effects beyond the inconsistent code:

- The call is bucketed as `ServerError("deadline-exceeded")` in your SLI stats although it is a client-configured ringing timeout, not a server fault.
- The 412 is `failed_precondition`, which suggests to the caller that their request was invalid.

## Suggested fix

Treat a ringing-timeout expiry that races the answer the same as one that doesn't. Either:

- at the post-ACK abort check, return `psrpc.NewError(psrpc.Canceled, ErrSIPRequestTimeout)` when `errors.Is(ctx.Err(), context.DeadlineExceeded)` and the deadline came from `ringingTimeout`, or
- in `classifyInviteError`, map `context.DeadlineExceeded` during the dial phase to the same `callUnavailable` / `ClientError("no-answer")` / `USER_UNAVAILABLE` verdict as `ErrSIPRequestTimeout` (which also stops `afterInvite()` from flipping it to `FailedPrecondition`).

The BYE teardown the check exists for is unaffected either way.

## Evidence

Python SDK (`livekit-api`), `CreateSIPParticipantRequest(wait_until_answered=True, ringing_timeout=35s, play_dialtone=..., krisp_enabled=False)`, Twilio outbound trunk, LiveKit Cloud.

Two occurrences on 2026-09-16 (UTC), both +35.2 s after the request:

| room              | CreateSIPParticipant sent | error returned |
| ----------------- | ------------------------- | -------------- |
| `RM_ykumPm2yyWkE` | 00:31:27.86               | 00:32:03.09    |
| `RM_LwUV3mhjkkvV` | 16:23:58.60               | 16:24:33.82    |

In both rooms the SIP participant joined and reached `ringing` (Twilio Call SID visible in participant attributes) ~1-2 s after the request. On the same day, the same request pattern produced 916 × `canceled` / "sip request timed out" and 1 × this 412, so the rate is ~0.1% of no-answers.

hechen-eng:
Fixes #857.

## Fix

Tag the ringing deadline with its cause (`context.WithTimeoutCause(..., ErrSIPRequestTimeout)`) and report `context.Cause(ctx)` at the abort check. Both expiry paths then hit the same `errors.Is(err, ErrSIPRequestTimeout)` branch → `callUnavailable` / `ClientError("no-answer")` / `USER_UNAVAILABLE`, `Canceled`.

A caller-side abort (participant disconnect, call closed) has cause `context.Canceled` and still classifies as `ClientError("canceled")` — this does not fold every abort into no-answer.


pkg/sip/outbound.go
```diff
@@ -781,11 +782,14 @@ func (c *outboundCall) sipSignal(ctx context.Context, tid traceid.ID) error {
...
 	joinDur()
 
-	if err := ctx.Err(); err != nil {
+	if ctx.Err() != nil {
 		// Aborted while the callee answered: the 200 is ACKed, so error out to
 		// tear the dialog down with a BYE instead of leaking a zombie call.
-		c.log.Infow("outbound call answered after abort; hanging up", "error", err)
-		return err
+		// Report the cause, not ctx.Err(): a ringing timeout landing here is
+		// the same no-answer as one that beats the 200 OK in sipResponse.
```
genseric-ghiro: "no-answer as one that beats the 200 OK in sipResponse." -> not sure that I understand this part of the comment

pkg/sip/outbound.go
```diff
@@ -781,11 +782,14 @@ func (c *outboundCall) sipSignal(ctx context.Context, tid traceid.ID) error {
...
+		cause := context.Cause(ctx)
+		c.log.Infow("outbound call answered after abort; hanging up", "error", cause)
```
genseric-ghiro, line 794: Does the cause contain the error? If not, then we stop returning it and don't log it anymore, so we lose visibility.

hechen-eng: The context.Cause(ctx) won't be nil when ctx.Err() is non-nil, so this will contain the error 👍

pkg/sip/outbound.go
hechen-eng: rephrased 👍

genseric-ghiro (approved):

PR #860, opened 2026-09-18
TEL-884: measure inbound call setup answered vs not answered

hechen-eng:


pkg/stats/monitor.go
```diff
@@ -524,6 +534,25 @@ func (c *CallMonitor) JoinDur() func() time.Duration {
...
+	// SetupAbandoned means the call closed before we ever tried to answer:
+	// caller CANCEL, a hangup while still ringing, or a setup failure.
+	SetupAbandoned SetupOutcome = "abandoned"
+)
+
+// SetupDur records how long call setup lasted, measured from the INVITE.
+func (c *CallMonitor) SetupDur(outcome SetupOutcome, dt time.Duration) {
+	c.m.durSetup.With(c.labelsShort(prometheus.Labels{
+		"outcome": string(outcome),
+	})).Observe(dt.Seconds())
```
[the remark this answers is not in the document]
hechen-eng: this is intended

pkg/sip/inbound.go
```diff
@@ -1576,6 +1576,15 @@ func (c *inboundCall) close(ctx context.Context, end EndCall) {
 	}()
 	c.setStatus(end.Status)
 	c.mon.CallTerminate(end.Term)
+	// AcceptTime is set when acceptCall finishes trying to answer, so a zero
+	// value separates calls we never answered from the rest.
+	if !c.sigTs.InviteTime.IsZero() {
+		if c.sigTs.AcceptTime.IsZero() {
+			c.mon.SetupDur(stats.SetupAbandoned, time.Since(c.sigTs.InviteTime))
+		} else {
+			c.mon.SetupDur(stats.SetupAnswered, c.sigTs.AcceptTime.Sub(c.sigTs.InviteTime))
```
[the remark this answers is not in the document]
hechen-eng: not a problem

pkg/stats/monitor.go
```diff
@@ -228,6 +229,15 @@ func (m *Monitor) Start(conf *config.Config) error {
...
 	}, []string{"dir"}))
 
+	m.durSetup = mustRegister(m, prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace:   "livekit",
+		Subsystem:   "sip",
+		Name:        "dur_setup_sec",
+		Help:        "Call setup duration: from INVITE to either 200 OK (outcome=answered) or to close without ever answering (outcome=abandoned).",
+		ConstLabels: prometheus.Labels{"node_id": conf.NodeID},
+		Buckets:     durBucketsOp,
+	}, []string{"dir", "outcome"}))
```
genseric-ghiro, line 239: If we only care about inbound, why keeping the direction label?

hechen-eng: ~~we don't have to.. will remove~~

we might need something for outbound later, so I'll just keep it here..

genseric-ghiro: Then we should set the direction in `SetupDur` to inbound, otherwise it gets confusing when looking at the metric in grafana (i.e. others won't know if it's inbound or outbound)

hechen-eng: Inside `SetupDur`, it does call `labelsShort` which sets `dir`

genseric-ghiro: Ah sorry, missed that, LGTM!

genseric-ghiro (approved):

PR #861, opened 2026-09-21
Propagate SIP_TRUNK_FAILURE to livekit server

alexlivekit:


pkg/sip/errors.go
```diff
@@ -68,7 +68,7 @@ func (e SDPError) ClassifyInvite() inviteFailure {
 	res := inviteFailure{
 		EndCall: EndCall{
 			Status: callRejected,
-			Reason: livekit.DisconnectReason_MEDIA_FAILURE,
+			Reason: livekit.DisconnectReason_USER_REJECTED,
```
alexlivekit, line 71: If the destination doesn't have the proper codec, or doesn't support encryption, that isn't a media issue - it's negotiation and being rejected by the remote. Also matches today's mapping from `Status: callRejected,` in the line above.

pkg/sip/outbound.go
```diff
@@ -381,7 +383,7 @@ func (c *outboundCall) close(ctx context.Context, end EndCall) bool {
 		}
 
 		if r := c.lkRoom; r != nil {
-			_ = r.CloseWithReason(end.Status.DisconnectReason())
+			_ = r.CloseWithReason(end.Reason)
```
alexlivekit, line 386: This is THE change here. Mostly everything stays the same, except a bunch of places where today we report `SIP_TRUNK_FAILURE`.

hechen-eng, line 386: we can remove `DisconnectReason` method in `participant.go` then

hechen-eng (approved):

